PR #3, opened 2017-10-09
Rudimentary Nissan Leaf Support + partial v2 Stat message implementation

carrott:
This adds initial Nissan Leaf support and a partial implementation of the v2 Stat message. Also a battery current metric.

I'm not sure if the AsStringDefault is idomatic C++, but I had trouble with overloaded methods and hiding.

Sending the Stat message is triggered from the wrong place.


markwj:
I think the standard C++ mechanisms for providing pre-initialised default values as parameters should work, so have modified the AsString, AsInt, etc functions to use that. It doesn’t change a normal call (so current behaviour is unchanged), but does give you a way of providing a default value for undefined metrics.

For example, for OvmsMetricBool, we have:

    std::string AsString(const char* defvalue = "");

    int AsBool(const bool defvalue = false);

Let’s look at the implementation of AsBool():

int OvmsMetricBool::AsBool(const bool defvalue)
  {
  if (m_defined)
    return m_value;
  else
    return defvalue;
  }

So, you can call metric->AsBool(true) to make ‘true’ the default. Or metric->AsString(“undefined") to make “undefined” the default.

I’ve pushed this, and I think it should satisfy your requirements. Can you modify to adopt this?

markwj:
You’re right, it is ugly and dangerous.

I re-worked my tpms example, to use std::string:

  std::string buffer;
  buffer.reserve(512);
  buffer.append("MP-0 W");
  buffer.append(StandardMetrics.ms_v_tpms_fr_p->AsString());
  buffer.append(",”);
  ...

That pre-allocates 512 bytes (for performance), but can grow if that reserved size is exceeded.

What do you think of that? It looks better, but I still don’t really like all those append(“,”) lines.

Can you try to re-work your code, to use these three new things, and then re-submit a pull request?

carrott:
Thanks for reviewing, I've updated the branch and also avoided the double free when the v2 server code is started from a script by calling ESP_LOGI() instead of writer->puts()

PR #4, opened 2017-10-17
Nissan leaf

carrott:
Fix bug in stat v2 server message
Fill in more of the environment v2 server messages
Ported more Nissan Leaf code from v2

Should the doors unions defined in the v2 server be in the .cpp file or the .h? They're private and don't need to be exposed to the rest of the code base.

Is there already better way to do the car off detection, which of my suggestions, if any, is most appropriate?

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
@@ -45,7 +45,8 @@ static const char *TAG = "ovms-server-v2";
 #include "esp_system.h"
 
 // should this go in the .h or in the .cpp?
-typedef struct {
+typedef union {
+  struct {
   unsigned FrontLeftDoor:1;     // 0x01
   unsigned FrontRightDoor:1;    // 0x02
   unsigned ChargePort:1;        // 0x04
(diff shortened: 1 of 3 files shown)


markwj:
Confused about your change from:
   if (me->m_can1 == frame.origin) me->IncomingFrame(&frame);
to:
  me->IncomingFrame(&frame);

Surely you need that check against the frame origin to ensure you only handle can1 frames. If you want to handle frames from different can buses, then better to send them to different IncomingFrame member functions. Like:

  if (me->m_can1 == frame.origin) me->IncomingFrameCan1(&frame);
  if (me->m_can2 == frame.origin) me->IncomingFrameCan2(&frame);
  etc

Can you review this?

carrott:
I moved the logic to select where frames go to OvmsVehicleNissanLeaf::IncomingFrame(). This facilitates unittesting because you can throw frames at that function rather than having the test framework invoke the infinite loop in NL_rxtask and have it pull messages out of a test queue by implementing a special version of xQueueReceive().

You don't see IncomingFrame() in the diff because it was already implementing the required logic which  didn't work because NL_rxtask() was only sending messages from one bus.

I don't have any unit tests just yet.

markwj:
Just accept as it is, as things will break anyway with the vehicle.{h, cpp} rework.

PR #6, opened 2017-11-01
v2 server: make various v2 values integer

carrott:
The android client crashes on receipt of a floating point value when it's expecting an integer. This change updates charge current, line voltage and charge current limit to integral values rather than the floating point in the storage.

I chose to add another unit conversion for float to int, but I'm not sure that this i right, since it precludes users of the api from performing a unit conversion and converting to integer. Perhaps AsIntString(default, unit) is a better interface?

I also did a simple cast to int rather than a round, I'm not sure if we care too much about that?


markwj:
I think I have a better implementation of this, by exposing the ostringstream precision+fixed parameters to metric accessor AsString(). That means we can do this, even if converting between units. That is committed, and pushed, so I don't think this pull is needed and can be dropped.

PR #11, opened 2017-12-02
Kia Soul EV - first working version.

goev:
Ported most of the important V2-functionality. Partly tested on real car.


slcasner:
The approach I took for the wolfssl and wolfssh directories was to
include the full distributions even though many of the files are not
used, including .project.  Since these distributions are GPL, it
seemed that including all the files would help meet redistribution
requirements, but I'm not sure what our strategy should be.

                                                        -- Steve

dexterbg:
My 2 cents:

- It's correct to include imported components as distributed.
- Our V3 core build environment is the esp-idf toolchain. Every developer is free to use any or no IDE on top of that, so I think we should not add IDE specific config files to the repository.

goev:
Oh... I thought I had excluded those files already. I’ll fix it. 

Geir 

Sendt fra min iPhone

PR #13, opened 2017-12-03
Kia Soul EV - First working version.

goev:
Compiles, partly tested on real vehicle.

Michael, is it looking better now?


goev:
Hehe! Thanks, Michael. At least I’m getting better 🙂

Geir 

Sendt fra min iPhone

PR #16, opened 2017-12-24
Kia Soul Branch

goev:
- Kia Soul EV requires OVMS GPS, since we don't know how to access the cars own GPS at the moment.
- Minor bugfixes after quick in-car test.
- Open trunk, lock/unlock doors and open chargeport + minor changes.
- Added xks sjb - command + RearDefogger, Leftindicator, right indicator.
- "Trunk"-button of keyfob opens up the chargeport
- Added temporary, developer-commands: sjb and bcm
- Added methods for controlling ACC-relay, IGN1-relay, IGN2-relay and start-relay.
- Electric park break service-command
- New config: xks.remote_charge_port - enables the possibility to open charge port from keyfob
- Renamed config-root from x.ks to xks
- Renamed xks to ks - Config, metric list and command.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_kiasoulev/src/vehicle_kiasoulev.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_kiasoulev/src/vehicle_kiasoulev.cpp
@@ -57,6 +57,10 @@
 ;		 0.2.6  22-Dec-2017 - Geir Øyvind Vælidalo
 ;			- Electric park break service-command
 ;
+;		 0.2.7  22-Dec-2017 - Geir Øyvind Vælidalo
+;			- New config: xks.remote_charge_port - enables the possibility to open charge port from keyfob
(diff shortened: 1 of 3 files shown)


dexterbg:
I see a problem with renaming your command root from "xks" to "ks". This will produce trouble when a framework module needs to use "ks" -- or any other vehicle code if others follow this. That's why Mark suggested using the "x" prefix for all dynamic modules, so framework modules don't need to know/care about namespace collisions with loadable modules.

I second dropping the dot from "x.ks" within config & metrics, that doesn't need to be there and would harmonize names with the command root.

So I suggest using "xks" for your commands, configs & metrics.

goev:
Ah, I see. I didn’t catch that suggestion. Too many mails🙂
Just ignore the pull request and I’ll go back to xks and send another pull request. 

Merry Christmas to all of you!!!

Geir 

Sendt fra min iPhone

PR #135, opened 2018-06-24
Nissan leaf remote climate control off

carrott:
Use the Nissan native facility to turn off climate control after it is started remotely. Replaces the same functionality implemented in the OVMS. Thank you Stein Arne Sordal for providing the necessary information to use this feature.


glynhudson:
I think 15 minutes is the default duration for the climate control to run when not plugged in as defined by the car. How long are you currently experiencing the climate control vent for when started via ovms?

ltjessem:
Alright, so it's completely up to the car now when to turn off, AUTO_DISABLE_CLIMATE_CONTROL just tells the car to turn it off when it sees fit?
Mine is not running more than 10 minutes at a time, but I've only tested it twice. I'll probably revert to the older code so I can keep it on much longer.

PR #137, opened 2018-07-01
Server V3: allow users to define a custom topic prefix

M4GNV5:


M4GNV5:
There seems to be a race condition in the mongoose MQTT implementation, which makes the OVMS crash when sending the all metrics at once, as it does after connecting. I've opened a ticket at mongoose: https://github.com/cesanta/mongoose/issues/943.

M4GNV5:
Alright, i did some further testing and implemented a fix for https://github.com/cesanta/mongoose/issues/943. Now the v3 server seems to be more stable.

I made a small change to the netmanager to prevent an connection attempt when `NetUp` is fired without any interface being up. Together with [this patch to mongoose](https://github.com/openvehicles/mongoose/pull/1) this prevents the v3 server from making the ovms crash during startup.

PR #142, opened 2018-07-18
Added support for Zeva BMS 

prensel:
I'm working on adding support for Zeva BMS based vehicles.
Can this be integrated into the master ?


dexterbg:
Yes it can, thanks for contributing :)

PR #166, opened 2018-11-03
Respect the users decision whether to enable or disable gps

M4GNV5:
Currently some vehicles force the OVMS GPS and thus the simcom to be enabled. Even when the users disables GSM and GPS via the web config it gets enabled. In fact various vehicle ports required gps, even though the only one using it seems to be the Twizy with its gps logging.

This pull request simply removes the "anti-feature" of forced gps & simcom.

This is sort of a design decision PR, so the following represents my point of view:
This reminds me a lot of the ["the program controls the user"](https://youtu.be/7twCCWjSnMg?t=57) term, introduced by Richard Stallman. Even though we give the user a choice to enable or disable GPS and the simcom module we ignore that decision and make the OVMS do what we want it to do, i.e. let it control the user. So by having this the OVMS acts a lot like a commercial product which does stuff you do not want it to do, even though you tell it not to, like Windows updating your PC when you don't want it to, or Android (the Google App to be specific) telling your Location to google when you don't want it to.

In the end, what do we really gain from it? What's the advantage of force enabling gps & simcom?

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg.cpp
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg.cpp
@@ -706,8 +706,7 @@ void OvmsWebServer::HandleCfgModem(PageEntry_t& p, PageContext_t& c)
   c.fieldset_start("Features");
   c.input_checkbox("Enable SMS", "enable_sms", enable_sms);
   c.input_checkbox("Enable GPS", "enable_gps", enable_gps);
-  c.input_checkbox("Use GPS time", "enable_gpstime", enable_gpstime,
-    "<p>Note: GPS &amp; GPS time support can be left disabled, vehicles will activate them as needed</p>");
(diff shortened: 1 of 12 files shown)


dexterbg:
I tend to agree. The original idea (see http://lists.openvehicles.com/pipermail/ovmsdev/2017-December/003637.html) was to let the vehicle module decide about requesting the optional gps and gpstime services from the system (which may be provided by a modem or other component, or may not be available).

It's a bug the modem signal handler still powers up the modem if the modem is explicitly disabled -- the auto start config was introduced later and I missed to add that check to the "require" handler.

But the signal scheme also does not adapt well to configurations with multiple gps/gpstime providers, and it does not allow a user to explicitly disable/select single services without introducing new config variables with essentially the same semantics as the existing.

Removing the code without a config update will leave many current users with broken configurations though. Also first time init of a module should enable the modem services on vehicles relying on them. Users can still disable them later on. So we'll need some config init/upgrade along with this change.

Other opinions on this?

PR #175, opened 2018-11-22
Add Leaf BMS voltage and temperature

caederus-ovms:


markwj:
I am looking at this IncomingPollComplete, pull request, but it seems very confusing to me.

The current design is that OvmsVehicle::PollerReceive calls IncomingPollReply() in several places, to pass the result up to the vehicle code for decoding. It pretty much just passes the data part of the frame (after decoding the first part containing PID, result code, etc).

Now, you have IncomingPollComplete(). You have changed OvmsVehicleIncomingPollReply from pure virtual to now have code?

You have a ’static uint8_t *buf’ defined in your new IncomingPollReply, but I can’t see where that get freed. Also there only seems to be one, so what happens when multiple ‘ml’ style polls are defined? I’m really not sure if this should be a static - perhaps a vehicle object member variable would make more sense (Iike the other m_poll member variables, as this IncomingPollComplete is concerning m_poll_ml)?

I am really not sure what is required/solved here? The vehicle.{h,cpp} change is mixed in with changes to the nissan leaf vehicle code, in the same commits, which makes things harder to see. In future, could you make one commit to the base code, and one to the vehicle module - if they are related/dependent, they can be in the same pull request, but mixing everything together with commits changing commits, makes it hard to see what is going on (or the impact that the changes to base code could have on other vehicle modules).

Regards, Mark.

dexterbg:
I like the idea of simplifying the assembly of multi frame replies, but I agree it's not solved correctly in this pull request.

The assembler does not know about aborts or possibly interleaved replies from multiple buses, it guesses a new message start from the remaining length -- that will totally break on abort or bus context switch.

As Mark pointed out, there is no freeing of the buffer (think about large downloads like logs or blobs), and there also is no out of memory check -- if realloc fails, it will crash.

A frame assembler like this makes sense as a utility, but not as an automatic part of the current frame processing implementation.

It could be an option to add assembly buffers to the poll definition, so the poller can associate buffers to polls.

caederus-ovms:
Closing this in favour of request #199, which just adds the LEAF-specific BMS stuff.

PR #200, opened 2019-02-05
Chose between two SOH metrics.

caederus-ovms:
Report SOH from passive 0x5b3 in private metric "xnl.v.b.soh.instrument".
Report SOH from active 0x7bb.6101 in private metric "xnl.v.b.soh.newcar".
Standard metric "v.b.soh" copies from "xnl.v.b.soh.instrument" by default,
or from "xnl.v.b.soh.newcar" if config option "xnl.soh.newcar" is set to "true".


markwj:
This seems to have a merge conflict now. Can you resolve?

caederus-ovms:
Re-applied cleanly after update.

Issue #181, opened 2018-12-30
PnP: Read Leaf max gids automatically


arsharpe:
It appears that for 30 kwh (2016 and later) Leafs the maxGids needs to be set manually using
config set xnl maxGids 356

It would make the user configuration a lot easier (ie plug-n-play) if the code could detect to the battery size (eg from available capacity or a component of the VIN number) and set the GIDs accordingly.

caederus-ovms:
It should no longer be necessary to set xnl.maxGids to get a useful SOC reading.
The default behaviour is now to report the "instrument" SOC reading, as computed by the car internally and reported on the dash of newer models.
The older alternate method, calculating SOC as a percentage of energy in the battery relative to an arbitrary "new car" value, can be selected by: "config set xnl soc.newcar true" and "config set xnl maxGids ..." to a value of your choice.
To revert to the default behaviour: "config set xnl soc.newcar false".

dalathegreat:
@dexterbg please close this issue, no longer relevant

Issue #180, opened 2018-12-30
Leaf model year needs to be set manually


arsharpe:
It appears that the model year for 2016 Leafs needs to be set manually using
     config set xnl modelyear 2016

It would make the user configuration a lot easier (ie plug-n-play) if the code could detect to the model year (eg from status signals that were introduced at that year or any year info in the VIN number)

ltjessem:
The 10th character of the VIN tells the model year, starting from A (2010)
My 2015 is F, 2016 is G, 2018 is J . 

The battery pack size is also indicated in the VIN, for the "old" pre-2018 models the 4th character is A = 24kWh and B = 30kWh. On the newer 2018+ then A = 40 kWh, B will presumably be the 62 kWh

ankohanse:
Cool, that will allow auto detection on cars that have such a VIN. 

Unfortunately cars made for the Japanese Domestic Market and frequently imported 2nd hand into Australia and New Zealand do not use a VIN but instead have some sort of serial number in the form of ZE0-123456 or AZE0-123456. I have done extensive searches via Google, but have not yet found a reliable relation between such serial numbers and the model-year or battery size...

ltjessem:
Found this on a NZ forum, I'll look into this a bit more and see if I can figure something out

> My understanding for a 24 or 30 kw is the chassis number if it is AZE0 2xxxxx then it is a 30kw if it is AZE0 1xxxxx then it is a 24 but I may be wrong

ankohanse:
Posted a request on that forum (www.geekzone.co.nz) for people to email me their JDM style VIN combined with year and battery size. Hopefully that will give us some new insights...

I very much hope that can confirm that AZE0-1xxxxx is 24kwh, AZE0-2xxxxx is 30kwh, AZE1-1xxxxx is 40kwh and possibly AZE1-2xxxxx will be 62kwh    :-)

ankohanse:
My conclusions after receiving input from users on GeekZone and Facebook:

A. The JDM-style VIN cannot be used to predict battery size. Typically, the 2016 and 2017 models have both 24kwh and 30kwh cars in the same VIN range. See https://www.japan-partner.com/check-manufacture-year.php
 
B The JDM VIN can be used as a rough indication of model year:
 
ZE0-   2011-2012  (24kwh)
AZE0-0 and AZE0-1  2013-2015  (24kwh) 
AZE0-2    2016-2017   (24 and 30kwh) 
[...]

PR #204, opened 2019-02-21
Nissan Leaf: add web config pages for parameters and battery

ankohanse:
Added web pages for configuration of:
- Main parameters: modelyear and maxgids (as 24kwh or 30kwh selection)
- Battery: target charge goal as SOC % or as range in km

Pages are based on similar functionality for Renault Twizy and Kia Soul.

Note that this is related to to functionality requests #180 (auto detect model year) and #181 (auto detect max gids). Though the code in this pull request does not do an auto detect, it should be noted that at this moment nobody has described a robust way to do such an auto detect. This pull request will at least provide a more convenient way of configuring the main parameters other than via shell commands.


ankohanse:
Spotted there are annoying merge conflicts. More files have been changed by others than I anticipated while I was testing my code for a few weeks. Will get a new fork, reapply my changes and then create a fresh pull request...

PR #205, opened 2019-02-22
Nissan Leaf: add web config pages for parameters and battery

ankohanse:
Added web pages for configuration of:

* Main parameters: modelyear and maxgids (as 24kwh or 30kwh selection)
* Battery: target charge goal as SOC % or as range in km

Pages are based on similar functionality for Renault Twizy and Kia Soul.

Note that this is related to to functionality requests #180 (auto detect model year) and #181 (auto detect max gids). Though the code in this pull request does not do an auto detect, it should be noted that at this moment nobody has described a robust way to do such an auto detect. 

It is unlikely we will find a suitable method for auto detect of modelyear (even when based on VIN). So this config setting remains valid.
It is likely that we will get a reliable method to detect max gids / battery size (for instance via the battery voltages/temeratures). At that point we can adjust the config page to remove that setting (but leave the modelyear).

This pull request will for now at least provide a more convenient way of configuring the main parameters other than via shell commands.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/nl_types.h
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/nl_types.h
@@ -1,0 +1,41 @@
+/**
+ * Project:      Open Vehicle Monitor System
+ * Module:       Nissan Leaf
+ * 
+ * (c) 2017  Michael Balzer <dexter@dexters-web.de>
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


dexterbg:
Nice work, merged.

Have you seen @ltjessem's info on model detection?
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/issues/180#issuecomment-465882036

ankohanse:
Yes, have read it and responded to it in that item.
Autodetect method described there will work for some Leaf markets but not for others (Japanese Leafs do not use a VIN, but return some sort of unique serial number instead). So config page is still needed for those cars.

PR #212, opened 2019-03-10
Nissan Leaf: battery metrics for 30kwh

ankohanse:
On a 30kwh leaf, the metrics for GIDs, battery energy available and battery energy capacity were randomly switching from actual value to maximum value.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -469,16 +469,6 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
 
   switch (p_frame->MsgID)
     {
-    case 0x1da:
-      {
-      // Signed value, negative for reverse
-      int16_t nl_rpm = (int16_t)( d[4] << 8 | d[5] );
-      if (nl_rpm != (int16_t)0xffff)
-        {
-        StandardMetrics.ms_v_mot_rpm->SetValue(nl_rpm/2);
-        }
-      }
-      break;
     case 0x1db:
     {
       // sent by the LBC, measured inside the battery box
@@ -506,7 +496,7 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
           StandardMetrics.ms_v_bat_soc->SetValue(soc);
           }
         }
-      }
+    }
       break;
     case 0x284:
     {
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -698,6 +698,11 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
          *  ---- | -- | --  --  -- | --  --  -- |
          *     1 |  ? |  ?   9  17 |  ?  10  18*|
          *     2 |  0 |  5   8  11 | 18* 21  24 |
+         *
+         * Only type 1 and type 2 24kwh models from before 2016 will report a valid 'range 80%'.
+         * Any type 2 24 or 30kwh models starting mid 2015 (USA/Jap) or 2016 (UK), will always 
+         * return 0x1fff, and therefore never enter this if with mx values 18, 21 or 24.
+         * This is linked to Nissan removing the 'long life mode (80%)' from the car settings.
          */
         int cd = -1;
         switch (mx)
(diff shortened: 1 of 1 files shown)


vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
```diff
@@ -469,6 +469,16 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
 
   switch (p_frame->MsgID)
     {
+    case 0x1da:
+      {
+      // Signed value, negative for reverse
+      int16_t nl_rpm = (int16_t)( d[4] << 8 | d[5] );
+      if (nl_rpm != (int16_t)0xffff)
```
caederus-ovms: The 'undefined' value I see is 0x3fff (not 0xffff, which is a valid, if unlikely, value of -1).

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
```diff
@@ -713,7 +721,17 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
...
           }
         if (cd != -1) m_charge_duration->SetElemValue(cd, val/2);
-        if (type != -1) m_battery_type->SetValue(type);
+        }
+      // If detected, save battery type
+      if (type != -1)
+        {
+        m_battery_type->SetValue(type);
+        }
+      else if (!m_battery_type->IsDefined())
```
caederus-ovms: Not sure if the IsDefined() test still make sense.  It was previously needed to account for the additional information that might have turned up independently from the battery poll reply, but we don't have to worry about that now.
I think we can instead set type=BATTERY_TYPE_2_24kWh in switch(mx) cases 21 and 24, since TYPE_1_24kWh uses mx 10 and 18, and I believe your TYPE_2_30kWh never has valid values for either of these?

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
ankohanse: This change was a result of a  commit done AFTER I created the pull request. Had not expected it to be then automatically included in the existing pull request. Have reverted this change and will create a new pull request for it (using 0x3ffff) once this one is merged.

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
ankohanse: Prefer to leave it as it is now. 
My Japanese TYPE2_30kwh has messages for mx cases 21 and 24, always with the invalid value 0xfff. I expect US models to behave the same. But I cannot yet rule out that UK 30kwh versions might show a valid value here. If so, then setting TYPE2_24kwh on these mx cases would be incorrect.

So for the moment I think it is safer to set a default of TYPE2_24kwh and only change it when we positively identify a TYPE1_24kwh or a TYPE2_30kwh.

ankohanse:
Did another google to figure out what the expected behaviour of 30kwh UK models would be.
Found 2 pages both indicating that the "long life mode (80%)" was removed from USA and Jap models somewhere in 2015 and from UK models as of mode year 2016.
    https://www.speakev.com/threads/long-life-mode-80-charge-no-longer-available.8261/   
    https://samholford.github.io/leafguide/

So Robin is probably right and if we see a val!=0x1fff with mx=21 or 24 then this must be a type 2 battery from before 2016, which can only be a 24kwh.
Adapted the code accordingly and added comments describing this behavior.

Note that we do not have a way to positively identify a TYPE_2_24kwh from 2016 or 2017.  So for the moment we still rely on using TYPE_2_24KWH as default when reading values from m_battery_type.

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
ankohanse: Did another google to figure out what the expected behaviour of 30kwh UK models would be.
Found 2 pages both indicating that the "long life mode (80%)" was removed from USA and Jap models somewhere in 2015 and from UK models as of mode year 2016.
    https://www.speakev.com/threads/long-life-mode-80-charge-no-longer-available.8261/
    https://samholford.github.io/leafguide/

So Robin is probably right and if we see a val!=0x1fff with mx=21 or 24 then this must be a type 2 battery from before 2016, which can only be a 24kwh.
Adapted the code accordingly and added comments describing this behavior.

Note that we do not have a way to positively identify a TYPE_2_24kwh from 2016 or 2017. So for the moment we still rely on using TYPE_2_24KWH as default when reading values from m_battery_type.

PR #226, opened 2019-05-14
SmartED: complete some missing metrics.

Dimitrie78:
add config page
add BMS Cell Monitor
add charging Status


dexterbg:
Dimitrie, your commit history is a mess, and you included a change to the charts.js.gz file, that just unnecessarily adds CR line endings to the enclosed JS files.

Please do not add commits like "some test" or "short fix". Descriptions like that won't help anyone trying to understand your submissions or track down bugs. There also are very few legit reasons to add "test" commits to the master history. Use local test branches, stashes or additional clones for tests, commit working results.

Please avoid including your .gz rebuilds unless you actually did change the sources.

dexterbg:
Just saw there are even german commit descriptions in there: please also avoid those.

Dimitrie78:
Thanks for the suggestion.
It's my first time working on a project.
the commits "some test" or "short fix" will i avoid next time.

PR #230, opened 2019-05-26
SmartED

Dimitrie78:
improve charging status and add metrics for charger can poll


dexterbg:
You should rework your metrics names to follow the namespace rule of prefixing custom metrics with the vehicle specific ID, i.e. instead of `v.c.active` you would use `xse.v.c.active`.

This is meant to avoid conflicts with future framework extensions. If the framework adds a metric `v.c.active`, your code will break.

`me li v.c.` lists both framework and custom metrics of path `v.c.`.

Dimitrie78:
thanks, i'll change them. that make sence.

PR #234, opened 2019-06-04
OVMS A5 flyer

glynhudson:


dexterbg:
The Twizy features on the back page are a bit outdated, I've added the current feature set to the V3 manual.

glynhudson:
Thanks, I've now submitted a PR to remove the rear compatibility table. It's already outdated https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/235

PR #242, opened 2019-07-19
feat(vehicle): hyundai kona ev

birkir:
Forked Kia e-Niro and updated the VIN decoder and charging steps

I think the car is mostly the same, but I think it's better to keep a separate copy.

**Untested**

Paging @goev and @markwj


dexterbg:
Besides renaming the vehicle I see only minor differences between the Kia and the Hyundai code, and the charging profile change could even be valid for the Kia as well (basically adding more detail).

In my eyes this is a case of introducing a user configurable subtype instead of adding a new vehicle class.

Separating the vehicles should only be done with a good reason, as it will add effort to maintaining both vehicles. They obviously share the same base. Can you please elaborate on your opinion to keep this as as separate copy?

glynhudson:
Interesting, are there *any* technical changes required between Kona / Niro Ev's? Will OVMS setup for Niro just work with a Kona? If this is the case the name in OVMS could just be renamed to Niro/Kona and graphics added for Kona to the OVMS mobile apps.

birkir:
Hyundai Kona and Kia Niro are essentially the same.

**But** they use different battery vendors

Kia e-Niro uses SK Innovation
Hyundai Kona uses LG Chem

They also have different features around charging and preserving energy that may need some additional code.

I think creating a shared class for these cars is alright

glynhudson:
Ah ok, that's interesting. Over the weekend I tested OVMS myself on a friends Kona and it seemed to work perfectly with OVMS set to 'Niro'. 

>They also have different features around charging and preserving energy that may need some additional code.

What features are these? Could they be incorporated as options in the Niro vehicle? This sounds like a perferable option to avoid code duplication. Would you like me to work on adding graphics for Kona to the Android app? I'm keen to see the Kona supported by OVMS

PR #247, opened 2019-08-10
Include GPS sat count in location status report

leres:
and use plural format for locations line


dexterbg:
Well… ok, you asked for it: singular with "are"… ;)

PR #249, opened 2019-08-11
Add some hints about how to create /store/plugin

leres:
... in the Installing Plugins section of the Web Framework & Plugins doc.

This was confusing to me when I first attempted it because the directory doesn't exist.


dexterbg:
The directory is automatically created as necessary by the editor when you save a plugin.

leres:
I guess what happened in my case was first I tried scp'ing it there. Since the directory didn't exist it created a file called plugin. Once that was there I couldn't create any plugins using the web interface.

At least I learned how to build a local copy of the html documentation!

PR #270, opened 2019-10-14
Zoe: most metrics added

Dimitrie78:
charging status, soc, odo, speed, trip, HV Volt, HV current, LV Volt, LV Current and some more


dexterbg:
Please don't forget to update `changes.txt`

glynhudson:
Any info on what versions of Zoe are supported? Is ZE40 supported for just older 24kWh versions?

Dimitrie78:
except ze52 should all go there all the same except on battery. but not tested at all.

PR #304, opened 2019-12-17
Add 62kWh Leaf GIDS and AH defines

dalathegreat:
No UI changes yet, but it's good to have the defines ready. We'll be seeing 62kWh adopters of this app soon, especially since the Nissan app works so "intermittently" to say the least.


glynhudson:
That's great, nice work. Is the 40kWh LEAF supported?

dalathegreat:
> 
> 
> That's great, nice work. Is the 40kWh LEAF supported?

I added the Gen2 leaf to the android app, so it's selectable in the UI atleast. The 40kWh defines are in OVMS webinterface, so it should be possible to start using it already, but the remote heating doesn't work yet.

PR #307, opened 2019-12-21
Set the Vehicle ID as the hostname for the DHCP request parameter

markfinn:
This will make DNS in the DHCP server (as seen by tools like Fing) match what is seen in MDNS

Currently this shows up as "espressif" which is annoying if you have let the Internet-of-things take over your house and have many poorly named ESP devices.

### DIFF
--- a/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
+++ b/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
@@ -863,10 +863,12 @@ void esp32wifi::EventWifiStaState(std::string event, void* data)
   {
   if (event == "system.wifi.sta.start")
     {
-      // Set hostname for DHCP request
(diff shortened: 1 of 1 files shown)


dexterbg:
Mark, welcome to the OVMS project & thanks for your submission!

I'd like to merge this, but I've got two minor issues you could resolve:
- We've got mixed indentation and block styles in the project. If you work on a module, please stick to the code style of that module. In this case, block brackets are indented, block content does not get another indentation beyond that (align to `AdjustTaskPriority()`).
- `ESP_ERROR_CHECK()` should not be used for non-critical calls, as it does an `abort()` on failure (i.e. crash & reboot). Check the return value instead and do a warn or error level log call. (And yes, there certainly are some more places in the module that need rework on this… copy&paste remains from the esp-idf templates)

Thanks!

markfinn:
Will do.  And copy and paste is how I made that line too!

PR #308, opened 2019-12-22
Update README file with more battery specs

dalathegreat:
- Added more sections about 40/62kWh LEAFs, 
- Tidied up the document a bit too, now the battery section comes last, and model year first.
- Added link to .DBC files


dexterbg:
Daniel, this change apparently has rst syntax errors at least with the headlines, haven't checked further. Please test documentation changes before submission by doing a local ``make html``, check for sphinx errors and check the result in your browser.

dalathegreat:
@dexterbg Even when I checked out the latest master it doesn't compile without warnings, same as on my branch
image

dexterbg:
You're probably the only one here working on a Microsoft platform… from the errors I would assume there is something wrong with the symlinks. I remember having read about issues with symlinks on MS somewhere, possibly in git context…

…ah, yes, found some threads on this on SO, maybe you should try this:
https://stackoverflow.com/questions/5917249/git-symlinks-in-windows#16754068

PR #316, opened 2020-01-02
quiet "maybe-uninitialized" error in ducktape.c

leres:
(the inline version of duktape that pulls in duk_js_call.c)


dexterbg:
Craig, you should submit the fix to the Duktape repository as well, so we don't pull it in again with the next update.

leres:
> Craig, you should submit the fix to the Duktape repository as well, so we don't pull it in again with the next update.

Done: [_quiet "maybe-uninitialized" error in src-input/duk_js_compiler.c #2218_](https://github.com/svaarala/duktape/pull/2218)

PR #322, opened 2020-01-11
Add smart ED Gen 3 eco score system

frantek:
Add access to eco scores similar to the old retired smart web app via web interface. Scores are reported at the end of the trip as well


dexterbg:
@frantek Please also announce user level improvements like this in the `changes.txt`.

frantek:
@dexterbg Will do next time, got some more changes in the pipeline

PR #366, opened 2020-05-04
Initial support for Mercedes B250E

r151809:
Passivel listener for CAN1.
Decodes only small set of messages, for example momentary power and ODO are found, but no SOC.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_mercedesb250e/component.mk
+++ b/vehicle/OVMS.V3/components/vehicle_mercedesb250e/component.mk
@@ -1,0 +1,14 @@
+#
+# Main component makefile.
+#
+# This Makefile can be left empty. By default, it will take the sources in the
+# src/ directory, compile them and link them into lib(subdirectory_name).a
+# in the build directory. This behaviour is entirely configurable,
+# please read the ESP-IDF documents if you need to do this.
+#
+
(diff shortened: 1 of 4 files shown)


dexterbg:
Welcome to the show, Jarkko :-)

Please remember point 1:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3#development-and-contributions

Regarding your submission, that's already nearly perfect. User documentation is lacking, but it's at an early stage and not yet enabled by default, so that's ok.

One thing: your custom metrics name space is the same as for the Smart ED. You should exchange the "xse" prefix by something specific, e.g. "xmb".

I'd also say version "1.0.0" shouldn't be assigned to an early stage of development like this, but that's up to you.

I'll merge this now.

dexterbg:
I've enabled the vehicle in my edge build, it's included in the current build 3.2.012-216-g89ee89ca.

r151809:
I'll write the introduction email in the evening. I guess that the mailing list would be the correct place to ask the following questionsa too :) I'll keep them here too so I do not forget..
1. Should I test the edge build also with my car, or shall I just continue the development in my own fork?
2. What's the intention of None vehicle? I assumed that it would be a template. In that case I would like to add more comments there for new developers (like me).

PR #370, opened 2020-05-13
Mercedes B250e - Still just listening CAN. Also doc update.

r151809:
Hi,

Edit: Consumption is still wrong - so I'll to do more "just listening" updates. Please reject this pull request if you do not want to merge and review too many updates.

Added index.rst (and related metrics.rst). Should it be added to some makefile, or is the .rst=>.html conversion automatic?

Metrics are improved and now I managed to do even some testing. Still very alpha level and the most important metric, SOC, is still hiding from me. In my opinion the level of functionality could be already interesting for a B250E owner, but hardly yet useful :)

There is one merge commit from the openvehicles repo. I could not figure, how to mask it. It should not show up in diff, but in commit list yes.. For the next development I'll try make it all in a feature branch, so this should not happen again.


dexterbg:
Jarkko,

see http://lists.openvehicles.com/pipermail/ovmsdev/2019-July/006196.html on how the documentation system works and how to build the html version locally.

If you want to build locally, check the git log for changes to the configuration, we added some plugins later on.

Regards,
Michael

r151809:
Thank you :)

PR #377, opened 2020-06-06
Make some metrics persistent across warm reboots

leres:
Use the RTC_NOINIT_ATTR attribute to put some metrics in RTC slow memory
These persist across crashes, firmware updates, and warm reboots and
are zero'ed on cold reboots
This is only implemented for OvmsMetricFloat
Unrelated fix: puts() -> writer->puts() in metrics_list()

### DIFF
--- a/vehicle/OVMS.V3/main/metrics_standard.cpp
+++ b/vehicle/OVMS.V3/main/metrics_standard.cpp
@@ -67,7 +67,7 @@ MetricsStandard::MetricsStandard()
   ms_v_type = new OvmsMetricString(MS_V_TYPE);
   ms_v_vin = new OvmsMetricString(MS_V_VIN);
 
-  ms_v_bat_soc = new OvmsMetricFloat(MS_V_BAT_SOC, SM_STALE_HIGH, Percentage);
+  ms_v_bat_soc = new OvmsMetricFloat(MS_V_BAT_SOC, SM_STALE_HIGH, Percentage, true);
   ms_v_bat_soh = new OvmsMetricFloat(MS_V_BAT_SOH, SM_STALE_HIGH, Percentage);
   ms_v_bat_cac = new OvmsMetricFloat(MS_V_BAT_CAC, SM_STALE_HIGH, AmpHours);
   ms_v_bat_health = new OvmsMetricString(MS_V_BAT_HEALTH, SM_STALE_HIGH);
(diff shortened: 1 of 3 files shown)


dexterbg:
Nice work. Please add a brief user level description to the `changes.txt`.

I also suggest making the inverter and charger temperatures persistent, and possibly the range estimation, as that normally is read from the vehicle. And the energy counters (`v.b.energy.used`, `v.b.energy.recd`)… guess you can raise the slot count to 20 to do this, the boot data takes only 268 bytes currently.

Note that vehicles using internal state variables being published to metrics need to add some init code to prime the internal state from metrics on boot to utilize this. Or use RTC memory for their internal core state as well.

dexterbg:
Internal state init for the Twizy:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/commit/b268b11bd08d3b898d90c4d2269a91b3cf850dcb

leres:
On 2020-06-06 23:31, Michael Balzer wrote:
> Nice work. Please add a brief user level description to the |changes.txt|.
> 
> I also suggest making the inverter and charger temperatures persistent,
> and possibly the range estimation, as that normally is read from the
> vehicle. And the energy counters (|v.b.energy.used|, |v.b.energy.recd|)…
> guess you can raise the slot count to 20 to do this, the boot data takes
> only 268 bytes currently.

Honestly I think that using RTC memory is overkill; I believe it's to park things during deep sleep which I don't think we do. I suspect all ram is always powered for us and so we could use as much as we want this way with regular ram. It's just a matter of putting it a memory section that is not initialized on boot and also in a section that is low in the address space to minimize the chances that the base address would shift due to the addition or removal of global variables across versions.

New [pull request](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/378) with various improvements.

Let me know if you want more for either changes.txt or the user guide. (One thing I noticed is that "metrics list -s" doesn't appear to be documented.)

> Note that vehicles using internal state variables being published to
> metrics need to add some init code to prime the internal state from
> metrics on boot to utilize this. Or use RTC memory for their internal
> core state as well.

The code I added tries to be careful and error on the side of not providing garbage. The struct has a magic number, version number, and size. If any are wrong on boot the entire struct is zero'ed freshly initialized.

dexterbg:
Deep sleep is an open todo (to cut down the 12V energy use of the module). We've got 8K of RTC slow RAM and currently nothing better to do with it. Btw, the esp-idf NVS lib plans to support RTC RAM as well in the future.

We should be able to fit all current metrics data into RTC RAM using an indexed structure. That may actually be necessary for deep sleep, I wouldn't want to wear down flash memory for this.

My main doubt about RTC RAM is the slow clock (150 kHz), and it possibly additionally being tagged as volatile for being shared with the coprocessor (may be not the case as we haven't enabled the ULP, haven't checked that).

I've got no issue reports yet but suggest using the m_value field as a read cache to minimize negative effects from this.

leres:
I'll take a look at using m_value as a read cache.

PR #388, opened 2020-07-09
Minor abrp plugin updates

inf0mike:
Fix for rounding of fractional SOC causing abrp to report SOC off by 1
Fix for altitude never being sent
New convenience method to reset config to defaults


dexterbg:
I suggest also bumping the plugin version to 1.2.1 or 1.3 to reflect the changes.

inf0mike:
> I suggest also bumping the plugin version to 1.2.1 or 1.3 to reflect the changes.
I will do that now.

Issue #386, opened 2020-07-07
Wifi Config Enhancement


drc38:
I'm looking to use Ovms via wifi only, with wifi going between two home networks and mobile hotspot plus potentially others. The current framework appears to be quite limited in terms of setting wifi client defaults and handling switching of networks when signal is lost. 

Often ovms will report good wifi signal but show no ip/dns/gateway etc in the status panel.

Would it be possible to extend management options/config by including a wifi manager such as esp32-wifi-manager in the build?

dexterbg:
Yes, that would be possible. Or you could simply add the functionality you're missing to the OVMS wifi manager.

PR #395, opened 2020-07-21
wifi enhancement #386

drc38:
add functionality to set static ip etc, and revert back to dhcp if required


dexterbg:
Derek, thanks for your submission.

After reviewing your changes I think this needs a little refinement:

- The command design is in conflict with our usual command style and counter intuitive: to switch back to DHCP mode, you still need to provide IP, netmask & gateway address first. Split these two functions in two sub commands, i.e. `wifi ip static` and `wifi ip dhcp`, the latter not needing any parameters. You can use the same handler for both and detect the operation by checking the command name (see e.g. `can_start()` for an example).
- `SetSTAStaticIP()` should also do what the name implies. Either change the name to a generic one and add an overload for a single bool parameter, or add a separate method to switch back to DHCP mode.
- `SetSTAStaticIP()` just checks for not being in AP mode, that is most probably not correct, as wifi could be off or currently doing a scan (manual scans are indicated by a temporary mode change). The correct check would be for CLIENT or APCLIENT mode.
- The else branch needs to check the mode as well, as it would otherwise start the DHCPC in any mode.

These issues need to be solved for the merge.

Additionally: I haven't tested the effect of setting a static IP and then changing the network, but suggest this setting should be bindable to an SSID configuration. It shouldn't need an explicit command, but automatically be applied when (re-)connecting to an SSID configured for static IP. That could also help with the mDNS entry, which could otherwise need some time to switch from the previously assigned DHCP address.

The `wifi.ssid` config param is currently defined to have a single entry per SSID containing the password. You can e.g. add a second entry by suffixing the SSID with something not normally used as part of an SSID, e.g. ".ovms.staticip", and store the combined "<ip>,<sn>,<gw>" data in it.

The web UI wifi config doesn't necessarily need to be extended for this option as that's more a power user feature, but you should explain the configuration in the user guide.

Regards,
Michael

drc38:
Cheers Michael

The Dhcp restart was a bit of an afterthought so I did not have to restart while testing. So will split into separate function and take care of other bullets below.

Setting the params per ssid is where I wanted to get to, my lack of C++ coding exp/understanding of how to do so with ovms params means I'd appreciate a bit more advice on how to define multiple entries ie for each ssid and then how to reference them with current ssid.

What's best way to update pull request? I may need a couple of iterations of your review?

Once happy with code I'll update docs to reflect the code changes in separate pull.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dexterbg:
You can keep the PR open and simply push more commits to your PR branch "static-ip" to update the PR, then post a message here when you'd like another review.

You can create a config entry by `config set`. If the SSID is named `FooBar`, you can e.g. create an IP config following my proposal by doing `config set wifi.ssid "FooBar.ovms.staticip" "192.168.0.42,255.255.255.0,192.168.0.1"`.

To get a config instance for an SSID, as we're using `std::string` for the instance name, you can simply add the suffix:
```
  std::string ssid; // or method parameter
…
  // read static IP config:
  std::string ipconfig = MyConfig.GetParamValue("wifi.ssid", ssid + ".ovms.staticip");

  // parse static IP config, pattern "<ip>,<nm>,<gw>":
  std::string ip, nm, gw;
  std::istringstream sbuf(ipconfig);
  std::getline(sbuf, ip, ',');
  std::getline(sbuf, nm, ',');
  std::getline(sbuf, gw, ',');
…
```

A little inconvenience: `wifi.ssid` is currently read protected, so you can't verify your config by `config list wifi.ssid`. But there is a plan on moving all passwords into the `password` param, so we can then remove the read protection from `wifi.ssid`.

Check out https://en.cppreference.com/w/ for a good C++ reference.

Regards,
Michael

drc38:
Hi Michael, if you could check the revised code. I will then look at binding to a SSID configuration. Cheers

dexterbg:
Derek, you made a mistake with your commit. It looks like you had a stash conflict you didn't resolve:

```
…
<<<<<<< Updated upstream
…
=======
>>>>>>> Stashed changes
…
```

These are the merge conflict markers. They must not be checked in, you need to use a merge tool or resolve conflicts manually before committing the result.

You can still solve the conflicts manually now and then push the result. Check all conflict markers in the changed files and pick the parts that shall remain.

A good git UI (e.g. "git gui") alerts you about unresolved conflicts from a stash-pop or merge. There's a wide range of three way merge tools you can use, I recommend "kdiff3". Duck "git merge tools" for more recommendations.

Regards,
Michael

drc38:
Not sure what happened to create the conflict as I could not find anything stashed... should be fixed now though

drc38:
Using the SSID in the param has the unwanted side effect of being pulled through to Wifi interface, will this cause an issue?
image

dexterbg:
Looks much better now.

Things to change:

- `StartDhcpClient()` should check the mode as well to be client or apclient.
- Code indentation style in `StartDhcpClient()`.
- Update the Wifi web editor: as you already noticed, you need to at least hide the staticip param instances from the UI. A better option would be to add the necessary elements & handling to properly support them.

Further notes:

- You didn't need to remove the parameters from the command in favor for the config. We normally use the config for automatic setups but still allow manual setups through the commands. By removing the parameters you now cannot set a static IP temporarily. A better implementation would allow parameters for manual/temporary configuration, and w/o parameters fall back to the config if defined.
- Did you test the DNS being remembered correctly by the network manager? `tcpip_adapter_set_ip_info()` triggers the "STA_GOT_IP" event, which triggers the DNS info to be saved by the netman. So there might be a race condition as you call `tcpip_adapter_set_dns_info()` after `tcpip_adapter_set_ip_info()`…?

Regards,
Michael

drc38:
Thanks Michael,

Hopefully the code is close to being ready to include, responses to your feedback above:
 - StartDhcpClient() should check the mode as well to be client or apclient. **Done**
 - Code indentation style in StartDhcpClient(). **Done**
 - Update the Wifi web editor: as you already noticed, you need to at least hide the staticip param instances from the UI. A better option would be to add the necessary elements & handling to properly support them. **Not sure how to tackle this one, would you be able to assist?**
- A better implementation would allow parameters for manual/temporary configuration, and w/o parameters fall back to the config if defined. **Reinstated option to supply parameters from command line and added extra logic**
 - Did you test the DNS being remembered correctly by the network manager? **Changed to setting dns configuration parameter to avoid potential race, I did not have issues with previous approach but this approach allows netmanager to take care of dns** 

If the structure of the two additional commands are acceptable I will work on updating the docs with their usage.

Cheers Derek

dexterbg:
Derek,

again a good step forward, but also a little one backward.

- `wifi_static_ip()` doesn't check `argc` before using `argv[]`: as the command registration allows 0-3 arguments, you need to check this or giving just 1-2 parameters would trigger an out of bounds array access.
- The command info description also no longer matches the functionality.
- `StartDhcpClient()` and `SetSTAStaticIP()` must not change the configuration, especially not the network DNS config, which applies to all networks. But they also should not change the static IP config. There are very few cases where changing the config from a non-config command is acceptable, e.g. to aid the user in creating/maintaining complex configurations or to store some persistent data. In this case, your new commands and methods should only reconfigure the current operation mode, but leave the persistent config up to the user.
- `tcpip_adapter_set_dns_info()` would now be called before `tcpip_adapter_set_ip_info()` (correct), but is commented out…?

Regards,
Michael

drc38:
Hi Michael,

With respect to items above:
1.  Thanks for pointing out the out of bounds issue. That can be easily resolved.
2.  Not sure what the issue is with the description can you clarify?
3.  As it was going to be for advanced use I was following a similar logic to "fixed" vs "scan" mode. That is with `SetSTAStaticIP()` the details are set/fixed and with `StartDhcpClient()` they are cleared. Probably a better approach is to introduce a new param eg `<ssid>.ovms.static` which can be set to true/false. The `StartDhcpClient()` would no longer be a user command as this would be handled in `SetSTAStaticIP()` based on the setting. True to use static details set and false to use dhcp client. Would that work for you?
4.  To follow the same principle with DNS I can revert to the direct call to `tcpip_adapter_set_dns_info()` before `tcpip_adapter_set_ip_info()`. It may require some extra code from netmanager to update DNS information displayed in the Status page to be added.

Let me know if this revised approach will be suitable.
Cheers Derek

dexterbg:
Hi Derek,

- The command argument description currently does not follow the convention of giving the argument template on the first line, and it implies you need to set the config before using the command. A good description for the current function could be:
```
  cmd_mode->RegisterCommand("ipstatic","Set static IP mode",wifi_static_ip,
    "[<ip> <subnet> <gateway>]\n"
    "Example: 192.168.12.34 255.255.255.0 192.168.12.1\n"
    "The gateway will also be used as the DNS.\n"
    "Omit arguments to read the configured setup for the current SSID:\n"
    "Use config set wifi.ssid \"<ssid>.ovms.staticip\" \"<ip>,<subnet>,<gateway>\"\n"
    "to configure static details for an SSID.\n", 0, 3);
```
- The single config instance `<ssid>.ovms.staticip` is sufficient. If it's set, the SSID needs the static IP setup. If not, it's a DHCP network. The system should work automatically that way, but also allow to manually change the current network setup. You just need to remove changing the user configuration from these commands to get there.
- DNS info isn't transported to the netmanager via the config store, the netmanager reads that info directly from LwIP (see `OvmsNetManager::SaveDNSServer()`). So if you add the `set_dns_info()` call, everything should be fine.

Regards,
Michael

drc38:
Hi Michael,

Hopefully this makes more sense. I've moved the underlying logic to test whether to use function parameters or `config set wifi.ssid` to `wifi_static_ip`. The renamed function `SetSTAWifiIP` then simply processes the details it receives or calls `StartDhcpClient` if `staticmode` is set to false.

`tcpip_adapter_set_dns_info` works but does not update the dns details on the web interface. `dns_setserver` resolves this but setting it to position 0 may not be ideal. If possible using `SetDNSServer` may be better?

Regards Derek

dexterbg:
Hi Derek,

first of all, you added `ignore = dirty` to the mongoose submodule declaration in this PR. That may make sense for you locally & temporarily, but certainly not for any developer working on the mongoose part. Even if it would do so, it would still need a separate pull request explaining why it's a sensible change, as this has no connection to the wifi enhancement.

Now for your changes :-)

OK, you now basically kind of came back to my first command structure proposal of having commands `wifi ip static …` and `wifi ip dhcp`, but with free arguments. The advantage of my proposal using sub commands is, the shell can expand the commands (it cannot do so with arguments), can give the list of available sub commands and can verify the number of arguments. You can still handle both commands by the same function, which simply checks the command name to know which command to process. See sub commands as fixed arguments. That's a standard pattern you can find throughout the system, so we should stick with that.

We also almost never use `true/false` for semantics like this, but instead `on/off`, `yes/no` or `start/stop`, depending on the context, as these terms are clear and understandable for users. In this case, `static` (or `fixed`) and `dhcp` are perfectly clear for a user.

Your refactoring of the config check & parsing to outside the `SetSTAWifiIP()` method doesn't seem to make sense. You now need to do the same config parsing twice, once in the shell command and once in `EventWifiStaConnected()`. Also, `SetSTAWifiIP()` now again is semantically overloaded by also doing the DHCP switch… for which it is simply a wrapper call. In all use cases, the client already knows which method to call, so directly calling `StartDhcpClient()` would be clearer & more readable. We already had that point: APIs shall be semantically simple, clear & unambiguous. I suggest you change that back to having a dedicated method that only does the static IP setup.

Regarding the DNS change: `SetDNSServer()` is a protected method of the network manager, meaning it is not allowed to be called from outside the class. The compiler would have told you if you had tried.

You're also not supposed to normally call `dns_setserver()`, as `tcpip_adapter_set_dns_info()` is the validating wrapper API call for `dns_setserver()` -- see source. So functionally there is no reason why `dns_setserver()` should work, _except_ it actually _is_ necessary to call it after `tcpip_adapter_set_ip_info()`, as you now do with `dns_setserver()`.

Looking into the documentation: 
```
 *   If DHCP client is enabled, main and backup DNS servers will be updated automatically from the DHCP lease if the relevant DHCP options are set. Fallback DNS Server is never updated from the DHCP lease and is designed to be set via this API.
 *
 *   If DHCP client is disabled, all DNS server types can be set via this API only.
```
So it's plausible to assume you first need to configure a static IP, or the call may just be ignored or the DNS get overwritten afterwards. If that really introduces a race condition with the network manager event processing (does it?), we may need to solve that by changing the event processing.

Regards,
Michael

PS: I would understand if you're growing some frustration now. If you'd like to, I can make the changes to your code to adapt it to our framework style. But I also think you're learning from our session.

drc38:
Hi Michael,

I meant to explain the .gitmodules update in my last message. I had stashed changes to work on another branch, but when I came back to this one git hub would not allow me to pop/restore the changes because of the "dirty" mongoose sub-module. I tried a few recommendations using the git command line from the net but they did not work. Even a local change to ignore the dirty condition was not sufficient and hence the commit. Let me know if there is another way in case I have the same issue in the future. I assume the commit can be ignored at your end? Equally it is ok to keep my branch up to date with periodic upstream merges?

I am happy to keep working through the issues as I am learning the OVMS framework, GitHub and C/C++ as I go. I  will take a look at some more of the reference code and come back to you with questions.

Cheers Derek

dexterbg:
Derek,

(your commit comment "Update .gitmodules" is also a good example of a bad commit comment -- fully redundant ;-))

I'm not sure I understand your git issue. Github doesn't have dirty branches or stashes, git does. I assume you forgot some submodule update when changing branches. The submodule checkout status are coupled to the main checkout status. When you checkout a branch or commit and see some open changes from submodules, you normally do `git submodule update --recursive` to fetch & checkout all submodules recursively so they match your main commit point. If you don't, git will think you actually want to add the changed submodule states to your next commit.

Stashes btw are stored separately from the branches, you can pop/apply any stash on any branch. That's very useful for transferring conflicting changes between branches.

The commit is now part of the pull request, which can only be merged completely. But you can do a `git revert` of that commit on this branch, which simply undoes the changes of that commit and stores that undo as a new commit.

Regards,
Michael

drc38:
Thanks for explaining Michael. I am using GitHub Desktop which, in general, I have been assuming takes care of updating the submodules, but perhaps not without using the command line. Following your suggestion I reverted the commit and ran  `git submodule update --recursive` in the OVMS.V3 folder - nothing is reported when it runs - and back in GitHub Desktop it is still showing the submodule as "dirty" (see screenshot below).

I could use this process as a workaround - that is commit the ignore change to .gitmodules, pop/restore my stashed changes and then revert commit - but it is not ideal, so if you have any other suggestions to fix it let me know.

image

drc38:
Hi Michael,
I went back to using `tcpip_adapter_set_dns_info()` instead of `dns_setserver()` and it is behaving in the same way - reporting the dns set in the web interface as it should. I will keep an eye on it in case there is occasionally a race condition that creates unexpected behaviour.

Having spent more time with the interface and code I now understand your comment on using sub commands `wifi ip static` and `wifi ip dhcp`. Unfortunately I didn't get it when I was starting out ;) hence the potential disconnect...
 
I'll revert back to having the config check and parsing to within the main function. 

If we're not there after the next commit I'll take you up on the offer of help with the code ;D
Cheers Derek

drc38:
Hi Michael,
There is still the issue of the wifi.ssid config parameter appearing in the web interface. Can you help with this?
Cheers
Derek

slcasner:
On Sun, 2 Aug 2020, Michael Balzer wrote:

> OK, you now basically kind of came back to my first command structure proposal of having commands `wifi ip static ...` and `wifi ip dhcp`, but with free arguments. The advantage of my proposal using sub commands is, the shell can expand the commands (it cannot do so with arguments), can give the list of available sub commands and can verify the number of arguments. You can still handle both commands by the same function, which simply checks the command name to know which command to process. See sub commands as fixed arguments. That's a standard pattern you can find throughout the system, so we should stick with that.

I second Michael's recommendation to use subcommands.  Sometimes this requires a token order that seems rather convoluted, though.  To address that a while back I extended the command parsing to allow a subcommand token to follow a free argument with appropriate validation of the free argument.  I wanted this for the "location" command (in components/src/ovms_location.cpp) which is probably the only instance of its usage so far.

I'm not asserting that this is needed for the command syntax currently under discussion, but wanted to point out that flexibilty if needed.

-- Steve

dexterbg:
Derek,

regarding your "GitHub Desktop" issue, I don't know that, I've always used native git / git gui / gitk. If you tell me the steps / git commands to get to that dirty tag, I can try to reproduce that here. I suggest checking the "GitHub Desktop" issues. A quick duck reveals: https://github.com/desktop/desktop/issues/8496

Regarding your code, that's nearly perfect now, nice work. Just one little bug remaining: if a config is defined, `SetSTAWifiIP()` will discard manual parameters unconditionally.

I'll merge now and take care of that. I'll also have a look at the web UI.

Regards,
Michael

dexterbg:
Ah, another little thing: I think this is enough of an enhancement for users to add a note to the `changes.txt`. I'll also take care of that.

PR #404, opened 2020-08-13
Increased Poller frequency

SokoFromNZ:
Currently the Poller sends a maximum of one poll per second/tick. Even when there are more than one poll to do in the particular second/tick.

Example for `OvmsVehicle::poll_pid_t vwup_polls[]`:        

    {VWUP_CHARGER_TX, VWUP_CHARGER_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_CHARGER_AC_U, {0, 0, 5}},
    {VWUP_CHARGER_TX, VWUP_CHARGER_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_CHARGER_AC_I, {0, 0, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_U, {0, 1, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_I, {0, 1, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_SOC, {0, 20, 20}},

The first 4 polls happen 1 second apart (taking 4 seconds in total) and the SOC poll comes in with a rough frequency of 42 seconds instead of 20.

I was expecting the first 4 polls go as fast as possible after each other and the SOC of course every 20 seconds.

My changes do exactly that but leaving the standard behaviour as is.

If active `PollerSend()` will be called immediately after the last incoming Poller reply was processed and no more data for this reply is expected.

If someone wants to active this in the vehicle-implementation just call `PollerSetState(state, false);`. So this behaviour can be conveniently changed with the Poller state.


SokoFromNZ:
Sorry, I'm kinda new to Git and all this weird procedures. Apparently I have to do branches for different pull request...

PR #405, opened 2020-08-13
Faster Poller option when needed

SokoFromNZ:
Currently the Poller sends a maximum of one poll per second/tick. Even when there are more than one poll to do in the particular second/tick.

Example for `OvmsVehicle::poll_pid_t vwup_polls[]`:        

    {VWUP_CHARGER_TX, VWUP_CHARGER_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_CHARGER_AC_U, {0, 0, 5}},
    {VWUP_CHARGER_TX, VWUP_CHARGER_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_CHARGER_AC_I, {0, 0, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_U, {0, 1, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_I, {0, 1, 5}},
    {VWUP_BAT_MGMT_TX, VWUP_BAT_MGMT_RX, VEHICLE_POLL_TYPE_OBDIIEXTENDED, VWUP_BAT_MGMT_SOC, {0, 20, 20}},

The first 4 polls happen 1 second apart (taking 4 seconds in total) and the SOC poll comes in with a rough frequency of 42 seconds instead of 20.

I was expecting the first 4 polls go as fast as possible after each other and the SOC of course every 20 seconds.

My changes do exactly that but leaving the standard behaviour as is.

If active `PollerSend()` will be called immediately after the last incoming Poller reply was processed and no more data for this reply is expected.

If someone wants to active this in the vehicle-implementation just call `PollerSetState(state, false);`. So this behaviour can be conveniently changed with the Poller state.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -1023,6 +1023,8 @@ OvmsVehicle::OvmsVehicle()
   m_brakelight_basepwr = 0;
   m_brakelight_ignftbrk = false;
 
+  m_only_one_poll_per_second = true;
+
   m_rxqueue = xQueueCreate(CONFIG_OVMS_VEHICLE_CAN_RX_QUEUE_SIZE,sizeof(CAN_frame_t));
   xTaskCreatePinnedToCore(OvmsVehicleRxTask, "OVMS Vehicle",
     CONFIG_OVMS_VEHICLE_RXTASK_STACK, (void*)this, 10, &m_rxtask, CORE(1));
@@ -2028,6 +2030,12 @@ void OvmsVehicle::PollSetState(uint8_t state)
     }
   }
 
+void OvmsVehicle::PollSetState(uint8_t state, bool only_one_poll_per_second)
+  {
+    m_only_one_poll_per_second = only_one_poll_per_second;
+    PollSetState(state);
+  }
+
 void OvmsVehicle::PollerSend()
   {
   OvmsMutexLock lock(&m_poll_mutex);
@@ -2136,7 +2144,7 @@ void OvmsVehicle::PollerReceive(CAN_frame_t* frame)
           (frame->data.u8[2] == m_poll_pid))
         {
         m_poll_ml_frame = 0;
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[3], 5, 0);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[3], 5, 0);
         return;
         }
       break;
@@ -2183,7 +2191,7 @@ void OvmsVehicle::PollerReceive(CAN_frame_t* frame)
         m_poll_ml_frame = 0;
 
         // ESP_LOGI(TAG, "Poll ML first frame (frame=%d, remain=%d)",m_poll_ml_frame,m_poll_ml_remain);
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, m_poll_ml_remain);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, m_poll_ml_remain);
         return;
         }
       else if (((frame->data.u8[0]>>4)==0x2)&&(m_poll_ml_remain>0))
@@ -2204,7 +2212,7 @@ void OvmsVehicle::PollerReceive(CAN_frame_t* frame)
           }
         m_poll_ml_frame++;
         // ESP_LOGI(TAG, "Poll ML subsequent frame (frame=%d, remain=%d)",m_poll_ml_frame,m_poll_ml_remain);
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[1], len, m_poll_ml_remain);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[1], len, m_poll_ml_remain);
         return;
         }
       break;
@@ -2249,7 +2257,7 @@ void OvmsVehicle::PollerReceive(CAN_frame_t* frame)
         m_poll_ml_frame = 0;
 
         //ESP_LOGD(TAG, "Poll ML first frame (frame=%d, remain=%d)",m_poll_ml_frame,m_poll_ml_remain);
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, m_poll_ml_remain);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, m_poll_ml_remain);
         return;
         }
       else if (((frame->data.u8[0]>>4)==0x2)&&(m_poll_ml_remain>0))
@@ -2270,16 +2278,22 @@ void OvmsVehicle::PollerReceive(CAN_frame_t* frame)
           }
         m_poll_ml_frame++;
         //ESP_LOGD(TAG, "Poll ML subsequent frame (frame=%d, remain=%d)",m_poll_ml_frame,m_poll_ml_remain);
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[1], len, m_poll_ml_remain);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[1], len, m_poll_ml_remain);
         return;
         }
       else if ((frame->data.u8[1] == 0x62)&&
                ((frame->data.u8[3]+(((uint16_t) frame->data.u8[2]) << 8)) == m_poll_pid))
         {
-        IncomingPollReply(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, 0);
+        IncomingPollReplyInternal(frame->origin, m_poll_type, m_poll_pid, &frame->data.u8[4], 4, 0);
         }
       break;
     }
+  }
+
+void OvmsVehicle::IncomingPollReplyInternal(canbus* bus, uint16_t type, uint16_t pid, uint8_t* data, uint8_t length, uint16_t mlremain)
+  {
+    IncomingPollReply(bus, type, pid, data, length, mlremain);
+    if (!m_only_one_poll_per_second && mlremain == 0) PollerSend();
   }
 
 /**
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.h
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.h
@@ -122,10 +122,12 @@ class OvmsVehicle : public InternalRamAllocated
     canbus* m_can4;
 
   private:
+    bool m_only_one_poll_per_second;  // Poller sends only one poll per tick/second even when more polls are active for the current tick/second
     void VehicleTicker1(std::string event, void* data);
     void VehicleConfigChanged(std::string event, void* data);
     void PollerSend();
     void PollerReceive(CAN_frame_t* frame);
+    void IncomingPollReplyInternal(canbus* bus, uint16_t type, uint16_t pid, uint8_t* data, uint8_t length, uint16_t mlremain);
 
   protected:
     virtual void IncomingFrameCan1(CAN_frame_t* p_frame);
@@ -308,6 +310,7 @@ class OvmsVehicle : public InternalRamAllocated
   protected:
     void PollSetPidList(canbus* bus, const poll_pid_t* plist);
     void PollSetState(uint8_t state);
+    void PollSetState(uint8_t state, bool only_one_poll_per_second);
 
   // BMS helpers
   protected:


dexterbg:
Soko,

thanks for your submission. Enabling more OBD-II/ISO-TP polls per second is a valuable framework extension.

Your current implementation however has three shortcomings I suggest to resolve:


1. Concurrency needs to be addressed

Your change introduces a second context for calls to `PollerSend()`: a) Ticker = Events task, b) (new) IncomingPollReply = Vehicle task.

Concurrent calls to `PollerSend()` may occur on late polls in a second or slow poll replies crossing the next ticker call.

`PollerSend()` is mutexed, but won't know / recognize a pending poll reply for data lengths <= 7, so will begin the next poll, discarding a potentially pending one just started by the other context.


2. Broken consecutive frame processing

The `m_poll_ml_remain` check in `PollerSend()` needs to be changed, or multi frame responses will get truncated.

The current solution with `m_poll_wait` was introduced in commit ace6ee7984abbd303e2f847c9316c84cad2eb16b by Dimitry. It was meant to address multi frame responses crossing the next ticker call, but also enable the poller to recover from a missed frame reply.

Dimitry's `m_poll_wait` mechanism relies on the fact `PollerSend()` is only called once per second. If it's called immediately from IncomingPollReply, I think it will cut off a large transfer after the third frame.

Btw: the poller doesn't replay polls, that's a typo and should be "replies".


3. Controllability needs to be improved

Options with your implementation are currently just "no throttling" (any number of polls per second) and "strict throttling" (one poll per second).

No throttling is normally not desirable, as CAN bus usage should generally be as cooperatively as possible, and a large polling list may infer single seconds with large numbers of polls due.

For better control, it should be possible to set a certain maximum number of sequential polls to allow. Throttling should also be controllable separately, i.e. by an API call `PollSetThrottling(max_polls_per_second)` (with default 1, and 0 = unlimited).

Changing the throttling needs to be possible any time and from another context as well, to allow the system to adapt to sudden external requirements.

Note: consecutive frame requests need to be taken into account for the throttling limit as well, but must not be throttled for a running response.


Please tell me if I need to be more specific or explain some detail.

Regards,
Michael

SokoFromNZ:
Hi,

I was already wondering about the mutex. Are there mulitple task/threads in OVMS? I thought this is not possible on simple embedded systems. But maybe its just in Arduino...

The issue with data length <= 7 is a known issue and has nothing to do with my change, correct? Or am I missing something? Just want to make sure I understand everything.

Everything else I have to look at in detail when I have more time. But answers to the stuff above will help me then for sure 👍 

thx
Soko

dexterbg:
Soko,

please take the time to understand the underlying system architecture and the basic framework concepts. The esp-idf is based on FreeRTOS with multi core extensions, as the ESP32 has two cores. You can get a list of the currently running tasks by the `module tasks` command.

There may well be an issue with `m_poll_ml_remain` <= 7 now, but it's no known issue. Known issues are those you can find in the open issues section here on github.

Thinking now about it: yes, I suppose there is actually an issue with this check, as the last pending frame of a multi frame transmission always has less than 8 bytes payload. Slipped my attention, Illustrates why we need to insist in keeping framework commits separated from vehicle commits.

@Dimitrie78 maybe we're wrong, please have a look.

Regards,
Michael

SokoFromNZ:
Ok, understood. Now that I know we are having a-synchronicity here I know how to think. Doing async development since years, although in C# where is it a little bit easier ;)

So the concept of the `m_poll_ml_remain` and `m_poll_wait` is to make sure no other poll is transmitted until all parts of the data from the current poll are received, correct?
Is there any handling of the case when the last (or parts in between) get lost or corrupted via CAN transmission? Is there a timeout?

I'm just thinking out loud here. When I'm on here I'm as well try to improve this as well.

dexterbg:
> So the concept of the m_poll_ml_remain and m_poll_wait is to make sure no other poll is transmitted until all parts of the data from the current poll are received, correct?

No. It's meant to delay the next poll by one additional second in that case.

> Is there any handling of the case when the last (or parts in between) get lost or corrupted via CAN transmission?

No.

> Is there a timeout?

Yes, one second by the ticker call.

> A new Timer is created which just calls the Poller with the set Frequency

That's also an option, more complex to implement, but also more cooperative to the bus as well. If going that way, you might also think about reinterpreting `poll_pid_t.polltime[]` as poll intervals rather than seconds.

Regards,
Michael

dexterbg:
On a second thought: a configurable poll frequency is more an additional feature. It won't solve the requirement of getting some poll responses as fast and close in time as possible.

For example, you might need to poll multiple PIDs to get an accurate overall continuously varying state.

Dimitrie78:
My problem was, when i query for battery cell voltage, the response take longer than one second. 
Example:
with a flow control of 25 ms and > 50 response messages it will take > 1250ms.

That was the reason why i add the waiting for one more second to komplete the responce

dexterbg:
@Dimitrie78 Thanks for joining. The question is, why do you check for a remaining data length of more than 7 bytes (= more than one frame)? What if just the last frame is still missing?

Dimitrie78:
Ohh, that's a good point. I didn't think of that at all. since it works well so far.

But we can change it, to wait for them.

SokoFromNZ:
OK guys.
Thanks heaps for your input. 
I think I will go ahead and try another implementation of this feature which uses the vehicleticker (so no separate timer) and does all the polls queued for this second as fast as possible (with the upper throttle limit).
Also trying to correct the 7 bytes remaining limit/wait.
It wont be perfect at the first try but with your input I will get to a worthy solution for the master. 
I will let you know as soon as I have a new commit done to this PR.

dexterbg:
@Dimitrie78 There's yet another issue with your change I didn't see before: you also didn't take concurrency into account. Your code introduces a race condition between `PollerSend()` and `PollerReceive()`, as both now read and write `m_poll_ml_remain`.

If both are called concurrently and pass the condition `m_poll_ml_remain > 7` at the same time, `PollerSend()` can set `m_poll_ml_remain` to 0 just prior to `PollerReceive()` decrementing it by 7, leading to an unsigned wraparound in `m_poll_ml_remain`.

I haven't checked if that leads to actual issues with any vehicle poll reply handler, but it at least can lead to the handler and the poller not recognizing the poll response is complete, so the poller would continue requesting further response frames. A poll reply handler will see a very large number (near 64K, uint16) of remaining response bytes, which may lead to memory issues. That's the kind of bug that occurs very rarely and can then lead to mysterious effects like the mystery WDT crashes I've been investigating lately…

I think we would need to add the mutex lock to `PollerReceive()` as well to fix this race condition. But as the check seems to be wrong anyway, maybe you can find a better overall solution?

Regards,
Michael

Dimitrie78:
Hi Michael,

i'll try to find a better solution.
btw. my module runs since 16 days without any crash

Regards,
Thomas

SokoFromNZ:
So, do you guy take care if this or should I during my changes. We would kinda work on the same thing here which is not neccessary imnho

Dimitrie78:
Hi Michael, 

i think i found why i set `m_poll_ml_remain` to 0 in `PollerSend()`
the Problem is in `PollerReceive()` 
```
case VEHICLE_POLL_TYPE_OBDIIEXTENDED:
......
// prepare frame processing, first frame contains first 4 bytes:
m_poll_ml_remain = (((uint16_t)(frame->data.u8[0]&0x0f))<<8) + frame->data.u8[1] - 3;
m_poll_ml_offset = 3;
```

it should be: 
```
m_poll_ml_remain = (((uint16_t)(frame->data.u8[0]&0x0f))<<8) + frame->data.u8[1] - 2 - 4; // six data bytes already read (+ two type and length)
m_poll_ml_offset = 4;
```

SokoFromNZ:
Hhmmm... I think this is why I have to skip the first byte on my polls which are longer than one msg. I was already wondering about that

SokoFromNZ:
About the race condition:
Would this work on the OVMS?
https://en.cppreference.com/w/cpp/atomic

SokoFromNZ:
@Dimitrie78 : I've added your PollerFix changes into here

SokoFromNZ:
Hi guys,

Have a look at the current state of the source. I think I've got everything.

Soko

Dimitrie78:
Hi Soko, 
your merge is not correct...

SokoFromNZ:
Hhmmm... I cant find a difference. Can you point it out for me pls. thx.

Dimitrie78:
i did my changes in `case VEHICLE_POLL_TYPE_OBDIIEXTENDED:`

image

you in `case VEHICLE_POLL_TYPE_OBDIIVEHICLE:
    case VEHICLE_POLL_TYPE_OBDIIGROUP:
    case VEHICLE_POLL_TYPE_OBDII_1A:`
image

SokoFromNZ:
Thanks Thomas,
First I couldn't understand how this mistake happened as I took a careful look at your changes. But I learned something new now: There are the blue arrows on the left which hide source lines. This is how I got mixed up with the case of the switch.
Should be all correct now. But please have another look to make sure.
Soko

Dimitrie78:
Looks correct now

dexterbg:
Soko,

you're heading in the right direction :)

Some new comments from auditing your new version:

1. A std::atomic does not solve the concurrency issue and introduces unnecessary overhead here. A std::atomic can only protect defined atomic operations (like +=, -=, >>= etc.), it cannot replace a mutex for a critical section. As we already have a mutex for the poller, the standard solution is simply applying that to the receive context as well, as suggested before.

Using the mutex will also protect the newly introduced concurrent accesses to the new `m_poll_cnt_this_ticker` and `m_poll_max_per_ticker` variables from the sender, receiver and control contexts.

Hint: you need to change the mutex into a recursive one, as `PollerSend()` can now be called from within `PollerReceive()`. That's simply exchanging the types for the recursive variants, see `ovms_mutex.h`.


2. I don't see how the current check for `m_poll_ml_remain > 0` in `PollerSend()` would protect a missing single frame response. Single frame response polls don't have an `m_poll_ml_remain` value, but can also occur on the edge of the next ticker.


3. Correct me if I'm wrong: your change to the `VEHICLE_POLL_TYPE_OBDIIEXTENDED` (type 0x22) receiver in regards to the remaining data length & offset doesn't seem to be correct as well.

You seem to be right: the `IncomingPollReply()` call arguments look wrong. Contrary to the comments (probably copy & paste…), the first frame of a 0x22 type message contains three data bytes, as the PID is 16 bit. The 8 bit PID multiple frame response handling excludes the PID from the payload, so we should do the same for 16 bit PID responses. Beginning the payload in the middle of the PID doesn't seem to make any sense at all.

But wouldn't that mean we should subtract 5 from the ISO TP length (2 header + 3 data bytes), keep the offset at 3 (3 bytes into the data), and then pass `&frame->data.u8[5], 3` to `IncomingPollReply()`?

Type 0x22 is called `ReadDataByIdentifier` by UDS / ISO 14229 and is specified to include the 16 bit PID as the first two bytes in the response payload. Take your discussed Ah / kWh message as an example: the first reply frame carries a length indicator of 19 bytes. That includes the two PID bytes (as ISO TP doesn't care about UDS), so the actual payload is 17 bytes beginning with the three remaining bytes in frame 1 + 2 x 7 bytes in the remaining two frames.

So after the first frame has been received, `m_poll_ml_offset` should be 3 (3 bytes read) and `m_poll_ml_remain` should be 14.


4. This handling of the payload on type 0x22 responses has been in the code since it was added by @goev 17 months ago (https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/commit/10d334c9f61af00a8aa2bc9bc45d331b123e93dc).

While the current handling may be wrong, it apparently has been working for all `VEHICLE_POLL_TYPE_OBDIIEXTENDED` users up to now. So if you change the handling now, you absolutely need to check the impact on all existing usages, and need to correct all incoming handlers that rely on the current behaviour.

You should try to get the other type 0x22 users (vehicle developers) on board for this.

@goev Geir, this is your code, and I think you're also the primary type 0x22 user. Can you please have a look here?


Regards,
Michael

SokoFromNZ:
Hi Michael,

**Point 1** 
Is correct now (hopefully ;) )

**Point 2**
Understood (but it didn't do that before either *gg*). Frankly I thought about that... But if I introduce a wait and the reply gets lost the poller would wait forever. So the the timeout functionality of the ticker would be lost.
So one option would be to save the timestamp of the last Tx and skip the ticker process of `PollerSend()` when the timestamp is younger than i.e. 500ms. Easiest would be a method like Arduinos `millis()` method but I couldn't find any. Can you point me to it if this is an option for you as well?

**Point 3 & 4** 
They kinda go together (if it's not too late for me to understand it). I just merged the changes of @Dimitrie78 here which seems to be not correct if I understand you correctly.
I also get the issue when changing this behavior for the implementations of `IncomingPollReply(..)` in all vehicles as the data shifts. But I'd like to have a more detailed look here tomorrow.

Soko

SokoFromNZ:
@dexterbg Please hold off with your review...

SokoFromNZ:
@dexterbg ready when you are :shipit:

drc38:
All,

Given we are adjusting the Poll logic in the main vehicle framework, appreciate if adding the following function can be considered. This is needed on the Leaf to switch can buses for the various polls to be successful depending on the awake state of each bus. It may be useful for other vehicles that are reliant on polling. If appropriate to include I can submit a separate PR for it.

`
// Use to switch can bus to poll without resetting poll ticker
void OvmsVehicleNissanLeaf::PollSetBus(canbus* bus)
  {
  OvmsRecMutexLock lock(&m_poll_mutex);
  m_poll_bus = bus;
  }
```

drc38:
Thanks Soko 👍

SokoFromNZ:
@drc38 I've added your method here to keep you from the overhead of an own PR but I've called it `PollerSetBusDirty` as it does not cleanly reset all the counters (as requested). Otherwise you can just use `PollSetPidList`.
If @dexterbg doesn't like the `PollerSetBusDirty` method as is (or here in this PR) I will remove it again and you should do a separate PR for it.

@dexterbg I think I've found the error you saw as well and corrected it. In addition I've made sure `m_poll_ml_remain` is set in all paths in `PollerRecieve`. I thought of using just the parameter `mlremain` in `IncomingPollReplyInternal` instead of `m_poll_ml_remain` but I think the intentions are clearer when using the instance variable instead of the parameter.

dexterbg:
Sorry guys, I'm not at all fine with adding the bus changing API call. Neither to this pull request nor at all in this form.

This pull request has become complex enough on it's initial purpose, with detecting additional bugs along the way. Generally spoken, I prefer having separate pull requests for separate features. Just like you would do with your commit history to keep things clear & reversible. If additions build a block or belong to each other, that may be OK, but even then you normally can split stuff into steps.

So, @SokoFromNZ, please revert the addition of `PollerSetBusDirty()`.

@drc38, please explain your idea and reasons for this totally dangerous method, either on the developer list, or in a separate pull request. I've had a quick look at your code and do not agree this method is necessary or even nice to have. Polling lists shall be kept separate for different buses, if you need to switch the bus, you shall switch the list as well. But explain your reasoning, maybe I misunderstood your requirements.

As a general note: if we add a method like this, it must at least include a very clear description about the purpose and dangers, and it must take measures to assert it's called in the right way and/or cannot corrupt a running poll.

Regards,
Michael

PS: @SokoFromNZ I'll continue checking your other changes.

dexterbg:
Soko,

sorry, have been interrupted.

Looks good now. Only minor code style & optimization issues remaining:

- The recursion in `PollerSend()` can be avoided quite simply by an outer loop. Although it can be seen as a tail recursion, I doubt the compiler can see that as well, as the compiler doesn't know the mutex lock can be skipped.
- It's a good idea to do a full polling list cycle on a ticker run, but I think you can implement that even better by remembering the initial cursor and looping once around until reaching that cursor again. That way you avoid checking the whole list twice if the ticker run started at the second entry.
- The call to `PollerSend()` from `IncomingPollReplyInternal()` can be skipped as well if `m_poll_wait` is set.
- The comment on line 2311 begs to be replaced by a comment telling you actually knew what you were doing… ;-) (and yes, it does, as the `PollerReceive()` lock object won't get deleted until leaving it's creating context)
- Line 2315 has an additional pair of parentheses…
- …which can be reused to make line 2061 more readable and less prone to future compiler warnings ;-)

That should give us a mergable state.

Poll timeout detection can be changed to be time stamp based later on, avoiding the two second timeout for polls on an inactive bus. `esp_log_timestamp()` and `esp_timer_get_time()` are simple options for this.

And yes, we should address the wrong type 0x22 receive handling separately. I'll start a thread on the mailing list after the merge.

Regards,
Michael

PR #406, opened 2020-08-13
VW e-Up source code split for T26A, OBD

SokoFromNZ:
So the development of the T26A and the OBD implementation of the VW e-Up can continue without disrupting each other I have split the source code for the eUp.

In addition the vehicle code for the eUp was renamed. This is taken care of in `OvmsConfig.upgrade()`.

Despite git showing a lot of deletes all I've done is renaming the original three `vehicle_vweup*` files to `t26_epup*`. It recognized the rename for the `..._web.cpp` but not for the other ones (don't now why).

As the OBD implementation is not complete yet the whole OBD vehicle type is deactivated and therefore not active for the user.


SokoFromNZ:
I see the point in not having unfinished code in the master so I removed my OBD implementation from this pull request.
Whats left is just the same source as before but with preparation for the OBD part.

dexterbg:
Soko,

two issues here:

1. as Chris vetoed against this PR and the PR changes files he is working on, I've asked him to check and approve this.

2. Please rename the vehicle classes to use the `OvmsVehicle` prefix, i.e. `OvmsVehicleVWeUpT26`, not just `VWeUpT26`. All subclasses of `OvmsVehicle` shall keep that prefix, so greps for vehicle methods that may need to be changed are simple & efficient.

Regards,
Michael

devmarxx:
I believe it is a to early to merge this code to the master. At the T26A part we are close to a working climate control. That feature would be a real benefit for the master so I would would to wait till that is done.

On the OBD part we have no real functionality at the moment and the code is not documented by now. 

If we would have real benefit from the OBD part I would agree to a commit, even without working climate control on the T26A part. But for now it seems to early to merge that code. 

There is no actual use for the users.

Regards,
Chris

dexterbg:
A suggestion: as you no longer seem to share a common base class, you maybe should also stop sharing a component directory.

As written before, these can be viewed as two separate vehicle adaptions, so you can also introduce two separate vehicle components.

That way you won't have disputes about which parts to add to the master.

Regards,
Michael

SokoFromNZ:
ad 1.
As I pointed out to Chris a couple of days ago and via the mailing list as well this is not about benefiting the user, its about being able to continue developing with a clear break between OBD and T26A and on separate forks as Chris' pace and style is different than mine. In addition Chris' fork is easily updateable/mergeable (if this are English words ;) ) after this PR is merged. Ohh, and there is no OBD part here in this PR. So I don't know why Chris is keeping talking about it...
So maybe its better to completely cut away from Chris' T26A and do a completely separate vehicle. So we don't even have to argue anymore

ad 2.
Ahh Ok. Was just writing that I did as everybody else. But you don't talk about the file names, you mean the class names. OK, understood.

SokoFromNZ:
OK, so I close this PR and do a completely own vehicle for my OBD implementation.

PR #411, opened 2020-08-22
Obd2 ext new handling

SokoFromNZ:
Hi @Dimitrie78 and @dexterbg 

Here is the change of the `case VEHICLE_POLL_TYPE_OBDIIEXTENDED` in `PollerReceive()` we talked about.

Like this it works for the VW e-Up.

In addition I changed the send interval according to the msgs VCDS is sending. In addition I want the full data asap...

Let me know your thoughts

Soko

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ dkms.conf
 # Sphinx builds
 docs/build/
 .idea/
+
+#configs for Visual Studio Code
+.vscode/
--- a/docs/source/userguide/components.rst
+++ b/docs/source/userguide/components.rst
@@ -20,6 +20,16 @@ At one end of the module is the main DB9 connector you will use to connect to th
 
 At the other end of the module is the DA26 expansion connector, the USB diagnostic connector, and a Micro SD card slot.
 
+|clearfix|
+
+.. warning::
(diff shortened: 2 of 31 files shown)


dexterbg:
Soko, please wait with further work on this topic. I'm thinking about a rewrite of the poll receiver, as it has some long known shortcomings along with the bug we detected. 

Regards,
Michael

SokoFromNZ:
Okeydokey. Let me know here if you have a rewrite...

drc38:
Is there any reason to set the interval for extra frames at 1ms? `txframe.data.u8[2] = 0x01; // with 1 ms send interval (as fast as possible)` I note the Leaf is ok with it being set to 0. Would it make sense to be able to alter this value by vehicle type?

dexterbg:
@drc38 I've already taken this into account in my yet to publish rework.

dexterbg:
Rewrite of the poller receiver is now pushed in branch `poller-rework`:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/commit/0b306ad4f5e7a1eb63a471a11ec7f1a6d69d8647

The rewrite obsoletes all or most of this PR, so I'll close this now. If you want to add to the poller rework, please create a PR for that branch.

PR #413, opened 2020-08-26
NL: add obdrequest command by canbus and improve polling

drc38:
Hi Michael,
Appreciate if you can review the changes I have made to the `ObdRequest` function to handle requests to either can bus, and that I'm using the semaphore and mutexes appropriately.

Also the shell command call to `ObdRequest` is working well but I cannot understand why the same function is not working from `Ticker60` (no TX/RX frames appear on the can bus before the timeout is logged), if you can let me know where I've gone wrong that would be great ;)


dexterbg:
Derek, please recreate this PR against the `poller-rework` branch, so effective changes can be seen clearly.

PR #414, opened 2020-08-28
Vehicle poller updates to allow changes in bus polled

drc38:
Hi Michael,
Appreciate if you can review the changes I have made to the `ObdRequest` function to handle requests to either can bus, and that I'm using the semaphore and mutexes appropriately. This approach removes the need to use the vehicle mutex.

Also the shell command call to `ObdRequest` is working well but I cannot understand why the same function is not working from `PollOnceRequests` called by `Ticker60` (no TX/RX frames appear on the can bus before the timeout is logged), if you can let me know where I've gone wrong that would be great ;) I assume it is some issue with my mutex or semaphore use?
Thanks

### DIFF
--- a/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
+++ b/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
@@ -1224,6 +1224,7 @@ void esp32wifi::EventWifiScanDone(std::string event, void* data)
         {
         SetSTAWifiIP();
         }
+      else StartDhcpClient();
       }
     }
   if (list)
--- a/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
+++ b/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
@@ -1224,7 +1224,6 @@ void esp32wifi::EventWifiScanDone(std::string event, void* data)
         {
         SetSTAWifiIP();
         }
-      else StartDhcpClient();
       }
     }
   if (list)
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -2044,13 +2044,13 @@ void OvmsVehicle::PollSetState(uint8_t state)
 /**
  * PollSetResponseSeparationTime: configure ISO TP multi frame response timing
  *  See: https://en.wikipedia.org/wiki/ISO_15765-2
- *
+ *  
  *  @param septime
  *    Separation Time (ST), the minimum delay time between frames. Default: 25 milliseconds
  *    ST values up to 127 (0x7F) specify the minimum number of milliseconds to delay between frames,
  *    while values in the range 241 (0xF1) to 249 (0xF9) specify delays increasing from
  *    100 to 900 microseconds.
- *
+ *  
  *  The configuration is kept unchanged over calls to PollSetPidList() or PollSetState().
  */
 void OvmsVehicle::PollSetResponseSeparationTime(uint8_t septime)
@@ -2104,9 +2104,6 @@ void OvmsVehicle::PollerSend(bool fromTicker)
         m_poll_moduleid_high = 0x7ef;
         }
 
-      ESP_LOGD(TAG, "PollerSend(%d): send [type=%02X, pid=%X], expecting %03x/%03x-%03x",
-               fromTicker, m_poll_type, m_poll_pid, m_poll_moduleid_sent, m_poll_moduleid_low, m_poll_moduleid_high);
-
       switch (m_poll_plcur->pollbus)
         {
         case 1:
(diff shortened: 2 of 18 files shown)


dexterbg:
Derek, there are Wifi changes in this PR now, maybe you meant to add them to PR #416 instead?

dexterbg:
Basically looks ok to me from a quick scan.

Just one thing: you should try to avoid doing the synchronous ObdRequest() calls from the events task context (ticker.60), as with a timeout of up to 2 seconds per call, you can delay the whole event processing by up to 6 seconds now.

I would rather create a second poll list for bus 2 containing the three polls, then based on timing or poll responses coming in switch to that list and back to the main list when the three polls are done. That way you don't have to deal with synchronous requests. But that's just a thought, I haven't fully checked your logic.

As an alternative: polling over multiple buses using the same polling list could be added as feature to the framework. That would be relatively simple, every poll entry could add an optional explicit bus tag, with `PollSetPidList()` setting the default bus. That would avoid the dangerous bus switching during a poller run from outside the poller, but still provide what you need, wouldn't it?

drc38:
> Derek, there are Wifi changes in this PR now, maybe you meant to add them to PR #416 instead?

sorry, was wanting to have both in my local build. Forgot merging after the initial pull request would add it here. Will revert.

drc38:
I will give changing the framework a go, with an optional extra field in `poll_pid_t` that is used to set an alternate bus to poll on (if specified). That would be the cleanest approach I think. Although some polls, like vin, are really only needed once it is not going to cause issues using an infrequent poll timer.

drc38:
Appreciate if you can take another look, thanks :D

dexterbg:
Derek,

please keep framework PRs separate from vehicle PRs. Also, please do not remove trailing whitespace from other developer's comments, they have been placed deliberately there (maybe your editor or git config did that).

Regarding your change to the poller: that's basically what I meant, but not implemented in the right way:

- Your current design changes the poller bus for all subsequent polls. You may have meant that for a polling list that will always be processed in a strict sequential order, but that is not the case here, polls will only be sent if they match the current poll tick.
- Even if you would change this to be applied to all list entries, that mode of operation introduces side effects (polls affecting other polls), which the list design does not do now. This requires the developers to always remember this and take this into account when rearranging the poll entries or adding new ones. As sending CAN frames to buses they are not meant for can have all kinds of bad effects, we should try to keep the polling structure simple to avoid this kind of mistake.

My solution would be to restrict the bus control strictly to the single poll entry. If an entry has no specific bus (pollbus=0), it shall always fall back to the default bus set by `PollSetPidList()`. That simply requires adding a member like `m_poll_bus_default` to store the default bus.

Style notes:

- Your code change in the sender now mixes the poll state preparation with the frame preparation. That's a bad style pattern which leads to poorly understandable code and mistakes from broken control/data flow. You should first determine the poll state (set `m_poll_bus`), then use that state to fill the frame.
- Regarding compiler warnings: you generally should not ignore them and not view them as optional. They include valuable information you should use to create better code. If you introduce a code change leading to warnings, you should rather correct the code than define the warning as ignorable. Updating the poller definitions is a simple option in this case.

Regards,
Michael

dexterbg:
PS: just saw you also forgot to support bus 4 (the optional SW CAN addon)

drc38:
Hi Michael,
There was an option, which is set by default on Atom, to remove whitespace and have now turned it off. Thanks for pointing it out.

I reverted the NL changes so that the vehicle framework side can be worked on / resolved first. 

Thanks for the code style notes, I see what you mean on item 1. The updated logic is much clearer. Let me know if I haven't covered all your points on implementation approach.

On item 2, I had not intended to imply that the vehicle module poll lists should not be updated, just that because the `pollbus` is initialised to `0` when not explicitly set, and `0` uses the bus set by `PollSetPidList` it will not break any existing lists defined.
Cheers
Derek

dexterbg:
Derek,

nearly perfect. Please check the diff for remaining whitespace changes to revert: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/414/files

Please also move the bus switch statement above the debug log output and add the bus number to the debug log message, e.g. `…send [bus=%d, type=%02X, pid=%X]…`. That will make reading the log on multi bus polls much easier.

Regards,
Michael

dexterbg:
PS: and please rebase your PR onto the latest `poller-rework` state to resolve the merge conflicts.

drc38:
Should be close now ;)

dexterbg:
OK. Please add the updates to the `poll_pid_t` usages to this PR, so we've got a complete package.

drc38:
In the `changes.txt` file or somewhere else?

dexterbg:
No, I meant the changes to the vehicle modules to avoid the compiler warnings.

drc38:
In order to keep the PR for the vehicle changes separate, how do I do that with this PR not merged into `poller-rework`? As the vehicle `poll_pid_t` changes rely on this PR to compile. Locally I have forked a new branch based on `poller-rework` and then merged the changes above from  `Leaf-updates`, vehicle code is updated but if I try and create PR it will bring through both rather than only the vehicle changes as a separate PR.

dexterbg:
You normally do that by separating the changes in the Leaf module, i.e. only add the `poll_pid_t` changes accompanying the framework change to the commit. In case you didn't know: git allows to add selected lines to commits, not just complete files. In `git gui` you can do that easily from the diff view's context menu, it provides both code context and selection based add operations.

If your changes to the Leaf module cannot be separated easily by their code lines, you can also leave them out here and move them to your Leaf PR, assuming that will follow shortly.

drc38:
Thanks, I wasn't aware I could select elements of a file. Got it the wrong way round first time though...hence revert.

PR #421, opened 2020-09-01
VW e-Up via OBD

SokoFromNZ:
Hi guys,

Here is my stable and tested implementation of the VW e-up via the standard OBD-II port.

For feature details see index.rst file.

Please let me know your thoughts.

Soko


dexterbg:
Soko, excellent work.

Things to change:

- Custom metrics, config params and commands name space is defined to use a unique common prefix/namespace `x<vehicleabbreviation>`. For the UpMiiGo, you can e.g. use something like `xupobd` or `xuo` (Up OBD) to keep this short. See other vehicles for examples and used prefixes.

- Do not introduce custom metrics for data already covered by the framework. This applies at least to your energy counters, see standard metrics `v.b.coulomb.…` and `v.b.energy.…`.

- Regarding real custom metrics names, if addressing a technical metrics area already defined in standard metrics, stick to the framework naming for your custom metrics as well. I.e. use `b` instead of `batmgmt`, and `c` instead of `chrgr`. That is to allow users to query all related metrics by a simple `.b.` or `.c.` selector. This applies to all name components, i.e. don't use `chrgd` if the framework calls it `recd`.

- The log tag to use for a vehicle shall be `v-<vehicleabbreviation>`. You may use a longer abbreviation here, for example `v-vweup-obd`. You should try to avoid introducing separate log tags for your functional modules. If you need to, keep the base prefix and add a suffix. Tags without a vehicle prefix are reserved for framework components.

- Exclude the `vehicle.cpp` whitespace change.

- Include an addition to the `changes.txt` file reflecting the vehicle addition (add latest changes at the head).

Regards,
Michael

SokoFromNZ:
Thanks Michael, I'll try ;)

I did:

- vehicle.cpp
- changes.txt
- log tag
- renamed custom metrics

What I didn't do was getting rid of any custom metrics for standard metrics. Reading your post I thought I overlooked some standard metrics... Because I looked at all standard metrics before creating customs ones of course!

Lets address them seperately:

- xuo.b.energy.used: The standard metrics for `coulomb` says [Ah], mine is [kWh]. There is an `...energy_used` but it says "... on trip". Mine is total and not for the trip.
- xuo.b.energy.charged: Same as above. About the naming: The standard naming `...recd` stands for recovered. My counter includes charging as well, not just recovered. Thats why I called it `...charged` (given that recovering is a kind of charging ;) )
- For all others I haven't found anything close in the standard metric.

So from my point of view everything should be OK for merge with the last commits. Or the descriptions of the standard metrics are wrong :D ;)

Soko

dexterbg:
Soko,

looks good. If your energy counters are total (i.e. battery life time), that's not yet covered by a standard metric, so a custom metric is OK. Proposing standard metrics for this would also be an option though.

Still your metric naming should be "recd" if your counter includes recuperation and charging. The term "recd" ("recovered" / "received") was specifically chosen to cover both types of energy gaining from the battery perspective, and users should not be confused by adding other terms for the same meaning. If your counter would only include charging, you could name it "charged".

Another thing you need to change (didn't see that before, sorry) is the poll definition list: the current master includes the multi bus polling extension by Derek. We will get compiler warnings now when building with the current version, as the bus number is missing from your entries.

Both isn't critical though, so I'll merge this now to get this into the next edge release.

Remember, if you make a PR for a standard metrics additions, keep that separate from the vehicle changes.

Regards,
Michael

PR #424, opened 2020-09-04
New standard metrics for total (life time) energy and coulomb

SokoFromNZ:
Hi Michael,

As discussed in the VWUP.OBD PR. Once this is merged I will adjust the VWUP.OBD implementation accordingly and also correct the poller-array.

Thanks
Soko


dexterbg:
Soko,

basically OK, also to request persistence for the energy counters, but you need to raise the persistence slot count as well:

```
OVMS# metrics persist 
version 1, serial 2, 420 bytes, and 20 of 20 slots used
```

I think to get some headroom we should raise the slot count to 25 or 30.

You also didn't take into account a persistent metric may currently have max 16 chars in the name. There is no good option to find shorter names without breaking the scheme, so I think we need to raise the name space to 24 bytes.

That should still fit into RTC RAM. We need to change the storage structure long term to a packed one.

Regards,
Michael

SokoFromNZ:
Hi,

Can it be as easy as my last commit? I looked but I couldn't find any other lines that needed change.

Soko

dexterbg:
Plus…

```
#define PERSISTENT_VERSION 1            /* increment when struct is changed */
```

Comment says all. Not technically necessary, as the size is also checked, but as it's meant to be incremented, we should do so.

SokoFromNZ:
Yeah, I've checked the size calculation and found its done automatically. Thanks for the hint: Incremented to 2.

dexterbg:
Another thing I missed pointing out: you forgot to update the documentation (metrics.rst). But that can be added later on.

PR #431, opened 2020-09-08
Add MG ZS EV support (fix #252)

### LINKED ISSUE #252: MG eZS support?
I'm keen to help with adding support for the MG eZS. Is anyone else interested? Has anyone got access or an order placed for an MG eZS? Getting access to the car for testing will be the main issue. I may be able to help with this. Please contact me if interested, financial support may be available to help with development.

It looks like a rather interesting EV, Saic Chinese built with a decent battery size and very competitive price.

There is a lot of excitement regarding the launch of this EV in the UK, pre-orders have sold out. In Europe (unlike in Asia) the MG eZS is not due to come with mobile app support, there is a lot of people keen to add third party OVMS app support.

https://mg.co.uk/mg-zs-electric/

chrisstaite:
This is an initial request which is not fully tested yet and has debugging code to allow use on CAN2 which causes the alarm to be activated and therefore should be removed before merge.

Am performing the request now to get style and general comments to the code while it iterates now that most of the basic functionality is in place.

Notably, all of the functionality is performed through polling.  As with other vehicles, we are requiring canwrite to be set.  However, since the module is pointless without this, I was wondering whether to remove this requirement.  What are your thoughts?


dexterbg:
Chris,

very nice & well written code, and closely following the general framework style, requirements and recommendations.

A minor note: we normally try to reflect the manufacturer plus the vehicle model in the car type, i.e. `NL` = Nissan Leaf, `MI` = Mitsubishi i-MiEV etc.. The ZS is probably (hopefully) just the first EV by MG, so `MGZS` or `MZ` could be a better choice (with `xmz` as the extension namespace). But I'll leave that up to you, `MG` & `xmg` is fine if you'd like to stick with it.

Regarding the CAN write permission, it's OK to not respect/support that config (the OBD vehicle as well doesn't), or interpret this as a permission to request actual register writes if you add such in the future.

I would have thought…
```
    // Trim the first character from the VIN as it's seemingly junk
```
…this was due to the bug in the framework's former 0x22 type OBD polling, but as that's only happening on the VIN, maybe that's a checksum or some VIN extension.

Regarding your tester presence: are you sure that's a type 0x02 (freeze frame) request for PID 0x3e? Freeze frame PID 0x3e is normally assigned to a standard OBD2 PID (Catalyst Temperature: Bank 1, Sensor 2), and "tester present" is normally a type (!) 0x3e request with an 8 bit PID (see my latest extension of the poller service type support). Maybe you got the service type mixed up with the request length?

Regards,
Michael

PS: Looking forward to your OK to merge this.

dexterbg:
Re tester presence: I just saw you actually "misused" the poll type:
```
        { .u8 = { VEHICLE_POLL_TYPE_OBDIIFREEZE, testerPresentPid, 0, 0, 0, 0, 0, 0 } }
```
The first byte in the frame is the ISO TP control byte, that's simply the length on a single frame request, the poll type is in the second byte. So I recommend changing that to…
```
        { .u8 = { (ISOTP_FT_SINGLE<<4) + 2, VEHICLE_POLL_TYPE_TESTERPRESENT, 0, 0, 0, 0, 0, 0 } }
```
…to make things clear. ISOTP_FT_SINGLE = 0, it's just to document the protocol.

Regards,
Michael

chrisstaite:
We believe that the MG5 will use very similar software to the MG ZS, hence why it was decided to name the module in the way it is.

The first value in the VIN is 0x01 on my vehicle, maybe it's a checksum?

Thanks for explaining the tester present PID to me, I am pretty new to OBD, so I was looking for results rather than correctness.

dexterbg:
`MG` is reasonable then. 0x01 could also be a VIN scheme version number.

chrisstaite:
Fairly happy with this now.  Still doesn't support getting the SoC when the car is woken by the charger, but it's a good start and it seems alarm safe now.

dexterbg:
OK, I'll take that as a "GO" for the merge :-)

PR #436, opened 2020-09-13
NL: add custom charger power and efficiency metrics

drc38:


dexterbg:
Derek & @SokoFromNZ,

you both have support for a charger efficiency now, so I've added this to the standard metrics. I've also checked and added the power & efficiency metrics for the inverter, so we have the full chain.

Please change your code to use the new standard metrics.

Regards,
Michael

drc38:
@dexterbg,

Thanks for adding these to the standard metrics,  is it worth adding "v.b.bms.balancing" or similar now that I've added to Leaf as well?

Cheers Derek

dexterbg:
I'm still thinking about that, because I also have the balancing status and time totals in the Twizy. But we wouldn't be able to use a bitset, as that would need a fixed size. At least not without changing the whole metrics callback implementation. I haven't decided on this yet.

Regards,
Michael

SokoFromNZ:
Heya,

Ok, but advice me which of my values I should put where:

I have an AC power, DC power and charger/inverter? efficiency.

You have Charger power+eff and inverter charger+eff.

Thanks

Soko

dexterbg:
Soko,

the inverter (german: Wechselrichter) is the motor driver, translating DC from the battery into AC for the motor and vice versa.

If I got you right, your "AC power" is the charger input power. Your "DC power" would be the battery power (negative there because input for the battery).

"AC power" may be a bit misleading if you take DC fast charging into account?

SokoFromNZ:
@dexterbg 

Ok, understood.

So...
my efficiency = charger-eff
and inverter is nothing for me ;)
so far so good.

But the ECU (https://www.goingelectric.de/wiki/VW-e-up-OBD2-SGC6) even calls it AC voltage/current. So I'm not sure what happens with DC fast charging here, but my values are AC.

Should I put my AC/input power to the standard metric or the DC/output value? I guess input, correct?
So I leave DC/output in my custom metrics.

Soko

dexterbg:
Why not set your DC/output as the battery power level as suggested?

SokoFromNZ:
You mean "v.b.power"?

dexterbg:
OK, I see you already set the battery power from voltage x current. If that is available during charging and equals the charger output power, that's fine. If not, you might consider using both sources depending on the vehicle state.

SokoFromNZ:
No, it's not equal as far as I remember. The battery power is less as we still have some power loss from the charger output to the battery input.

dexterbg:
The charger efficiency is currently meant to be the ratio of energy actually stored into the battery vs the energy fed into the charger. If there are other energy consumers needed for the charging process (for example cooling, charging the 12V battery), they need to be considered as charge losses, so should reduce the charge efficieny. After all, the user isn't interested in how efficient the charger could work if it had no car attached to it.

We may extend the model to provide some detailed energy use metrics of auxiliary electronics & devices later on. At the moment, you can set/translate them to the 12V current metric, which is (together with the 12V voltage) meant to show how much overall power is going into auxiliary devices.

SokoFromNZ:
Okeydokey, will change my code according to this and create a new PR.

drc38:
Hi Michael,
In terms of the new standard metrics power flow definitions are the following correct?
`v.c.power + v.b.power == 0` when charging with `v.b.power` sign being negative (if efficiency was 100%)
`v.i.power + v.b.power == 0` when driving with `v.b.power` sign being positive (if efficiency was 100%)

dexterbg:
No, see docs & comments: `v.i.power` is meant to have the same sign as `v.b.power`, both are defined as the **output** power, i.e. negative = input.

The charger power is defined as the **input** power, so is meant to have the opposite sign of the battery power. That will also apply on a V2G energy flow, the battery power will then be positive (output to charger), the inverter power zero and the charger input power negative (= output to grid).

drc38:
Ok thanks, so `v.b.power == v.i.power` with both being positive when driving or both being negative on regen (if efficiency is 100%)

drc38:
Michael, should be ready to merge now. Cheers

PR #439, opened 2020-09-14
Add a PID scanning tool.

chrisstaite:
Not really sure if this belongs in the master, but it was useful for developing the MG ZS vehicle.  I have defaulted it to not be compiled in because I see it as a developer tool.


dexterbg:
Chris,

that's a good idea, but it's not a vehicle type. I would expect this to be part of the reverse engineering toolkit (`re` root command). Not only would it fit better semantically, it would also be available any time on any vehicle.

`re` has already a task context that can be started and stopped, an `obdii` subcommand and OBD2 response recording including a rudimentary PID extraction scheme. So the scanner fits nicely in there.

Command scheme could be something like…

* `re obdii scan start <bus> <txid[-txid]> <polltype> <pid[-pid]>`
* `re obdii scan status`
* `re obdii scan stop`

…maybe with `start` automatically also starting the recording if not already running.

@markwj Mark, what do you think?

Regards,
Michael

chrisstaite:
I have re-written as a command.  I decided not to use the current queue in re as it would have involved too many changes to it I think, so I just made a new one while a scan was running, should be ok.

PR #443, opened 2020-09-15
Store the responses of the PID scan and output as part of the status.

chrisstaite:
Analysis of the log messages was difficult, so doing it this way seems more sensible.


dexterbg:
Chris, I'm working on the PID scanner as well (to make it suitable for other vehicles). Is this complete for now? If so I'd merge this, after the merge please wait for my changes.

Regards,
Michael

chrisstaite:
Hi Michael, I'm finished now.  Great that you're making it better 👍

dexterbg:
Pushed my changes, you're welcome to have a look.

I'll also add a PID parameter to the tester presence tool, but not today.

Regards,
Michael

PR #458, opened 2020-10-09
New module VWEUP.ALL: Comfort-CAN and OBD connection

sharkcow:
This is a new module for the VW e-Up (seems like we can't get enough... :) which merges devmarxx' VWEUP.T26 module connected to the Comfort CAN via T26A port and SokoFromNZ's VWEUP.OBD module connected to the OBD2 port. Some OBD metrics have already been added, more to come...


SokoFromNZ:
Heya sharkcow.

Good idea to make both VWUP possibilities available under one vehicle-ID. I reckon you have tested your new vehicle and the hardware of OVMS if it allows to connect to both (ODB and T26A) at the same time?

You did what @devmarxx and me weren't able to do, nice 👍 
Although you don't use our classes/source directly, you've copied and changed/adapted it, correct?

Imho this is a big No-No ;)
Not that you've copied it, just the fact that there is now [duplciate source](https://en.wikipedia.org/wiki/Duplicate_code) in the code.

So either you use our classes (which is not easily done afaik) or you delete our both vehicles.

What do you, @devmarxx and @dexterbg think?

So long
Soko

devmarxx:
Things are too "beta" at the moment in all three modules and still in rapid development, from my point of view. So I would not want to delete anything, but for now would continue the promising, separated, ways we all go.

That said, I believe a well documented, well managed, merged VWUP.ALL module will be the way to go in future times ...

Chris

SokoFromNZ:
The VWUP.OBD vehicle is neither beta nor in rapid development: Nothing will be added from my point of view and it's running stable since weeks.

As @dexterbg pointed out in the mailing list there is only one way to go: Merge both vehicles and get rid of the other two.

This is only feasible like this:

1. Latest VWUP.* sources need to be merged in the openvehicles/master branch
2. No further development is happening for VWUP.*
3. One of us three creates a fresh source and merges both vehicles to a new one:
   a. One vehicle ID must be created which keeps both sources (T26A & OBD) separated. Other parts, like the configs con_obd/con_t26, need to be merged.
   b. Choosing/activating the busses via Web-UI needs to be (re)implemented.
   c. A little internal VWUP framework needs to be created for accessing the metrics. The current source needs to be changed to this framework. This framework decides if the value of the T26A or OBD will be forwarded to the StandardMetrics/CustomMetrics.
   d. The old/merged sources need to be deleted.
   e. It needs to be decided if (and how) old OVMS/Configs need to be upgraded to the new vehicle.
   f. index.rst needs to be rewritten.
   g. All three of us need to be happy with the merged source code.
4. Once all this is done this fork needs to be tested for each bus separately and both active.
5. A PR for the openvehicles/master branch gets created to make it official.
6. The PR gets merged.
7. Further development on T26A/OBD can continue under the new single vehicle but in the separate sources.

The VWUP.OBD vehicle is ready to start this process as initially said.

devmarxx:
I'm not very happy with the latest commit a0fc766.

I originally started the development of the vw e-up module fpr OVMS under the vehicle module directory name "vehicle_vweup" and I'm still developing code in my repository under this module directory name. I surely can understand that it is better to rename my directory to vehicle_vweup_t26, but as founder of the vwup module I certainly would like to be asked before this is done. 

The "united" vweup vehicle module needs to be understandable for the user and I still don't see that in the latest commit. It is not clear for the user how to connect to the car to make to get T26A working in combination with  OBD. No pictures, no explanation of the wiring. Also no explanation of how to configure the software by the webui or what can be seen in the app and what not.

As long as the "united" approach is not well documented and not user friendly I will stay with "vehicle_vweup" as directory name for my T26A approach and would like to suggest to keep "vehicle_vweup_all" for the "united" approach. Later, when the "united" approach is better documented, we can change the names, but from my point of view not now.

And again I would like to remind that in the master repository we are developing this stuff for users and not just for ourselfs. Users need to be able to easily understand how to connect the OVMS to the car, how to configure the module and what functionality they get within the app.

I'm convinced that we will get there. It takes just a bit more efforts towards to users.

Regards

Chris

sharkcow:
Sorry Chris,

this is just due to my git noobness. I was trying to clean up towards a unified e-Up module in my own repo and didn't realize my commit would show up in the main repo immediately due to the pull request. I will therefore cancel the pull request and will start a new one as soon as we agree the time is right.

sharkcow

devmarxx:
Super. Thanx.

We will get there :-)

Chris

PR #462, opened 2020-11-06
TLS: Embed certificates as DER rather than PEM to save space.

chrisstaite:


markwj:
Sorry for the delay. I'm reviewing this now, and should be able to process it this week.

markwj:
Merged ok to for-v3.3

PR #464, opened 2020-11-06
Net: Support for connection timeout in NetConns.

chrisstaite:
This adds support for connection timeouts (and fixes a race condition with the user data) but I don't think this class is actually used by anything.  Perhaps it should simply be removed?


markwj:
Sorry for the delay. I'm reviewing this now, and should be able to process it this week.

markwj:
Merged ok to for-v3.3

Issue #311, opened 2019-12-27
Start Battery Heating remotely on Nissan Leaf


dalathegreat:
Now when we are approaching the really cold months, it would be extremely beneficial for a Nissan Leaf to have more control for the battery heating. As you northern Leaf drivers might know, if the temperature of the battery drops too low, the Leaf will not fast-charge. What if we could control the heating of the battery like on a Tesla? This would be a game changer! 

First of all, to see if your Leaf is equipped with DC heating elements inside the battery, there is a signal in the 0x5C0 frame that has 0/1 if pads are present or not.
![image](https://user-images.githubusercontent.com/26695010/71509168-08227180-2893-11ea-81d5-d6300553224d.png)

In 0x5C0 frame 0 bit 5, there is a heating **start** send request and in bit4 there is a **stop** heating request.
![image](https://user-images.githubusercontent.com/26695010/71509253-694a4500-2893-11ea-829b-707f8770da70.png)

The theory is that to start heating, you write bit5 to 1. BUT there is also a send request bit that might be necessary to write also. It's in frame 1 bit 0 (Batt_Heater_Mail_Send_Request)
![image](https://user-images.githubusercontent.com/26695010/71509378-c219dd80-2893-11ea-8e28-d32b7d654a85.png)

Anyone got a Leaf equipped with heater pads? :)

The-Alchemist:
@dalathegreat : How can you check if you have those heaters in the battery?  Is that the LEAF Winter Package?

I think I have it, because I've got a 2012 LEAF SL. =)

Could you clarify what you mean by "this should only be possible on AZE0 Leaf"?

Thanks!  Looking forward to working on this. 🎆

dalathegreat:
@The-Alchemist The documentation I found doesn't indicate that the heater pad fields are present on the bus. Maybe the ZE0 (2011-2012) model doesn't communicate the heating with the VCM. Maybe it just handles it internally?

You can try to look at the 0x5C0 frame4 bit0, if it's high then you have the heater. This worked on AZE0 (2013-2017) atleast.

mjkapkan:
This is on my to-do list as well. I already have a plugin for controlling the heating in the cabin with an additional eating element, so using that logic to control battery temperature should be relatively easy. Will try to test sending the command on my Leaf.

mjkapkan:
@dalathegreat Let's add this as new metric. Which one do you prefer?
`heater `or `heaterpresent`?
[...]

PR #477, opened 2020-12-01
fix cabintemp, add battery heater metric

mjkapkan:
- Cabin temperature readings seem to differ between cars. Added configurable offset in NL Features GUI.
- Added battery heater present (yes/no) metric as discussed in #311 `v.b.heaterpresent                        yes`

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -138,7 +138,7 @@ OvmsVehicleNissanLeaf::OvmsVehicleNissanLeaf()
   m_battery_energy_capacity = MyMetrics.InitFloat("xnl.v.b.e.capacity", SM_STALE_HIGH, 0, kWh);
   m_battery_energy_available = MyMetrics.InitFloat("xnl.v.b.e.available", SM_STALE_HIGH, 0, kWh);
   m_battery_type = MyMetrics.InitInt("xnl.v.b.type", SM_STALE_HIGH, 0); // auto-detect version and size by can traffic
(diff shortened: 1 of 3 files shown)


mjkapkan:
@drc38 Added what we discussed on cabintemp topic. Feel free to review.

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
```diff
@@ -1264,6 +1264,11 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
         {
         StandardMetrics.ms_v_bat_temp->SetValue(d[2] / 2 - 40);
         }
+      // Battery Heater existance as reported by the LBC. Frame 4, bit 0.
```
dalathegreat, line 1267: Keep in mind this only works on AZE0 not ZE0, it will say 0 on the ZE0 even though heater is present!

mjkapkan: Ah right, fixed that so that it does not show up positive on Leafs prior to 2013. Will compile, test and add to this pull rq.

mjkapkan: fixed

dalathegreat (approved):
LGTM!

dexterbg:
Vehicle modules shall not create metrics in the standard metrics namespace, so please either move `v.b.heaterpresent` into the `xnl` namespace, or introduce a full standard metrics set for battery heating/cooling in a separate pull request.

A full metrics set would better be discussed on the mailing list first, so please post a proposal. The set should be able to cover all standard cases.

Regards,
Michael

mjkapkan:
@dexterbg Yeah, make sense. Change is too small for now to add standard metric group. Moved to xnl for now.

PR #482, opened 2020-12-08
NL: use setpoint to indicate hvac status

drc38:
change from using fan speed as it can result in hvac reporting as on when it is not

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -952,7 +952,7 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
         fanspeed_int = 0;
       }
 
-      m_climate_fan_speed->SetValue(fanspeed_int);
+      m_climate_fan_speed->SetValue(fanspeed_int / 7.0 * 100);
       m_climate_fan_speed_limit->SetValue(7);
 
       bool fan_only = (d[1] == 0x48 && fanspeed_int != 0);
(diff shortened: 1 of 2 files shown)


mjkapkan:
We have extensively tested the behaviour with @ankohanse last year and it seem to work for both 2013 and 2016 models. Under what conditions you see it as ON?

drc38:
@mjkapkan here are my metrics when it reports as hvac on, with a solar panel if cabin temp is high does run fan at low speed for circulation?
v.e.alarm
v.e.aux12v
v.e.awake                                yes
v.e.c.config
v.e.c.login
v.e.cabinfan                             3%
v.e.cabinfanlimit                        7
v.e.cabinintake                          fresh
v.e.cabinsetpoint                        0°C
v.e.cabintemp                            25.5556°C
v.e.cabinvent                            face
v.e.charging12v                          yes
v.e.cooling                              no
v.e.drivemode
v.e.drivetime                            0Sec
v.e.footbrake
v.e.gear                                 0
v.e.handbrake                            yes
v.e.headlights                           no
v.e.heating                              no
v.e.hvac                                 yes
v.e.locked                               no
v.e.on                                   no
v.e.parktime                             612Sec
v.e.regenbrake
v.e.temp                                 23°C
v.e.throttle                             0%
v.e.valet

mjkapkan:
I see. Let's try it out then.

drc38:
@mjkapkan yes that's my view, as you want to know when there's a significant power draw from heating or cooling. Strictly speaking hvac includes ventilation but this appears to be a feature on solar panel models and is not controllable by the user (either in car or with ovms).

drc38:
@mjkapkan also noticed v.e.cabinfan reports 3%, does this metric need scaling based on the fanlimit?

mjkapkan:
@drc38 The cabinfan was added before it was a part of standart metrics, so it actually represents the exact steps you can see on the dash. I personally did not agree with having the metric in percentage as it adds additional calculation requirements to the GUI. And usually what user needs is total steps and current step. But I guess it is now decided so we need to go with the flow. You can do that by multiplying it by 100 and dividing by fanlimit.

drc38:
@mjkapkan for my 2016 model remote cooling wasn't being identified in the code, so a few more tweaks required. <2013 should be unaffected.

drc38:
@dexterbg ok to merge? As I have a couple of other fixes on this file to submit. Thx

PR #485, opened 2020-12-09
ovms_server_v2: change charge current to float

sharkcow:
otherwise the decimals in the android app aren't of any use :)


dexterbg:
Btw, same applies to the charge voltage…

PR #487, opened 2020-12-12
NL: improved climate settings >2015

drc38:
Hi @mjkapkan, please try this on your 2013. I have looked at the d[1] byte to separate out the various climate control functions as follows: a7 .... a0
a7 = 0
a6 = 1 for fan
a4,a5 = 1 for cooling
a2,a3 = unsure but if byte is 0x08 or 0x04 climate control is off 
a1 = 1 for auto hvac mode
a0 = 1 for heating
fan speed d[4] reports even when climate control is off, so this needs to be set to zero from hvac status. I have gone back to hvac status reporting as on when ventilating, heating or cooling as climate control is on when d[1] is not 0x08 or 0x04.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -1067,8 +1067,8 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
         StandardMetrics.ms_v_env_heating->SetValue(heating);
         StandardMetrics.ms_v_env_cooling->SetValue(cooling);
         // The following 2 values work only when climate control is activated while connected to charger.
-        m_climate_remoteheat->SetValue((d[1] & 0x0B) && heating); // needs confirming
-        m_climate_remotecool->SetValue((d[1] & 0x0A) && cooling);
+        m_climate_remoteheat->SetValue((d[1] & 0x0B) == 0x0B && heating); // needs confirming
+        m_climate_remotecool->SetValue((d[1] & 0x0A) == 0x0A && cooling);
         m_climate_auto->SetValue(d[1] & 0x02);
         
         hvac_calculated =  (d[1] != 0x08); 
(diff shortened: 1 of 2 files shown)


mjkapkan:
For me everything used to work perfectly before. It took a lot of time to perfect the metrics last time. Would't it make sense to just separate the >2013 and >2015 metric logic?

dexterbg:
If you create pull requests, I assume they have been tested and are meant to be included in the master. That's what pull requests are meant for.

If you're still testing and just want some other developer to test some change, add your changes to your repositories only and ask the other developers to pull from there. Don't create the pull request until the code is ready to be merged. Changes included in the master will be rolled out to users.

I understand this is not ready to be merged yet?

drc38:
Thanks for clarifying the process, and yes not yet ready to merge.

drc38:
> For me everything used to work perfectly before. It took a lot of time to perfect the metrics last time. Would't it make sense to just separate the >2013 and >2015 metric logic?

Good suggestion, it could be the change in TCU bus for 2016+ vehicles causing my issues. @mjkapkan updated to use the original code for 2013-2015 MY. You can pull from my repo to test / confirm things are back working for you.

mjkapkan:
@drc38 That is not what I meant. Let me explain once more: So, before your changes, we had <2013 version climate metrics and >=2013 metrics. What you did was remove the 2013 metrics and add >2016 metrics. Instead, what we need is to have 3 different methods. Hence first we need to rollback your changes that introduced false values for 2013, would suggest just putting back the old code. Then you can add your new section with the code for >2016 MY below the >=2013 MY code. And then finally we can complete the pull request and hopefully all 3 vehicle models are supported correctly.

drc38:
@mjkapkan not sure I follow,  the 3 different methods are in the if MY<2013, else if (covers 2013-2015), else (2016+) MY logic. The 2013-2015 section is copied from the original code? I'm not a coder so if there’s a better way than if...else if let me know with an example

drc38:
@dexterbg, as 2016 model appears to behave quite differently to earlier models it has taken a while to flush out the different possible states - match features of earlier versions. This PR is now ready to merge in my view but I have not heard back from @mjkapkan to confirm that it causes no issues with his MY. As far as I can tell the code for 2013-2015 MY in the PR is functionally unchanged.

mjkapkan:
Finally had some time to review this. Yes, the <2013 and <2016 code looks unchanged. So if you tested this with >=2016 car we can merge it.

drc38:
@dexterbg all good to merge, thanks

PR #499, opened 2020-12-29
new standard metrics for DC/DC-converter and maintenance/service

sharkcow:
as discussed on the list. I tried to get everything right, incl.
documentation. Hope I didn't miss anything :)
Service values are transmitted via server V2 to match up with the
android app.

### DIFF
--- a/docs/source/userguide/metrics.rst
+++ b/docs/source/userguide/metrics.rst
@@ -145,6 +145,10 @@ v.b.soc                                  96.3%                    State of charg
 v.b.soh                                  85%                      State of health
 v.b.temp                                 13°C                     Main battery momentary temperature
 v.b.voltage                              57.4V                    Main battery momentary voltage
+v.c.12v.current                          7.8A                     Output current of DC/DC-converter
+v.c.12v.power                            0.123kW                  Output power of DC/DC-converter
+v.c.12v.temp                             34.5°C                   Temperature of DC/DC-converter
+v.c.12v.voltage                          12.3V                    Output voltage of DC/DC-converter
 v.c.charging                             no                       yes = currently charging
 v.c.climit                               0A                       Maximum charger output current
 v.c.current                              1.25A                    Momentary charger output current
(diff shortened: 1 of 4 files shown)


dexterbg:
sharkcow,

as this covers two different topics (DC converter & service interval), please split this into two pull requests.

Regarding the DC converter metrics, as there were no objections on the list I think the structure is OK. But I think measuring the power output in kW is a little over the top, as these devices will normally operate far under 1000 watts.

Regarding the service metrics, you've got units in the doc, but not in the code. Also… on a second thought now on the "service days" metric: shouldn't that better be a UTC time? Days is relative here, so would need to be updated by the system if no new data is available for a day or more. A UTC time remains valid. If displaying in days is preferred by users (is it?), clients (Apps) can do the math easily. To keep the semantics identical I would then also redefine the "service range" as "service odo".

Regarding the V2 transmission, AsInt() should here be called with a default value of -1, as I introduced that (see my App changes) as the placeholder for "unknown". By changing the metrics into UTC time & odometer we could theoretically use 0 as that placeholder, but -1 is a good choice to represent "unknown" on normally positive integer values.

Regards,
Michael

sharkcow:
Michael,

agreed on all comments except the odo: what we're displaying is a range left to drive, not a total distance as in an odometer. Thus I'd like to keep that name as "range", ok?

Trying to implement the UTC bit: not knowing anything better, I'm using System.currentTimeMillis() on Android. However, I get a difference of 31 days (too many on android). Any simple explanation for that (or a better time reference)?

Thanks!

sharkcow

sharkcow:
OK, I would now deduce that android is the culprit. If I just do:
long now = System.currentTimeMillis();
Log.d(TAG, "time: " + now);

I get a value of 1606535724792, which is clearly one month behind. The OVMS code seems to be right, so I'll proceed with the PRs (assuming you are fine with the "range" vs. "odo" issue :)

sharkcow:
Duh... the date of my virtual android machine was set wrong... sorry for my babbling on. At least things work now :-)

sharkcow:
split into two new PRs as requested

PR #500, opened 2020-12-30
new metrics for maintenance/service, transmit via server v2

sharkcow:
separate PR only for maintenance metrics as discussed.

### DIFF
--- a/docs/source/userguide/metrics.rst
+++ b/docs/source/userguide/metrics.rst
@@ -197,6 +197,8 @@ v.e.locked                                                        yes = Vehicle 
 v.e.on                                   no                       yes = "Ignition" state (drivable)
 v.e.parktime                             49608Sec                 Seconds parking (turned off)
 v.e.regenbrake                                                    yes = Regenerative braking active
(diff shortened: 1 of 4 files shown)


dexterbg:
Thanks, looks much better. Just one issue left: I think you misread the unit `TimeUTC` to represent the Unix time (epoch), which is understandable given there are currently only `TimeUTC` and `TimeLocal` as explicit timestamp-like units defined.

The right unit (for now) is `Seconds`, as with `m.time.utc`. `TimeUTC` and `TimeLocal` actually get converted to a time of day, with string format scheme `HH:MM:SS`. These are used for scheduler entries (timed charging etc.).

Another option, if you'd like to go that far: add unit support for `DateUTC` & `DateLocal`, and/or `DateTimeUTC` & `DateTimeLocal`.

Regards,
Michael

sharkcow:
Thanks for your feedback. Just corrected the units as a quick fix, please merge. If I find more time, I'll look into the unit support.

dexterbg:
No changes, did you push?

sharkcow:
Oops, sorry :-/ Now I did :)

PR #506, opened 2021-01-04
Initial BMW i3 / i3s support

elbow:
Support for the BMW i3/i3s vehicles.
Probably also a good basis to do the Mini.


elbow:
Documentation still to follow (can't get Sphinx to work on my mac so I'll do it on another box).

slcasner:
Sphinx worked for me on my MacBook after I installed all three files mentioned in docs/README.md:

pip install -U sphinx
pip install sphinx_rtd_theme
pip install m2r

(I'm still running Mojave because I need 32-bit support, and my default python is still 2.7.)

elbow:
OK - ready please to have a look at with a view to merging.

Thanks,
Steve

elbow:
Don't merge just yet - want to tweak metric polling intervals first

dexterbg:
Steve,

awesome work! Looking forward to merge. I've got just a few questions & issues:

- You're including a bunch of JSON files defining the available PIDs, which is great (if only we had stuff like this for all vehicles…), but I need to ask: are those public domain? Are they allowed to be published under the OVMS license? And if so, don't you need to include some info on their origin and/or terms of use?
- Some of the JSON files and the generated code from them have character encoding issues in the german parts. Looks like a double UTF8 encoding. Can you please recode these into clean UTF8? It's proably best to simply run all files through a UTF8 fixer like ftfy. Examples:
  - https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/ccfa47197fe80be77fe2ac07db63e4b8ec7cebd9/vehicle/OVMS.V3/components/vehicle_bmwi3/dev/amphhb20.json#L29
  - https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/ccfa47197fe80be77fe2ac07db63e4b8ec7cebd9/vehicle/OVMS.V3/components/vehicle_bmwi3/ecu_definitions/ecu_bdc_code.cpp#L16
- Log tags and metrics names are perfect, but the vehicle code `I3` needs to be changed. We use combinations of the car manufacturer/brand and vehicle type. I'd suggest to use `BI3`, `BMI3` or `BMWI3`.
- As this is already a working module with far more support than an alpha release, please enable the module in the Kconfig by default.
- Please also add an entry to the `changes.txt` file (at the top).

Regards,
Michael

elbow:
Hi Michael,

The JSON files were derived from the GPL 3.0 licenced Ediabaslib project (https://github.com/uholeschak/ediabaslib).  

So they are my creation in a sense, but generated from "PRG" files associated with Ediabaslib.

They are in turn used as input to generate the ecu_definitions, which are the only files required to actually compile it.

Whilst Ediasbaslib is GPL 3 open source, the specific requirements of GPL may complicate your simple MIT licence.  Since they are not required to build or use the vehicle interface I think on reflection it is better that I remove them.

What do you say?

On the other points I'll make the requested changes.

dexterbg:
Steve,

including ressources with other public domain licenses than MIT is perfectly OK (we've got a bunch of different licenses in the web UI assets), but you need to add a README in that directory stating that (see `components/ovms_webserver/assets` for example). It's also good style to reference the source.

Regards,
Michael

elbow:
Hi Michael,

Just to say I used the short code "I3" on the basis of this sentence in the developer guide:

> Decide on a short (2 character) code for the vehicle, that doesn’t conflict with any existing vehicles. This is normally an abbreviation (such as TR for Tesla Roadster).

Anyhow - I'm going to make it BMWI3 then

dexterbg:
Steve, thanks for the pointer, I've just updated & extended that section.

elbow:
OK - ready with requested changes made.
Hoping the UTF8 is now right für dich ;-)

PR #507, opened 2021-01-04
Move the float "smoothed average" consumption to zero avoiding the tiny values we see otherwise

elbow:


elbow:
This branch is relative to the i3 one.  Difference is just the last commit.

dexterbg:
Steve, I guess this PR can be closed?

elbow:
I'll close this and make a new PR with the change since this is hopeless mixed up with the other I3 one.

PR #509, opened 2021-01-06
VWUP: cleanup before re-formatting

sharkcow:


devmarxx:
I'm having a little problem with the changes on t26 wake/sleep.

StandardMetrics.ms_v_env_on->SetValue(false);

should not be set on line 453

We should not be interfering with the "Ring" if the vehicle is turned on. See line 570 of the unmodified code. What we should do is prevent the start of the "Ring" at line 631 of the unmodified code if the vehcle is turned on.

The "Ring" is very, very tricky ...

Chris

sharkcow:
undo T26 change

PR #511, opened 2021-01-06
VWEUP: T26 -Minor bugfix preventing wake up when car is on plus beautified code

devmarxx:
Changed:
components/vehicle_vweup/src/vweup_t26.cpp
components/vehicle_vweup/src/vweup_t26.h


devmarxx:
Title says it all :-)

PR #515, opened 2021-01-08
Round v.b.consumption so as to avoid the tiny numbers that result from the moving average calculation

elbow:
Due to the use of a smoothed average, when v.b.power goes to zero the smoothed average v.b.consumption only tends to zero and eventually gives crazy numbers like 1e-45.

This change just chops the v.b.consumption down to 0 once it falls below 1.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -776,12 +776,11 @@ void OvmsVehicle::NotifyBmsAlerts()
 // Override if your vehicle provides more detail.
 void OvmsVehicle::CalculateEfficiency()
   {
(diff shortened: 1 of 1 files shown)


dexterbg:
Steve, this looks like an unnecessarily complicated solution, also the name "efficiency" for the temporary variable is a little bit confusing. Why not simply trunc the precision on the smoothed value? i.e.
```
StdMetrics.ms_v_bat_consumption->SetValue(
  TRUNCPREC((StdMetrics.ms_v_bat_consumption->AsFloat() * 4 + consumption) / 5, 1));
```

elbow:
Hi,

Didn't notice that macro - I'll try it that way.

Steve

elbow:
Hi @dexterbg ,

Just did it with the TRUNCPREC as you suggested.

Steve

PR #516, opened 2021-01-08
Change sm_v_bat_voltage to SM_STALE_HIGH

elbow:
Metric ms_v_bat_soh uses "stale-threshold" SM_STALE_HIGH.
ms_v_bat_voltage is a closely related metric, but was SM_STALE_MID.

This means that when your car is asleep the voltage metric expires but the soh does not.

This impacts the dashboard.

For me it makes since that of the SOH is SM_STALE_HIGH then the voltage should also be HIGH since the SOH fundamentally is derived from the pack voltage.

### DIFF
--- a/vehicle/OVMS.V3/main/metrics_standard.cpp
+++ b/vehicle/OVMS.V3/main/metrics_standard.cpp
@@ -72,7 +72,7 @@ MetricsStandard::MetricsStandard()
   ms_v_bat_soh = new OvmsMetricFloat(MS_V_BAT_SOH, SM_STALE_HIGH, Percentage);
   ms_v_bat_cac = new OvmsMetricFloat(MS_V_BAT_CAC, SM_STALE_HIGH, AmpHours);
   ms_v_bat_health = new OvmsMetricString(MS_V_BAT_HEALTH, SM_STALE_HIGH);
-  ms_v_bat_voltage = new OvmsMetricFloat(MS_V_BAT_VOLTAGE, SM_STALE_MID, Volts);
+  ms_v_bat_voltage = new OvmsMetricFloat(MS_V_BAT_VOLTAGE, SM_STALE_HIGH, Volts);
   ms_v_bat_current = new OvmsMetricFloat(MS_V_BAT_CURRENT, SM_STALE_MID, Amps);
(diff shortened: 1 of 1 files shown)


dexterbg:
Steve, your assumption is wrong. The SOH isn't derived from the pack voltage, it's an intrinsic metric of the battery pack defined by it's calendary & cycle age plus possibly cell defects and more. The voltage is changing coninuously, the SOH is normally stable over long periods of time.

elbow:
HI Michael,

Yes - I do hear you.  I just meant really that they tend to follow one another.

Fundamentally here's the issue I'm trying to resolve:

My car was sitting overnight.  

I open the web interface on the dashboard and I see the SOH gauge show say 80% but the voltage shows 0.

The car is asleep and the data is old, but since the car is just sitting the SOH is nearly right and since the car is just standing a voltage that was captured just before it went to sleep will be very close to the voltage the next morning.

So for me it makes since that if the SOH still shows then the pack voltage should still show.

Steve

dexterbg:
I now think you maybe confused SOH (state of health) with SOC (state of charge)…?

elbow:
Oh boy - I'm sowing all kinds of confusion with a typo/brainfart.

In my PR explanation I mis-wrote SOH where I meant SOC.

Let me write it again:

Metric ms_v_bat_soc uses "stale-threshold" SM_STALE_HIGH.
ms_v_bat_voltage is a closely related metric, but was SM_STALE_MID.

This means that when your car is asleep the voltage metric expires but the soc does not.

This impacts the dashboard.

The default looks like so:

image

Voltage at the top and SOC at the bottom.

So what I'm trying to fix is that in the morning when the car was asleep for the night then the SOC gauge still shows the value, the Voltage shows zero.

dexterbg:
Steve, thanks for the clarification.

The reason your voltage drops to zero over night isn't the staleness. A stale metric doesn't lose its value, it just gets tagged as stale. Actual example from my car:
```
OVMS# me li -s v.b.voltage
[99S] v.b.voltage                              342V
```
…and of course the web dashboard shows the same value.

I suggest checking your poll processing code, maybe you set the voltage from an invalid response, i.e. some zero during bus/device shutdown?

Regarding the staleness change: the battery voltage is technically much more volatile than the SOC, the voltage staleness time could argued to be lowered, but not raised.

Regards,
Michael

PR #517, opened 2021-01-08
VA: new metrics, BMS metrics

kssmll:


dexterbg:
Alexander, thanks for the submission & fix of the SWCAN module, I missed checking the build with that module enabled.

Regarding your extensions to the VA module, I think these are worth to be mentioned in the user history (file `changes.txt`).

I'll merge now, please make another PR for the history addition.

Regards,
Michael

kssmll:
OK, done.

Regards,
Alexander

PR #519, opened 2021-01-08
NL: Simplify fan speed read

dalathegreat:
Use actual bits for fanspeed, remove unused code and old todo comment


mjkapkan:
Nice! That was my to do. Thanks for taking care @dalathegreat !

PR #527, opened 2021-01-11
bmwi3: Polling tweaks 

elbow:
Adjust polling intervals
Private metric xi3.s.age - how many minutes since we receieved fresh data from car
Doc tweak to document that and car behaviour when charging.


dexterbg:
Steve, regarding the "age" metric, I assume you want to have a single metric to have that value for all your metrics. But if you've got a metric that's updated in any case, that is superfluous: any metric can give you it's age and last modification time.

Regards,
Michael

elbow:
Hi,

The idea was simply to have a simple and single "staleness" value to show users so they have a way to judge when we last heard from the car.  Since in the app you currently see "Live" but whilst the OVMS might be connected and sending data that data could be out of date.

PR #530, opened 2021-01-14
NL: Save memory by casting as uint8_t

dalathegreat:
Save tiny amount of memory by defining 3x variables as uint8_t instead of int. A range of 0-255 is enough for these variables.


mjkapkan:
Is one of those mV diff between the cell pairs? If so I have seen more than 400mV on some cars.

dexterbg:
Daniel, it's OK to change these variables from int to uint8_t to properly represent their value scopes, but it won't save any memory. The ESP32 has a minimum register width of 32 bits, and aligns all variables to 32 bit addresses except when typed explicitly as packed (`__attribute__ ((__packed__))`). Packing can reduce the memory footprint of e.g. a struct, but comes at the cost of additional operations needed to access members, as the ESP32 then needs to mask and shift bits.

dalathegreat:
@dexterbg Ah, I see, thanks for the info! I'm not so familiar with the ESP32, so I guess we need to save 32bits before we really notice any improvement? Maybe just close this whole PR then?

@mjkapkan the i does not store the millivolt, that is done in the `int millivolt` variable

dexterbg:
I'll leave that decision to you, but to me it looks like only the third change actually matches the value scope. Index loops using ints are perfectly OK.

PR #532, opened 2021-01-15
KS: Modify BMS call depending on battery size

dalathegreat:
Attempted fix for the 30kWh Kia Soul, so that it gets 100S BMS arrangement instead of 96S.

This change is not tested, but it compiles. Can some more experienced dev say if this is OK? Constructor only called once, so requires reboot?

.bin file also included if someone with Kia Soul 30kWh wants to test.


dalathegreat:
[ovms3.zip](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/5820922/ovms3.zip)

mnmn2:
Thank you very much!
I can test it tomorrow.

dexterbg:
Looks like a use before init case: `ks_battery_capacity` is initialized by `ConfigChanged()`, which is called after your lines.

dalathegreat:
So we could just run the check after the ConfigChanged(NULL); line? If so, I'm attaching that as a file you can test @mnmn2 !

dalathegreat:
[build_initfix.zip](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/5822407/build_initfix.zip)

mnmn2:
> So we could just run the check after the ConfigChanged(NULL); line? If so, I'm attaching that as a file you can test @mnmn2 !

ok, thanks, tomorrow I will test it :)

dalathegreat:
Ping @mnmn2 , any luck with the testing? :)

dexterbg:
Is this work in progress or mergable?

dalathegreat:
No this is very WIP, wouldn't want to merge anything untested. Ping @mnmn2 or any other Kia Soul owner who could test this?

bofi99:
Hi. I own an Kia Soul MY2018 30kWh. I recently bought an ovms and like to support your developmemt.
My skills programming and scripting are very low (0) but i would be an good guinea pig and test some experimental builds.

dalathegreat:
Hi @bofi99 , the attached file build_initfix.zip in a few responses up could still be loaded onto your OVMS, easiest to just load it onto a sd-card.

When you get it installed, please check if the BMS monitor shows 100 cells in the web-view!

bofi99:
Hi @dalathegreat. Thank you. I will try the build!

bofi99:
# @dalathegreat no luck:
Screenshot_20210628_180453_com android chrome
there is no data.

bofi99:
Hello
under metrics list there is no data at:
v.b.c.voltage
v.b.c.voltage.alert
v.b.c.voltage.dev.max
v.b.c.voltage.max
v.b.c.voltage.min

dalathegreat:
Hi, thanks for testing! The only change was this:
image

But it seems like this then broke the whole BMS section. I don't really understand why, so maybe some more experienced dev can step in?

dexterbg:
If that was the only change, the reason is simple: if you configure a layout of 100 cells, but only fill 96 voltages, the BMS will never consider the voltages to be complete.

- https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_kiasoulev/src/ks_can_poll.cpp#L307
- https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_kiasoulev/src/ks_can_poll.cpp#L319

PR #537, opened 2021-01-22
VWeUP/OBD: add service notifications

sharkcow:
workaround to exclude values of 0 seems to be necessary (haven't found the cause yet)


sharkcow:
threshold wrong, gotta fix... :-/

PR #538, opened 2021-01-23
mqtt/v3: add more control of publishing intervals.  Also SendAll doesn't send metrics for which we have no value

elbow:
- When sending all metrics we don't send those with no value.
    Helps ensure we only feed valid data
  - New updatetimes.  If they are not set they all use updatetime.idle
    so behaviour doesn't change unless you set these:
      updatetime.charging: update interval to use when the car is charging
      updatetime.on: update interval to use when the car is on
      updatetime.sendall: set this and all metrics will be sent at this interval
        this is helpful for charting to make sure there are regular data points.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp
@@ -197,6 +197,10 @@ OvmsServerV3::OvmsServerV3(const char* name)
   m_streaming = 0;
   m_updatetime_idle = 600;
   m_updatetime_connected = 60;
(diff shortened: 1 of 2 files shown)


elbow:
The "sendall" doesn't bootstrap properly - I'll update and send another PR

dexterbg:
As this is a new feature users should know about, please add an info to `changes.txt`. Btw: as the file grows at the top, I insert changes at the top as well.

You could also consider adding the new interval configs to the web UI (Config → Server V3).

dexterbg:
Side note: maybe you'd also like to add a separate update interval for the `aux12v` state (vehicle base system awake).

elbow:
Thanks for the input.

PR #552, opened 2021-01-31
VWeUP/OBD: set charge duration to 0 on charge stop

sharkcow:
so e.g. Android app stops displaying remaining charging time after charging stops.


dexterbg:
sharkcow, as the car doesn't provide updates to the time prediction when not charging, you're right. But I suggest moving the ETR reset to line 328, so the time will remain set as long as the charge port is (virtually) open. That way the time will remain available in case of a stopped / interrupted charge.

PR #554, opened 2021-02-01
MGEV Updates to Jan 2021

brickmen:
Complete rebuild of the logic that controls OVMS for the MG. Now it is all based around the 12V battery state to avoid risk of flattening the 12V.

Fixed the Zombie Mode override logic and tested successfully to capture time delayed charge sessions.

Added some new knowledge and workarounds for the MG ZS including the buggy R40 BMS firmware.

This new logic has been trialled successfully by Discourse users to fix a few small issues for the past month.

Updated documentation.


dexterbg:
Mike, the new code throws a lot of missing initializer warnings. It seems you forgot to update your polling table for the protocol extension. Please fix.

dexterbg:
Mike, Peter fixed this in
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/556

PR #555, opened 2021-02-05
Cumulative charge energy

GreyPeter:
Only added a few lines of code to allow the display Charge kWh during a charge. Takes a while to show on the app as it needs to go over 1kWh of charge to trigger the display


dexterbg:
Peter, the charge_kwh field in the status message has a resolution of 0.1 kWh, but the message is only sent on important changes and for regular periodic updates as configured. The default update period with a connected client is 60 seconds (see Config → Server V2).

PR #556, opened 2021-02-05
poll_pid_t code modified for updated declaration

GreyPeter:
Looks like the declaration for poll_pid_t has been changed in vehicle/vehicle.h. See OvmsVehicleRenaultTwizy::ObdRequest() for an example.


dexterbg:
Peter, thanks for fixing this.

https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/554#issuecomment-771722604

PR #559, opened 2021-02-07
Added code so Charge kwh reset to zero when charge complete

GreyPeter:


dexterbg:
Peter, you're free to do that, but as a user I'd rather be able to see how much I charged after finishing/stopping the charge. We normally clear the value on closing the charge port / starting the next drive. Clearing the values immediately also removes the value from the status notification.

Are you sure about this change?

GreyPeter:
I was thinking about that myself. I will consider it a bit more. Maybe I will reset it at the start of the next trip. I will close the pull request.

PR #562, opened 2021-02-09
VWeUP/OBD: add service notifications

sharkcow:
Second try :)


dexterbg:
Nearly there: the subtype "Service" for the notification needs to be refined. Subtypes are a) by convention all lowercase, and should b) include an area/detail indicator to allow precise filtering. A good choice for these could be "serv.range" & "serv.time" to reflect their respective metric names, you can also write "service.range" & "service.time".

sharkcow:
done :)

PR #567, opened 2021-02-13
SmartED and ZOE: 

Dimitrie78:
SmartED: add Brabus Drivemode and change Automatic set recu.
Zoe: add BMS Cell Voltage and Temps, fix bat temp and wakup command


dexterbg:
Is this meant for release 3.2.016 (i.e. has been tested already)?

Also it seems this has some new user visible features, if so please add an info to `changes.txt`.

Thanks & regards,
Michael

Dimitrie78:
Yes i have tested it already and update changes.txt

PR #571, opened 2021-02-15
Feature/disp charge metrics

GreyPeter:
Adds new Charging Metrics page.

Image 15-2-21 at 7 50 pm


dexterbg:
Peter,

1. Setting the cell standard deviation metrics to the voltage/temperature ranges (max-min) is wrong both mathematically and in terms of the vehicle BMS framework. You normally do not set these metrics yourself but let the BMS framework determine them -- just as the min & max values. I see you don't get the current values of the cells, only min & max. If you don't want to use the min value as a replacement for the current value, you should still calculate an actual standard deviation of either min or max, or maybe an average of both. In any case, please don't use standard metrics to store other values than they're meant for. If you need to have a metric for the min-max range (redundant in my opinion), define a custom metric for this.

2. Please fix the units in your remarks:
```
            // Convert battery power (kW/hr) to kW/sec
            // Convert Ah to Amps/sec
```
It's kWh, kWs and As. Not "per". I know it's just comments, but it's public code.

3. Please add an entry to `changes.txt`, as these are user visible features.

Thanks & regards,
Michael

GreyPeter:
Removed standard deviation code and will revisit this at a later date. Updated comments in processEnergy(). Added features to changes.txt.

PR #575, opened 2021-02-18
bmwi3 only init v e on if no value

elbow:


elbow:
all my merges in from upstream master made a mess

PR #576, opened 2021-02-18
Only initialise v.e.on in module startup if it currently has no value.

elbow:
this avoids overwriting the value now saved through persistence.

Did the same for v.e.awake though not sure it it is persisted.
Harmless if it is not.


dexterbg:
All state flags are persistent now.
List of all persistent metrics: `metrics list -p`

Also a hint on your "merge mess": use rebase whenever possible.
My standard update command: `git pull --rebase --autostash`

leres:
> All state flags are persistent now.
> List of all persistent metrics: `metrics list -p`

Ah, and that includes v.e.on which is why you previously suggested:

> A better init option:
> 
>    if (!StdMetrics.ms_v_env_on->AsBool())
>      StdMetrics.ms_v_env_on->SetValue(false);

Otherwise ms_v_env_on might get set to false/off if we reboot while running, right?

dexterbg:
Exactly.

PR #582, opened 2021-03-06
Report invalid can_tx() CAN ID and data values

leres:


vehicle/OVMS.V3/components/can/src/can.cpp
```diff
@@ -204,11 +209,24 @@ void can_tx(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const
...
+    writer->printf("Error: Invalid CAN ID \"%s\" (0x%lx max)\n", argv[0], idmax);
+    return;
+    }
+  frame.MsgID = uv;
   frame.callback = NULL;
   for(int k=0;k<(argc-1);k++)
     {
-    frame.data.u8[k] = strtol(argv[k+1],NULL,16);
+    uv = strtoul(argv[k+1], &ep, 16);
+    if ((*ep != '\0' && !isblank(*ep)) || uv > 0xff)
```
dexterbg, line 224: Please explain the purpose of allowing blanks in octet args

leres: I was initially confused about how arguments were parsed; this is left over from when I thought can_tx() worked; will fix.

leres: I amended/re-tested my pull request.

PR #588, opened 2021-03-13
Protect against corrupt can frames; limit the maximum dlc value

leres:
The DLC field is only 4 bits wide (15 bytes max) but it seems like a good practice to limit it to the size of the payload buffer (8 bytes).

The issue is described here:

- https://github.com/ThomasBarth/ESP32-CAN-Driver/issues/26


dexterbg:
So, yet another hardware bug in the ESP32 that could well be responsible for some of the issues.

> Naturally, if length is >8, frame is possibly totally corrupted anyway and the best would be not to queue it at all.

I'd rather go with the option of dropping a frame that's probably corrupted than feeding trashed data to the application. What do you think?

leres:
I was thinking this would be the result of some kind link layer issue.

Personally, I would pass the frame up and let a higher level decide if it's garbage or not. But if you prefer to toss it what kind of error indication should be used? We can't log from an interrupt handler so it seems the options are  silently eat the frame or tally an error counter and eat it.

dexterbg:
The level above the driver is the application, and that needs to be able to rely on the driver delivering valid data only.

We've got the error indication for other kinds of errors already from within the ISR. `CAN_status_t` and the loggers just lack a counter for this specific type of RX error. I'd introduce something like `invalid_rx` or `rxbuf_corruption`.

leres:
How do you want to handle the transmit case?  I had added a check there. An invalid dlc when sending a frame would probably indicate a programmatic error.  It probably can't happen from the cli due to argument checking. Can it even easily happen without going to some work because the struct you'd be initializing only has 8 elements in the  array? If so then there's probably no good reason to check outgoing frames?

dexterbg:
If we should check the TX case, we wouldn't do it in the esp32can driver but in the CAN framework, as that would mean the application is broken and could as well try to transmit broken frames on another bus. But I don't think we need to check the TX case. It's a total edge case if detectable by the DLC. A broken application would normally just write behind the buffer without noticing itself, so wouldn't indicate a wrong DLC.

leres:
How does this look? I removed the TX check and added invalid_rx to CAN_status_t. So now if the DLC is invalid on RX, the new counter is incremented, we request the next frame and continue to the

dexterbg:
Craig,

1. Your changes to `canbus::GetErrorState()` are wrong. Invalid frames must not have any effect on the bus error state, only real rx & tx error counters may be used for this.
2. You don't signal a change of the counter to the ISR & application, so `invalid_rx` will only get logged by chance if another event triggers a status/error log entry. I think such errors should be logged, to be able to track down CAN RX issues resulting from this effect.

You probably can introduce an artificial error IRQ to signal this error.

Regards,
Michael

leres:
>     1. Your changes to `canbus::GetErrorState()` are wrong. Invalid frames must not have any effect on the bus error state, only real rx & tx error counters may be used for this.

Got it; that change reverted.

>     2. You don't signal a change of the counter to the ISR & application, so `invalid_rx` will only get logged by chance if another event triggers a status/error log entry. I think such errors should be logged, to be able to track down CAN RX issues resulting from this effect.
> 
> You probably can introduce an artificial error IRQ to signal this error.

I took a run at this, does it look correct now?

Are the "7 commits ahead" a problem? I tried to squash them but could not figure out how to 
squash "Detect and count invalid SJA1000" into "Detect, report, and discard invalid SJA1000 rx can frames". There's so much I don't understand about git.

dexterbg:
"7 commits ahead" just means there are now 7 commits to be merged into master from your branch.

Sending the log message solves logging these errors, but breaks with the standard scheme of the RX function. That's to first collect error IRQ flags and then send a collective error/status log request after processing all IRQs. The error_flags member currently provides space for 16 IRQ flags, only 8 are defined by the hardware. That's why I suggested defining an artificial error bit for this, e.g. 0x100.

With your separate log message, the event will request an error log entry at least once, but possibly multiple times, if the error occurs multiple times or mixes with other errors. That's probably acceptable, but maybe you want to check if you can extend the error flag collection scheme?

Regards,
Michael

leres:
> Sending the log message solves logging these errors, but breaks with the standard scheme of the RX function. That's to first collect error IRQ flags and then send a collective error/status log request after processing all IRQs. The error_flags member currently provides space for 16 IRQ flags, only 8 are defined by the hardware. That's why I suggested defining an artificial error bit for this, e.g. 0x100.
> 
> With your separate log message, the event will request an error log entry at least once, but possibly multiple times, if the error occurs multiple times or mixes with other errors. That's probably acceptable, but maybe you want to check if you can extend the error flag collection scheme?

I think I've caught up with you. I added __CAN_IRQ_INVALID_RX, how does the current version look?

leres:
(Oops, I hadn't written out a change to a comment...)

PR #590, opened 2021-03-14
Fix the build

chrisstaite:
The default configuration was not updated again.  Would be useful to have the CI enabled for the main repository to verify this for pull requests.


dexterbg:
Oops, sorry Chris, missed that PR… I've already added the MAXED3 line in one of my later commits.

But I didn't see that as a fix, just wanted to enable the vehicle by default as it seems to be mature enough to be shipped on "egde". The automatic build process uses defaults for all missing config options.

PR #598, opened 2021-04-03
Suppress false theft alerts due to a suspected SIMCOM GPS bug

leres:
(the reported location goes from: A,B -> A,B -> 0,B -> 0,A -> A,B -> A,B)


leres:
I forgot to mention I had my third false theft alert last night, hence this pull request...

PR #599, opened 2021-04-03
Skip over non-vin data in first multi-frame response to vin polling

leres:
Both the obdii and chevrolet_c6_corvette vehicles had this problem (at least with the two GM vehicles I have access to)

Here is an (anonymized) example captured on a Corvette:

    1617485504.286504 1R11 7E8 10 14 49 02 01 31 47 31
    1617485504.292771 1R11 7E8 21 59 59 32 36 45 39 37
    1617485504.311466 1R11 7E8 22 31 32 33 34 35 36 37

The frames contribute 1G1, YY26E97, and 1234567 to the vin.

While we're here add the -v (verbose) flag to metrics list to make it easier to see this kind of problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,16 @@ status and control the car. This is a foundation that others hopefully will inte
 - Provide documentation in the user guide
 - Use pull requests to submit your code for inclusion
 
+**A note on pull requests:**
+
+Pull requests shall focus on one specific issue / feature / vehicle at a time and shall only mix 
+vehicle specific changes with framework changes if they depend on each other. If changes are not 
+or only loosely related, split them into multiple PRs (just as you would do with commits).
+
+Usage hint: create a branch for each pull request, include only those commits in that branch (by 
+cherry-picking if necessary) that shall be included in the pull request. That way you can push 
+further commits to that branch, Github will automatically add them to an open pull request.
+
 
 ## Donations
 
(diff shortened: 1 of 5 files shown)


slcasner:
Our command parser includes a dynamic usage string mechanism.  The new option flag should be inserted in this line:

  cmd_metric->RegisterCommand("list","Show all metrics", metrics_list, "[<metric>] [-ps]", 0, 2);

And rather than outputting an explicit usage string, you should call:

void OvmsCommand::PutUsage(OvmsWriter* writer)

leres:
Thanks Stephen, this is much better than "Invalid flag" (although I'd like to see the flags before the non-flag arguments in the actual message).

leres:
FYI, the way I discovered this bug was by noticing the ^A (0x01) in a downloaded log message:

    main (build idf v3.3.4-848-g1ff5e24b1 Mar 21 2021 17:25:35),^A1G1YY26E971234567,23,1,C6CORVETTE,Team America,-1,-1

dexterbg:
Craig, two minor issues with this PR:

1. A PR shall focus on one specific change / addition and shall only mix vehicle specific changes with framework changes if they depend on each other. Please split this into two PRs, one for the vehicle changes (they are OK a a combined PR as the change is the same) and one for the metrics change.
2. As you've extended the metrics list command, you should add a real usage info now. That was missing before, but gets worse with each new flag – none of the flags is obvious. For examples, see `obd can1 req dev ?` or `re obd scan start ?`.

Thanks & regards,
Michael

leres:
>     1. A PR shall focus on one specific change / addition and shall only mix vehicle specific changes with framework changes if they depend on each other. Please split this into two PRs, one for the vehicle changes (they are OK a a combined PR as the change is the same) and one for the metrics change.

I've pared down this pull to just the vin fix.

PR #603, opened 2021-04-05
Improve command line argument min/max checks and add metrics list -v

leres:
OvmsCommand::RegisterCommand() tries enforce minimum and maximum arguments
but doesn't handle flags very well. For example:

    metrics list -psS [works]
    metrics list -p -s -S [fails]
    metrics list -p -S [works]
    metrics list -p -S v.t.pressure [fails]
    metrics list v.e.on m.b.12v.voltage [only prints v.e.on]
    wifi scan [works]
    wifi scan -j [works]
    wifi scan -x [ignores invalid flag works]
    wifi scan a [ignores bogus argument]

To fix this count non-flag, positional arguments for use with
the m_min/m_max test in OvmsCommand::Execute(). Require flags to
come before positional parameters. Make OvmsCommand::PutUsage()
public.

Detect invalid flags in wifi_scan(), event_raise(), module_factory_reset(),
and VfsTailCommand::Execute().

Add a verbose flag in metrics_list() to display non-printing characters
and improve usage strings for the metrics list and metrics persist commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ provides data logging to SD card and to a server.
   - [BMW i3 / i3s](https://docs.openvehicles.com/en/latest/components/vehicle_bmwi3/docs/index.html)
   - [Fiat 500e](https://docs.openvehicles.com/en/latest/components/vehicle_fiat500/docs/index.html)
   - [Hyundai Ioniq vFL](https://docs.openvehicles.com/en/latest/components/vehicle_hyundai_ioniqvfl/docs/index.html)
+  - [Jaguar Ipace](https://docs.openvehicles.com/en/latest/components/vehicle_jaguaripace/docs/index.html)
   - [Kia e-Niro / Hyundai Kona](https://docs.openvehicles.com/en/latest/components/vehicle_kianiroev/docs/index.html)
   - [Kia Soul EV](https://docs.openvehicles.com/en/latest/components/vehicle_kiasoulev/docs/index.html)
   - [Maxus eDeliver 3](https://docs.openvehicles.com/en/latest/components/vehicle_maxus_edeliver3/docs/index.html)


slcasner:
I would appreciate time to review this commit more carefully tomorrow.

dexterbg:
Craig,

thanks for joining in on this, but I think you haven't read @slcasner and my thread on option flags yet. See early January on the developer list, topic "Addition to OVMS Developer Guide".

I've got a couple of issues with your changes. On the basic command framework:

- Option flags must not be fixed to generally prefixing all other arguments. Instead they shall normally be placable anywhere, unless they modify the following argument semantics. Some of our simpler commands currently go the easy way of assuming options at the beginning, but that's not the way it's supposed to be.
- Regarding any "-" prefix argument as an option is wrong, it removes all actual values prefixed by "-" (like negative numbers) from the argument space. Deciding which argument is an option currently must be done by the application. We discussed about ways to register optional arguments explicitly, but that's nothing we should change without deep consideration.
- Changing the meaning of the min/max command constraints involves many more commands to be changed. But see above, I don't think we can agree on your change in the first place.

On `metrics list`:

- Your option `-S` now becoming a documented feature should get another option letter. Chosing an upper case option letter for a totally different function as the already present lower case option should generally be avoided and seems to be unnecessary here. As the option is meant to create commands, how about `-c`?
- The new option letter `-v` and the concept of verbosity IMO also doesn't fit the functionality. Verbosity on commands normally means the command will add debug output on what it actually does. In this case you actually mean "encode the values by escaping special characters", so a natural representation could be `-e` for "escape / encode".

If you separate the `metrics list` changes, I can merge these.

Regards,
Michael

leres:
>     * Changing the meaning of the min/max command constraints involves many more commands to be changed. But see above, I don't think we can agree on your change in the first place.

So is the correct way to implement flags for "metrics list" to use a validate function? Would it be ok to allow a min/max of -1 to disable that check when the the validate function will handle this? That would allow positional argument count checks to be done with whatever remains after checking flags.

I assume making OvmsCommand::PutUsage() public is ok.

> On `metrics list`:
> 
>     * Your option `-S` now becoming a documented feature should get another option letter. Chosing an upper case option letter for a totally different function as the already present lower case option should generally be avoided and seems to be unnecessary here. As the option is meant to create commands, how about `-c`?

I wanted -s for "set" but it was already in use. But -c is fine.

>     * The new option letter `-v` and the concept of verbosity IMO also doesn't fit the functionality. Verbosity on commands normally means the command will add debug output on what it actually does. In this case you actually mean "encode the values by escaping special characters", so a natural representation could be `-e` for "escape / encode".

I used -v to mimic "cat -v", cat also as the -e flag so maybe I use that and, "display  a dollar sign (`$') at the end of each line." In fact I'll see how cat deals with characters with the high bit set, just outputting a '.' was kind of a cheat. (Looks like it adds the "M-" prefix!)

I'd like have -e only apply to OvmsMetricString metrics, is there a way to test for this (maybe the typeid operator?)

slcasner:
As I said in the "Addition to OVMS Developer Guide" thread, the concept of option flags does not fit cleanly with the design of the OVMS command parser.  I would prefer that they not proliferate.  But as Craig notes, the current implementations have some problems, so some additional mechanisms are needed.  We have not taken the time to determine the requirements and design a good solution.

As an example of my preferences, rather than "wifi scan -j" I would prefer "wifi scan json" or maybe "wifi scan json-format".  This could be implemented as follows:

static OvmsCommand* json_format;

void wifi_scan(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
  {
  esp32wifi *me = MyPeripherals->m_esp32wifi;
  if (me == NULL)
    {
    writer->puts("Error: wifi peripheral could not be found");
    return;
    }

  me->Scan(writer, cmd == json_format);
  }

  OvmsCommand* cmd_scan = cmd_wifi->RegisterCommand("scan", "Perform a wifi scan", wifi_scan);
  json_format = cmd_scan->RegisterCommand("json","Output in JSON format", wifi_scan);

The usage comes out as:

OVMS# wifi scan ?
Usage: wifi scan [json]
json                 Output in JSON format
OVMS#

This scheme works fine where there are multiple optional arguments of which one may be chosen.  There isn't a mechanism for choosing multiple children of a node, although perhaps we could design such.

slcasner:
On Mon, 5 Apr 2021, Craig Leres wrote:

> So is the correct way to implement flags for "metrics list" to use a
> validate function? Would it be ok to allow a min/max of -1 to
> disable that check when the the validate function will handle this?
> That would allow positional argument count checks to be done with
> whatever remains after checking flags.

The validate function is only called on TAB or when the command has
child subcommands.  It was added to allow argument validation when the
execute command was not being called.  For the "metrics list"
scenarios we're talking about here, the execute command would be
called.

> I assume making OvmsCommand::PutUsage() public is ok.

Yes.

                                                        -- Steve

slcasner:
Regarding requirements for option flags, some comments on Michael's points:

> * Option flags must not be fixed to generally prefixing all other arguments. Instead they shall normally be placable anywhere, unless they modify the following argument semantics. Some of our simpler commands currently go the easy way of assuming options at the beginning, but that's not the way it's supposed to be.

Michael, what are the use cases where restricting option flags to be prefixes would be a problem?  Perhaps for commands that accept multiple value arguments and where more than one value argument needs to allow option flags to prefix it?  When there is only one value argument it's not obvious to me that we would need to allow option flags to come after the value.

> * Regarding any "-" prefix argument as an option is wrong, it removes all actual values prefixed by "-" (like negative numbers) from the argument space. Deciding which argument is an option currently must be done by the application. We discussed about ways to register optional arguments explicitly, but that's nothing we should change without deep consideration.

I agree with this one.  There could be negative value arguments, or string arguments that need to allow the first character to be a hyphen.

> * Changing the meaning of the min/max command constraints involves many more commands to be changed. But see above, I don't think we can agree on your change in the first place.

A while back I did add a feature to allow looking at the whole command space (as registered at the time).  That is the "test commands" command.  It walks the whole command tree and outputs it as a tab-separated spreadsheet format.

leres:
> I'd like have -e only apply to OvmsMetricString metrics, is there a way to test for this (maybe the typeid operator?)

Any hints here?

slcasner:
On Wed, 7 Apr 2021, Craig Leres wrote:

> > I'd like have -e only apply to OvmsMetricString metrics, is there a way to test for this (maybe the typeid operator?)
>
> Any hints here?

Sorry, was watching Hemingway.

Perhaps std::is_base_of<B, D> will work in the OVMS environment?

                                                        -- Steve

slcasner:
Alternatively, if p is a pointer to an object of type OvmsMetricString or a derived class, then

OvmsMetricString* q = dynamic_cast <OvmsMetricString>(p)

will be non-NULL, else NULL.

leres:
> Alternatively, if p is a pointer to an object of type OvmsMetricString or a derived class, then
> 
> OvmsMetricString* q = dynamic_cast (p)
> 
> will be non-NULL, else NULL.

Great, that gives me some things to try; thanks!

slcasner:
Sorry, mis-typed.  Should be:

OvmsMetricString* q = dynamic_cast<OvmsMetricString>(p)

I was sure I  typed that, but I see it's not there.

slcasner:
Oh, I did type it.  Needs to be in code context: 

`OvmsMetricString* q = dynamic_cast<OvmsMetricString>(p)`

leres:
> `OvmsMetricString* q = dynamic_cast<OvmsMetricString>(p)`

That was really close; but  it appears I cannot use dynamic_cast at all! I had:
```
// Only apply -e to strings
OvmsMetricString* q = dynamic_cast<OvmsMetricString*>(m);
if (q == NULL || !eflag)
  {
```
but Mr. Compiler says:
```
... main/ovms_metrics.cpp:136:60: error: 'dynamic_cast' not permitted with -fno-rtti
```
The workaround proposed in an  [arduino.stackexchange.com](https://arduino.stackexchange.com/a/55159) post is to is to  add a class type function to  each class:
```
class A {
    public:
        virtual int classType() { return CLASS_A; }
};
```
But that seems like too much just to use it once.

I also found typeid() but it also needs rtti:
```
writer->printf("[%s]\n", typeid(m));
```
I suppose I should try this on linux, maybe there's something defective with the way I build the xtensa-esp32-elf toolchain under FreeBSD but I suspect the espressif-crosstool-NG config turns it off to save space.

It was worth a try, thanks for the assist.

slcasner:
Sigh, even my third try was wrong because I dropped the '*'.  Anyway, yes, dynamic_cast has a cost in memory and runtime that we (in the compiler) may choose not to pay.

It would be easy enough to add `virtual bool isString() { return false; }` in the OvmsMetric base class and implement it in OvmsMetricString to return true.  There are already other virtual functions in those classes, so no additional cost.

slcasner:
> Ok, I've done that. The issue is c_str() can return UTF-8 characters for temperature metrics (for example). And really -t (was -e but why not display tabs too) is to look for issues with string metrics, e.g. ^A in the vin number.

Then I think your original question was not well-formed since all the OvmsMetric subtypes need to implement AsString().  Are you saying that -t doesn't act on temperature metrics (that aren't OvmsMetricString) but you want it to?  In which case you don't need the isString() check.

slcasner:
Check the indentation of line 112 of ovms_metrics.cpp.

It would be nicer to have net differences applied to a clean branch so that the change/revert doesn't show in the annotate output.  But settle the net change first.

leres:
> Check the indentation of line 112 of ovms_metrics.cpp.

That was one of the last things I fixed; it looks correct in my version.

> It would be nicer to have net differences applied to a clean branch so that the change/revert doesn't show in the annotate output. But settle the net change first.

I've been doing my diffs against a  /tmp checkout of the current master branch.

Is it possible for me to copy my changes to the side, force my fork to be the same as the parent, and then re-commit/push my fork? So I'd end up "1 commit ahead of openvehicles:master"  without breaking this PR? I've definitely squashed commits for a pull request before but the times I've attempted this when I also have merge commits in my fork have not ended well.

leres:
> > Ok, I've done that. The issue is c_str() can return UTF-8 characters for temperature metrics (for example). And really -t (was -e but why not display tabs too) is to look for issues with string metrics, e.g. ^A in the vin number.
> 
> Then I think your original question was not well-formed since all the OvmsMetric subtypes need to implement AsString(). Are you saying that -t doesn't act on temperature metrics (that aren't OvmsMetricString) but you want it to? In which case you don't need the isString() check.

Was it this question?

 > I'd like have -e only apply to OvmsMetricString metrics, is there a way to test for this (maybe the typeid operator?)

What I wanted was to have -e (now -t) only apply to OvmsMetricString metrics; adding isString() as you suggested solved that.

slcasner:
> > Check the indentation of line 112 of ovms_metrics.cpp.
> 
> That was one of the last things I fixed; it looks correct in my version.

As it appears to me in the github diff, the `continue` on 112 is indented the same as the statement before it, but that statement is the "then" clause of an `if` without braces.
> 
> > It would be nicer to have net differences applied to a clean branch so that the change/revert doesn't show in the annotate output. But settle the net change first.
> 
> I've been doing my diffs against a /tmp checkout of the current master branch.
> 
> Is it possible for me to copy my changes to the side, force my fork to be the same as the parent, and then re-commit/push my fork? So I'd end up "1 commit ahead of openvehicles:master" without breaking this PR? I've definitely squashed commits for a pull request before but the times I've attempted this when I also have merge commits in my fork have not ended well.

After Michael has a chance to comment and we're agreed on the changes, then I suggest closing this PR and opening a new one.  I'm not sufficiently skilled in git to squash this to a clean commit, so I'd do it the old-fashioned way.  I would get the diff of the working copy from the master branch like you are doing, reset the working copy to the master branch, make a new branch, and apply the diff.  At that point make one new commit and push that to a new PR on github.

dexterbg:
Sorry, busy week.

Craig, on your current PR first, I've only suggestions left for the `metrics list` changes:

* You've selected `-t` now for the encoding option, taken from the UNIX `cat` command. Your implementation doesn't match the `-t` option on `cat`: as you also add `$` to show line ends, it would rather be the `-A` option. `-t` still is OK for me, as `metrics list` isn't `cat` and `t` is a better mnemonic for `terminal`, but…
  * you forgot to change the short usage string (line 443), and
  * you should use a "speaking" variable name in the implementation instead of `tflag`.
* As Steve mentioned, line 112 (`continue`) has a wrong indentation.
* Your encoder implementation doesn't encode null chars but stops, working on & assuming a C string. As you added this to debug invisible binary data, you should walk through the whole string by index instead. Btw, the encoder could better be refactored into an ovms_utils function, e.g. placed after `json_encode`, to unclutter the `metrics list` implementation.

A side note: adding another standard encoding would normally have been implemented by adding another `As…` method to the metrics classes, as was done with `AsJSON` (and therefore eliminating the need for / replacing the type check method). But as this encoding is very special and without any relevance outside the terminal, restricting the support to `metrics list` is OK.

You asked:

> Is it possible for me to copy my changes to the side, force my fork to be the same as the parent, and then re-commit/push my fork? So I'd end up "1 commit ahead of openvehicles:master" without breaking this PR? I've definitely squashed commits for a pull request before but the times I've attempted this when I also have merge commits in my fork have not ended well.

That's the purpose of stashing and rebasing in git. Stashing stores your open changes on a stack, optionally including untracked (new) files, stash pop reapplies stored changes to the current working tree. Rebasing moves your changes to base at the head of a branch, if possible. Both operations of course may need resolving conflicts. `git pull --rebase --autostash` combines both operations.

Using a dedicated branch (instead of `master`) for each pull request is nevertheless the recommended method. That way you have full control over which commits will be added to the PR. You still can pop stashes onto that branch and rebase that branch onto `master`, or merge in any direction. And you can also cherry-pick single commits from other branches into the PR branch.

It's also a quite natural work flow, as you normally will want to create a branch to develop a new feature anyway.

Regards,
Michael

dexterbg:
Steve,

> > * Option flags must not be fixed to generally prefixing all other arguments. Instead they shall normally be placable anywhere, unless they modify the following argument semantics.
> 
> Michael, what are the use cases where restricting option flags to be prefixes would be a problem? Perhaps for commands that accept multiple value arguments and where more than one value argument needs to allow option flags to prefix it? When there is only one value argument it's not obvious to me that we would need to allow option flags to come after the value.

The purpose of this rule is to not force the user into unnecessary and unusual syntax restrictions. All sophisticated command based platforms support option flags at arbitrary positions. The gold standard to support options is GNU getopt(), which isn't available in the esp-idf newlib, but we should try to come as close as possible.

A typical use case is a complex protocol command that turns out to need a raised timeout in a specific case. With forced prefixing the user needs to insert the option at the beginning, which is cumbersome. If he/she then fails to place the option at the correct place, which is anything but obvious in our command structure as the command may consist of a sequence of words, the command shall still throw an error. But: if the command scans for options anyway to be able to throw errors, there is no point in not handling options at any position.

Regards,
Michael

slcasner:
Michael, I didn't suggest `git stash` because my understanding of it was that it would save the diff from the last commit of the working copy.  (I do use stash fairly frequently myself.)  Is that understanding wrong?

> The purpose of this rule is to not force the user into unnecessary and unusual syntax restrictions. All sophisticated command based platforms support option flags at arbitrary positions. The gold standard to support options is GNU getopt(), which isn't available in the esp-idf newlib, but we should try to come as close as possible.

I don't hold such strong feelings about this. Our command line UI is not an OS shell that's pounded upon by millions of people performing a wide range of tasks.

> A typical use case is a complex protocol command that turns out to need a raised timeout in a specific case. With forced prefixing the user needs to insert the option at the beginning, which is cumbersome. If he/she then fails to place the option at the correct place, which is anything but obvious in our command structure as the command may consist of a sequence of words, the command shall still throw an error. But: if the command scans for options anyway to be able to throw errors, there is no point in not handling options at any position.

In almost all of our commands the sequence of command words comes before any option flag may be entered.  It is only the command arguments that may be interspersed with option flags.

I can see that if you've run a command and discover that you needed to add another option flag to get the desired result then being able to use Ctlr-P to bring back the command as previously entered and just add the option flag at the end is more convenient than needing to back-arrow past an argument value to insert the option flag.

dexterbg:
> Michael, I didn't suggest git stash because my understanding of it was that it would save the diff from the last commit of the working copy. (I do use stash fairly frequently myself.) Is that understanding wrong?

My understanding is, stashing uses the same mechanism as regular commits, i.e. snapshots, not diffs. I think of stashes as local unnamed branches. Applying is done via diff/patch, and as the stashes automatically include an index state commit, the diffs always have a defined reference independent of any commits added after the stashing point.

leres:
> Sorry, busy week.

Did yours involve planned power work including an unplanned generator failure? Heh.

I had planned on spending more time on this this weekend but instead got drawn into rewriting some fancy kqueue code for another project...
 
> Craig, on your current PR first, I've only suggestions left for the `metrics list` changes:
> 
>     * You've selected `-t` now for the encoding option, taken from the UNIX `cat` command. Your implementation doesn't match the `-t` option on `cat`: as you also add `$` to show line ends, it would rather be the `-A` option. `-t` still is OK for me, as `metrics list` isn't `cat` and `t` is a better mnemonic for `terminal`, but…

Sorry, having grown up on BSD, BSD/OS, and finally FreeBSD my mental model for linux is that it is "unix like"; my original code was based on the FreeBSD version of cat -- it didn't occur to me it would  be so different! I've reworked the code to emulate cat -t on an ubuntu system (except it displays \n as ^J instead of outputting a new line).

>       * you forgot to change the short usage string (line 443), and
>       * you should use a "speaking" variable name in the implementation instead of `tflag`.

Got it.

>     * As Steve mentioned, line 112 (`continue`) has a wrong indentation.

I've almost adapted to your coding style but my 40 years of looking at BSD style occasionally blinds me to deviations.

>     * Your encoder implementation doesn't encode null chars but stops, working on & assuming a C string. As you added this to debug invisible binary data, you should walk through the whole string by index instead. Btw, the encoder could better be refactored into an ovms_utils function, e.g. placed after `json_encode`, to unclutter the `metrics list` implementation.

Good point, I have a version that handles that.
 
> A side note: adding another standard encoding would normally have been implemented by adding another `As…` method to the metrics classes, as was done with `AsJSON` (and therefore eliminating the need for / replacing the type check method). But as this encoding is very special and without any relevance outside the terminal, restricting the support to `metrics list` is OK.

I'm happy to do it this way if you like; aside from adding encode_display() to ovms_utils.h and OvmsMetric::AsDisplay() to ovms_metrics.cpp, how do I make it only apply to OvmsMetricString?

But at a minimum I'll add encode_display().

> You asked:
> 
> > Is it possible for me to copy my changes to the side, force my fork to be the same as the parent, and then re-commit/push my fork? So I'd end up "1 commit ahead of openvehicles:master" without breaking this PR? I've definitely squashed commits for a pull request before but the times I've attempted this when I also have merge commits in my fork have not ended well.
> 
> That's the purpose of stashing and rebasing in git. Stashing stores your open changes on a stack, optionally including untracked (new) files, stash pop reapplies stored changes to the current working tree. Rebasing moves your changes to base at the head of a branch, if possible. Both operations of course may need resolving conflicts. `git pull --rebase --autostash` combines both operations.
> 
> Using a dedicated branch (instead of `master`) for each pull request is nevertheless the recommended method. That way you have full control over which commits will be added to the PR. You still can pop stashes onto that branch and rebase that branch onto `master`, or merge in any direction. And you can also cherry-pick single commits from other branches into the PR branch.
> 
> It's also a quite natural work flow, as you normally will want to create a branch to develop a new feature anyway.

Ok, I definitely need to learn this workflow.

I tried `git pull --rebase --autostash` but it didn't do anything. I then tried my original plan of "force my fork to be the same as the parent, and then re-commit/push my fork" using this [github recipe](https://stackoverflow.com/a/42332860) but that seems to have closed this PR... Before I make any commits in my fork, should I just create a branch in my fork and then open a new PR?

I appreciate your and Stephen's patience with me on this.

dexterbg:
Craig,

> I'm happy to do it this way if you like; aside from adding encode_display() to ovms_utils.h and OvmsMetric::AsDisplay() to ovms_metrics.cpp, how do I make it only apply to OvmsMetricString?

No, don't do that: as I said, that would only make sense for an encoding with broad use. This is very special, so should not be added to the general metrics API. (JFTR: you would apply it to `OvmsMetricString` just the same way as the `isString()` now is applied, i.e. override the base class implementation.)

> I tried `git pull --rebase --autostash` but it didn't do anything. I then tried my original plan of "force my fork to be the same as the parent, and then re-commit/push my fork" using this github recipe but that seems to have closed this PR... 

The recipe you've referenced is meant to discard any local changes in `master`. You can use `git stash` to save your open (!) changes, and/or you can copy your already checked in commits to a local branch before resetting `master`.

> Before I make any commits in my fork, should I just create a branch in my fork and then open a new PR?

Yes. Create a branch, commit your changes there, push that branch into your github repo, then create the PR from that branch.

Any new commits to that branch will be added to the PR automatically (by pushing).

After the branch has been merged, you can delete it.

> I appreciate your and Stephen's patience with me on this.

You're welcome.

Regards,
Michael

slcasner:
Should I have suggested IsString() rather than isString() to Craig?

I may have been letting my workday conventions show through rather
than those of the OVMS code base.

                                                        -- Steve

dexterbg:
You're right, OVMS framework convention is upper case initial. Sorry, that slipped my attention.

Craig, can you please add that change to the new PR?

Thanks,
Michael

leres:
> As an example of my preferences, rather than "wifi scan -j" I would prefer "wifi scan json" or maybe "wifi scan json-format".

Is it ok to consider that change for a future PR?

slcasner:
On Fri, 16 Apr 2021, Craig Leres wrote:

> > As an example of my preferences, rather than "wifi scan -j" I
> > would prefer "wifi scan json" or maybe "wifi scan json-format".
>
> Is it ok to consider that change for a future PR?

That would be fine with me, but I don't know what motivations there
might have been for the current design.  Changing this command would
probably require some changes to web code or other parts, so there
would probably be objections.

                                                        -- Steve

dexterbg:
I also would vote against changing an established command syntax. We need to think of these as a API definitions, as commands are meant to be called by plugins. There may be special purpose non public plugins or forks in the field, so if we see a need to change a command syntax, we need to introduce that change in a compatible way, i.e. with a deprecation phase for the former syntax.

In cases like this I don't actually see a need to change anything. Shortcut options are a well-known standard. Keyword options could be introduced as aliases for the shortcuts when we add option support to the command parser.

Regarding that: on the developer list, I already mentioned the famous AmigaOS command template style as a way to get more towards Steve's preferences of verbose option keywords while retaining positioning freedom. The AmigaOS command template engine supports keyword aliases of arbitrary length and keywords being optional themselves if the arguments were given in the template defined sequence.

https://wiki.amigaos.net/wiki/AmigaOS_Manual:_AmigaDOS_Command_Reference
https://wiki.amigaos.net/wiki/Basic_Input_and_Output_Programming#Standard_Command_Line_Parsing

Not saying we should use that engine (I think it still hasn't been open sourced), but it could be an inspiration for our option support.

dexterbg:
That is, if we decide against the `getopt()` or `getopt_long()` way, which is also quite versatile and easy to use.

dexterbg:
There's a re-implementation in the AROS project:
https://github.com/aros-development-team/AROS/blob/master/rom/dos/readargs.c
Their open source license is based on the MPL: https://aros.sourceforge.io/license.html

leres:
Once again I'm biased by BSD where getopt(3) parses option arguments  before positional arguments. But I knew GNU getopt allowed mixing of option and non-option (positional) arguments:

`[https://www.gnu.org/software/libc/manual/html_node/Using-Getopt.html](https://www.gnu.org/software/libc/manual`

- The default is to permute the contents of argv while scanning it so that eventually all the non-options are at the end. This allows options to be given in any order, even with programs that were not written to expect this.
- If the options argument string begins with a hyphen (‘-’), this is treated specially. It permits arguments that are not options to be returned as if they were associated with option character ‘\1’.
- POSIX demands the following behavior: the first non-option stops option processing. This mode is selected by either setting the environment variable POSIXLY_CORRECT or beginning the options argument string with a plus sign (‘+’). 

So I guess BSD is POSIX compliant. But I don't like the idea of getopt() either  (a) making changes to argv or allocating new memory to store the array of positional arguments.

I was also surprised one time when I saw a colleague use a python script I had written and put flags after positional arguments (a clear indication that the argparse module allows mixing!)

Anyway, whatever you guys want to do (or not do).

PR #610, opened 2021-04-14
update energy and coloumb factors for vw e-UP

jagheterfredrik:
Coulomb: Volkswagen defines `0.0018204444`, which is closer to `3909375/2^31`
Energy: Volkswagen defines `0.0001165084`, which is closer to `250200/2^31`

Not sure if it makes sense to keep representing them in this way when it doesn't look as nice and even?
Also as we are on quite powerful hardware it might make sense to just use `double` everywhere?

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_vweup/src/vweup_obd.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_vweup/src/vweup_obd.cpp
@@ -713,8 +713,8 @@ void OvmsVehicleVWeUp::IncomingPollReply(canbus *bus, uint16_t type, uint16_t pi
       break;
 
     case VWUP_BAT_MGMT_ENERGY_COUNTERS: {
-      const float coulomb_factor = 4000000.0 / std::pow(2, 31);
-      const float energy_factor  =  250000.0 / std::pow(2, 31);
+      const float coulomb_factor = 3909375.0 / std::pow(2, 31);
+      const float energy_factor  =  250200.0 / std::pow(2, 31);
       bool charge_inprogress = StdMetrics.ms_v_charge_inprogress->AsBool();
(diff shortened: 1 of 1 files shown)


dexterbg:
Fredrik,

thanks for joining in & welcome to the team. We had the "uneven" factors before, I changed them to the current ones because I was hunting down discrepancies in the observed counters and assumed the factors to be wrongly reverse engineered. Do you have an official VW source for these factors? Please contact me by mail on this: dexter@dexters-web.de

I chose the representation by 2^31 as that seemed to be the source for the strange factors. If yours are the official factors we should just use the numbers.

Using double instead of float has a severe performance impact on the ESP32, as the chip has very limited floating point support. All double operations need to be implemented by software. We only use double where the calculation needs the precision, which isn't the case here. Another general factor for using float on ESP32 is that float reads & writes do not need synchronisation, as all 32 bit memory accesses are atomic.

Regards,
Michael

jagheterfredrik:
> Fredrik,
> 
> thanks for joining in & welcome to the team. We had the "uneven" factors before, I changed them to the current ones because I was hunting down discrepancies in the observed counters and assumed the factors to be wrongly reverse engineered. Do you have an official VW source for these factors? Please contact me by mail on this: [dexter@dexters-web.de](mailto:dexter@dexters-web.de)

Email sent 📫 

> I chose the representation by 2^31 as that seemed to be the source for the strange factors. If yours are the official factors we should just use the numbers.

I also would like the factors to be less strange but.. 🤷‍♂️

> Using double instead of float has a severe performance impact on the ESP32, as the chip has very limited floating point support. All double operations need to be implemented by software. We only use double where the calculation needs the precision, which isn't the case here. Another general factor for using float on ESP32 is that float reads & writes do not need synchronisation, as all 32 bit memory accesses are atomic.

I see, thank for the thorough explanation 👍

PR #615, opened 2021-04-24
CAN: First attempt at porting the first of four espressif TWAI/CAN errata fixes

leres:
For reference, here is Michael's original issue: [https://github.com/espressif/esp-idf/issues/4276#issuecomment-548753085](https://github.com/espressif/esp-idf/issues/4276#issuecomment-548753085) and the commit of interest is [https://github.com/espressif/esp-idf/commit/2f5806092135e3d991057bc06225bdcf536e93a5](https://github.com/espressif/esp-idf/commit/2f5806092135e3d991057bc06225bdcf536e93a5)

Here is a reformatted description of this fix:

> TWAI_ERRATA_FIX_BUS_OFF_REC
> Add SW workaround for REC change during bus-off
>
> When the bus-off condition is reached, the REC should be reset to 0 and frozen (via LOM) by the driver's ISR. However on the ESP32, there is an edge case where the REC will increase before the driver's ISR can respond in time (e.g., due to the rapid occurrence of bus errors), thus causing the REC to be non-zero after bus-off. A non-zero REC can prevent bus-off recovery as the bus-off recovery condition is that both TEC and REC become Enabling this option will add a workaround in the driver to forcibly reset REC to zero on reaching bus-off.

The actual change is simple:

```
    // esp/components/hal/twai_hal_iram.c

    //Handle low latency events
    if (events & TWAI_HAL_EVENT_BUS_OFF) {
        twai_ll_set_mode(hal_ctx->dev, TWAI_MODE_LISTEN_ONLY);  //Freeze TEC/REC by entering LOM
        //Errata workaround: Force REC to 0 by re-triggering bus-off (by setting TEC to 0 then 255)
        twai_ll_set_tec(hal_ctx->dev, 0);
        twai_ll_set_tec(hal_ctx->dev, 255);
        (void) twai_ll_get_and_clear_intrs(hal_ctx->dev);    //Clear the re-triggered bus-off interrupt
    }
```

TWAI_HAL_EVENT_BUS_OFF is set in the routine above, twai_hal_decode_interrupt():

```
    //Error Warning Interrupt set whenever Error or Bus Status bit changes
    if (interrupts & TWAI_LL_INTR_EI) {
        if (status & TWAI_LL_STATUS_BS) {       //Currently in BUS OFF state
            if (status & TWAI_LL_STATUS_ES) {    //EWL is exceeded, thus must have entered BUS OFF
                TWAI_HAL_SET_BITS(events, TWAI_HAL_EVENT_BUS_OFF);
                TWAI_HAL_SET_BITS(state_flags, TWAI_HAL_STATE_FLAG_BUS_OFF);
```

My change looks for the __CAN_IRQ_ERR_WARNING interrupt, for the
__CAN_STS_BUS_OFF bit to be on (indicating bus off), and the
__CAN_STS_ERR_WARNING status bit to be on (indicating error status).

The only testing I've done is to confirm it that ovms can still use the ESP32 CAN bus to communicate with a car.

I think this PR will need some work, for example I haven't addressed, "Freeze TEC/REC by entering LOM".

### DIFF
--- a/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
+++ b/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
@@ -209,15 +209,29 @@ static IRAM_ATTR void ESP32CAN_isr(void *pvParameters)
       xQueueSendFromISR(MyCan.m_rxqueue, &msg, &task_woken);
       }
 
-    // Errata workaround: CONFIG_TWAI_ERRATA_FIX_BUS_OFF_REC
+    // Errata workaround: TWAI_ERRATA_FIX_BUS_OFF_REC
+    //
+    // Add SW workaround for REC change during bus-off
+    //
+    // When the bus-off condition is reached, the REC should be
+    // reset to 0 and frozen (via LOM) by the driver's ISR. However
+    // on the ESP32, there is an edge case where the REC will
+    // increase before the driver's ISR can respond in time (e.g.,
+    // due to the rapid occurrence of bus errors), thus causing the
+    // REC to be non-zero after bus-off. A non-zero REC can prevent
+    // bus-off recovery as the bus-off recovery condition is that
+    // both TEC and REC become Enabling this option will add a
+    // workaround in the driver to forcibly reset REC to zero on
+    // reaching bus-off.
+
     // "Force REC to 0 by re-triggering bus-off (by setting TEC to 0 then 255)"
     if ((interrupt & __CAN_IRQ_ERR_WARNING) != 0)
       {
       uint32_t status = MODULE_ESP32CAN->SR.U;
       if ((status & __CAN_STS_BUS_OFF) != 0 && (status & __CAN_STS_ERR_WARNING) != 0)
         {
-        // Explicitly enter reset mode
-        MODULE_ESP32CAN->MOD.B.RM = 1;
+        // Freeze TEC/REC by entering listen only mode
+        MODULE_ESP32CAN->MOD.B.LOM = 1;
 
         // Re-trigger bus-off
         MODULE_ESP32CAN->TXERR.B.TXERR = 0;
@@ -226,8 +240,8 @@ static IRAM_ATTR void ESP32CAN_isr(void *pvParameters)
         // Exit reset mode
         MODULE_ESP32CAN->MOD.B.RM = 0;
 
-        // Clear the re-triggered bus-off interrupt
-        interrupt = MODULE_ESP32CAN->IR.U & 0xff;
+        // Clear the re-triggered bus-off interrupt, collect any new bits
+        interrupt |= MODULE_ESP32CAN->IR.U & 0xff;
         }
       }
 
--- a/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
+++ b/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
@@ -177,6 +177,19 @@ static IRAM_ATTR void ESP32CAN_isr(void *pvParameters)
   while ((interrupt = MODULE_ESP32CAN->IR.U & 0xff) != 0)
     {
     me->m_status.interrupts++;
+    // Errata workaround: CONFIG_TWAI_ERRATA_FIX_BUS_OFF_REC
+    // "Force REC to 0 by re-triggering bus-off (by setting TEC to 0 then 255)"
+    if ((interrupt & __CAN_IRQ_ERR_WARNING) != 0)
+      {
+      uint32_t status = MODULE_ESP32CAN->SR.U;
+      if ((status & __CAN_STS_BUS_OFF) != 0 && (status & __CAN_STS_ERR_WARNING) != 0)
+        {
+        MODULE_ESP32CAN->TXERR.B.TXERR = 0;
+        MODULE_ESP32CAN->TXERR.B.TXERR = 0xff;
+        // Clear the re-triggered bus-off interrupt
+        interrupt = MODULE_ESP32CAN->IR.U & 0xff;
+        }
+      }
 
     // Handle RX frame(s) available & FIFO overflow interrupts:
     if ((interrupt & (__CAN_IRQ_RX|__CAN_IRQ_DATA_OVERRUN)) != 0)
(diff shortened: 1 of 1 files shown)


leres:
A few more thoughts:

- After I set TXERR to 0 and then 0xff then re-read the interrupt register (IR) to clear the interrupt that is supposed to occur; should I "or" the new value into the previous  value?
- My assumption is there is no side effect to  reading the status register (SR)  as there is with  reading the interrupt register
- I see that `twai_ll_set_mode(hal_ctx->dev, TWAI_MODE_LISTEN_ONLY)` sets the mode register (MOD) to enable "listen only mode" and disable "self test mode", do we need to do that?

dexterbg:
Craig,

> * After I set TXERR to 0 and then 0xff then re-read the interrupt register (IR) to clear the interrupt that is supposed to occur; should I "or" the new value into the previous  value?

The way you currently do it would drop the bus error completely, as the interrupt collector some lines after your insertion will never see the bus error.

If you keep the handling code at that place, you also will drop RX and TX events received just before the bus error and yet to be handled by the driver.

So I suggest rethinking the handling sequence.

> * My assumption is there is no side effect to  reading the status register (SR)  as there is with  reading the interrupt register

That's correct.

> * I see that `twai_ll_set_mode(hal_ctx->dev, TWAI_MODE_LISTEN_ONLY)` sets the mode register (MOD) to enable "listen only mode" and disable "self test mode", do we need to do that?

The controller will continue incrementing the error counters unless you enter listen only mode or reset mode.

According to the SJA1000 documentation, the `TXERR` register can only be written to in reset mode. The ESP32 may allow writes in listen only mode? I suggest checking the original driver for entering reset mode in this case. In any case, I assume either mode is necessary to be able to set `TXERR`.

Regards,
Michael

leres:
> > * After I set TXERR to 0 and then 0xff then re-read the interrupt register (IR) to clear the interrupt that is supposed to occur; should I "or" the new value into the previous  value?
> 
> The way you currently do it would drop the bus error completely, as the interrupt collector some lines after your insertion will never see the bus error.
> 
> If you keep the handling code at that place, you also will drop RX and TX events received just before the bus error and yet to be handled by the driver.
> 
> So I suggest rethinking the handling sequence.

Right, moving the check after the TX/RX handling makes more sense.

> > * I see that `twai_ll_set_mode(hal_ctx->dev, TWAI_MODE_LISTEN_ONLY)` sets the mode register (MOD) to enable "listen only mode" and disable "self test mode", do we need to do that?
> 
> The controller will continue incrementing the error counters unless you enter listen only mode or reset mode.
> 
> According to the SJA1000 documentation, the `TXERR` register can only be written to in reset mode. The ESP32 may allow writes in listen only mode? I suggest checking the original driver for entering reset mode in this case. In any case, I assume either mode is necessary to be able to set `TXERR`.

Here is twai_ll_enter_reset_mode() from the new esp-idf:
```
/**
 * @brief   Enter reset mode
 *
 * When in reset mode, the TWAI controller is effectively disconnected from the
 * TWAI bus and will not participate in any bus activates. Reset mode is required
 * in order to write the majority of configuration registers.
 *
 * @param hw Start address of the TWAI registers
 *
 * @note Reset mode is automatically entered on BUS OFF condition
 */
static inline void twai_ll_enter_reset_mode(twai_dev_t *hw)
{
    hw->mode_reg.rm = 1;
}
```
So it seems likely when we detect the bus off condition we're already in reset mode but it shouldn't  hurt to explicitly set the mode. I also added code to exit reset mode.

What do you think about adding the "reformatted description of this fix" near the top of the module? I'm thinking there is useful info there about the errata fix(es) that should be included somewhere.

dexterbg:
> Right, moving the check after the TX/RX handling makes more sense.

You still clear the other error interrupt flags collected before by overwriting `interrupt` on line 230:

```
        // Clear the re-triggered bus-off interrupt
        interrupt = MODULE_ESP32CAN->IR.U & 0xff;
```

Maybe that could just be an `|=`?

> So it seems likely when we detect the bus off condition we're already in reset mode but it shouldn't hurt to explicitly set the mode. I also added code to exit reset mode.

But… if it's in reset mode anyway at that point, then why does the TWAI driver still set listen mode? Is listen mode possibly a necessity here? Maybe, if in doubt, we should just copy the handling of the Espressif driver.

> What do you think about adding the "reformatted description of this fix" near the top of the module? I'm thinking there is useful info there about the errata fix(es) that should be included somewhere.

Definitely, but not at the top of the module, rather close to the handling section as I did with the RX fixes.

leres:
> You still clear the other error interrupt flags collected before by overwriting `interrupt` on line 230:
> 
> ```
>         // Clear the re-triggered bus-off interrupt
>         interrupt = MODULE_ESP32CAN->IR.U & 0xff;
> ```
> 
> Maybe that could just be an `|=`?

I had the same thought right after I created the PR:

>     * After I set TXERR to 0 and then 0xff then re-read the interrupt register (IR) to clear the interrupt that is supposed to occur; should I "or" the new value into the previous  value?

I made that change. But is there a problem that we might find a new __CAN_IRQ_TX interrupt bit that we ignored because the TX check happens first? Maybe we're not likely to receive new TX packets when we've entered an error?

The errata twai just discards all new interrupt bits in this case, leaving its version of "interrupts" (called events) unmodified.

> > So it seems likely when we detect the bus off condition we're already in reset mode but it shouldn't hurt to explicitly set the mode. I also added code to exit reset mode.
> 
> But… if it's in reset mode anyway at that point, then why does the TWAI driver still set listen mode? Is listen mode possibly a necessity here? Maybe, if in doubt, we should just copy the handling of the Espressif driver.

I changed the code to enter listen instead of reset mode;

> > What do you think about adding the "reformatted description of this fix" near the top of the module? I'm thinking there is useful info there about the errata fix(es) that should be included somewhere.
> 
> Definitely, but not at the top of the module, rather close to the handling section as I did with the RX fixes.

Ok, I added it right before the place where it happens.

dexterbg:
> I made that change. But is there a problem that we might find a new __CAN_IRQ_TX interrupt bit that we ignored because the TX check happens first? Maybe we're not likely to receive new TX packets when we've entered an error?

Pretty unlikely, but still possible. RX as well I think. What about moving the workaround block to the start of the loop, right after counting the interrupt?

leres:
> > I made that change. But is there a problem that we might find a new __CAN_IRQ_TX interrupt bit that we ignored because the TX check happens first? Maybe we're not likely to receive new TX packets when we've entered an error?
> 
> Pretty unlikely, but still possible. RX as well I think. What about moving the workaround block to the start of the loop, right after counting the interrupt?

Ok, I moved the block. This is the first version of the fix that I don't have reservations about.

PR #618, opened 2021-04-25
Use MAP_CAN_SPEED() to get accurately return gvret 33.333 and 83.333 speeds

leres:
canformat_gvret_binary::PopulateBusList12() and canformat_gvret_binary::PopulateBusList3() were just multiplying
bus->m_speed by 1000; the MAP_CAN_SPEED() understands CAN_SPEED_33KBPS and CAN_SPEED_83KBPS

(I'm trying to get savvycan working with a 33.333k GMLAN and I'm pretty sure this is one of my problems)

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/canformat_gvret.cpp
+++ b/vehicle/OVMS.V3/components/can/src/canformat_gvret.cpp
@@ -380,16 +380,16 @@ void canformat_gvret_binary::PopulateBusList12(gvret_replymsg_t* r)
 
   canbus* bus = (canbus*)MyPcpApp.FindDeviceByName("can1");
   r->body.get_canbus_params.can1_mode = (bus->m_mode != CAN_MODE_OFF) | ((bus->m_mode == CAN_MODE_LISTEN)<<4);
-  r->body.get_canbus_params.can1_speed = (int)bus->m_speed * 1000;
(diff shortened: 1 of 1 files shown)


markwj:
This is good. Thanks.

Note that I am also actively working on this code (in particular the canlog framework with issues transmitting frames at high speed - more than one frame in a network buffer received from mongoose). As it is now, the gvret-b support is a lot better than before, and savvycan more stable for me; but still some issues seen.

leres:
I'm just trying to get something working; I'm pretty sure when I initially tested savvycan I streamed from ovms to a socat running on a unix host and then analyzed the resulting file offline. But so far I can't get it to say "connected" even though I can see occasional packets with tcpdump.  But after this PR, I  can occasionally get savvycan to list 33333 as a valid speed.

markwj:
Best to discuss on the developers mailing list if you still have problems. I've committed all my fixes, and running on 3.2.016-156-g1199544f with 'can log start tcpserver transmit gvret-b :23' well. The savvycan connects and reads data fine now. Also the transmit issues I was working on have been fixed.

I think the bus speeds in savvycan are comments only, unless you are trying to configure the speed from savvycan. They don't seem 100% correct to me, but all buses operate fine.

If you still can't get it working, you can try the 'enable console' in the connection settings, to see the GVRET-B messages transferred.

PR #623, opened 2021-05-02
MG:Battery config - Range display in miles

GreyPeter:


GreyPeter:
I have resubmitted this pull request (#619) as it appears to have been overwritten by request #621.

PR #632, opened 2021-05-23
NL: use charge pilot stale to set charge door closed

drc38:


dexterbg:
Derek, did you test that? I think the pilot signal normally should/could be off by the time the charge notification is sent, so you again wouldn't get the charge state variant.

You maybe could use the `NotifiedVehicleChargeState()` hook to copy the pilot state to the charge door state after the notification has been sent.

drc38:
Hi Michael, yes I have tested and it works as intended. When the pilot signal is set to false due to loss of voltage from the charger the charge state "stopped" is set which triggers the ovms notification. The pilot signal has a 2 min delay before it is set to stale which allows sufficient time for the notify to correctly report. On my Leaf it actually takes around 5min before the charge door is set to false because the pilot signal does not go stale until the car stops transmitting a 0x390 message.

mjkapkan:
Hi, is the purpose here purely for displaying the info in the app or there's some "hidden" agenda as well?

drc38:
It's just a work around for the ovms app as there does not appear to be a can message that provides charge door state. Previously both were set at the same time, which because the notify is async sometimes caused a notify with no kWh reported in app. Generally the door will be closed after a charge is finished so the PR code is a good-ish approximation without it showering as being left open for long periods.

PR #633, opened 2021-05-26
serverV2: add v.g metrics

drc38:
@dexterbg, Hi Michael, could you check whether the following is consistent with the rest of the V2 messaging framework. Is it necessary to include the "legacy keys" or just use the string versions of mode, state, substate? 

Is there anything else required to be updated to include the v.g metrics?

### DIFF
--- a/docs/source/protocol_v2/messages.rst
+++ b/docs/source/protocol_v2/messages.rst
@@ -530,3 +530,37 @@ Batch client connections do not trigger any peer count change for the car, but t
 <data> is:
 
 * Number of peers connected, expressed as a decimal string
+
+-----------------------------
+Car export power message 0x47 "G"
+-----------------------------
+
+This message is sent <cartoserver> "C", or <servertoapp> "s" and transmits "v.g" metrics from the vehicle.
+
+<data> is comma-separated list of:
+
+* v.g.generating (1 = currently delivering power)
+* v.g.pilot (1 = pilot present)
+* v.g.voltage (in V)
+* v.g.current (in A)
+* v.g.power (in kW)
+* v.g.efficiency (in %)
+* v.g.type (eg "chademo")
+* v.g.state (eg "exporting")
+* v.g.substate (eg "onrequest")
+* v.g.mode (eg "standard")
+* v.g.climit (in A)
+* v.g.limit.range (in km)
+* v.g.limit.soc (in %)
+* v.g.kwh (in kWh)
+* v.g.kwh.grid (in kWh)
+* v.g.kwh.grid.total (in kWh)
+* v.g.time (in s)
+* v.g.timermode (1 = generator timer enabled)
+* v.g.timerstart 
+* v.g.duration.empty (in min)
+* v.g.duration.range (in min)
+* v.g.duration.soc (in min)
+* v.g.temp (in deg C)
+
+Refer https://docs.openvehicles.com/en/latest/userguide/metrics.html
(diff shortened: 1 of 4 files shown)


dexterbg:
Derek,

the code integration looks good to me, but we shouldn't replicate the old "S" message structure & legacy stuff here. There will be no applications reading this message with their "S" message handler, so we're free to define the record structure as appropriate.

For example, there is no need to include the distance unit here. And we should totally drop converting floats into scaled ints for new messages. We also should not convert float metrics to unscaled ints as done with the climit, it's ok to limit the precision to e.g. 1 digit, but not to drop it altogether.

As you already suggested, we should omit the "legacy keys" altogether here, and further must not apply the charging state & mode default strings to the generator, as those most probably won't fit. Accordingly, you should not add "exporting" to the charge states.

We also should include all v.g. metrics, and we should order them semantically. Maybe you can take my grid log message as an orientation -- it doesn't include all v.g. metrics, but the remaining ones should be easily placable.

Regards,
Michael

drc38:
Michael, thanks for the feedback. I understand most of the above to provide an updated version of the commit with the exception of your last paragraph, is the grid log in reference to the order [here](https://docs.openvehicles.com/en/latest/userguide/metrics.html) or something else? what is a semantic order? If you can clarify a bit more that would be great.

dexterbg:
By the grid log I mean these records:
https://docs.openvehicles.com/en/latest/userguide/notifications.html#grid-history-log

By semantic order I mean this: I've tried to arrange the fields in logical groups, i.e. build a block of all state & mode related metrics, group energy related metrics etc.. Applied to the 'G' record, I mean for example to put voltage, current & power in a group instead of having these spread throughout the message. That results in a more convenient layout for manually processing the table in a spreadsheet.

The 'S' fields have grown over time into the current random looking structure. Can't help later additions breaking the field order on the 'G' structure, but maybe there will be few.

drc38:
Thanks for clarifying, I've set the precision to 1dp on floats as I could not see a reason to go to 2dp. If needed precision could be set differently for each metric. Let me know if there's anything else required.

drc38:
Hi Michael, have you had a chance to review whether the proposed messaging structure is workable?

dexterbg:
Derek,

the message structure looks good to me. You can add the documentation for this structure now.

Regarding the update transmission, the standard transmission should be done once initially (at `m_lasttx==0`, line 1947) and only on major state changes afterwards. Add the appropriate control for `m_now_gen` to `MetricModified()`.

Regards,
Michael

drc38:
Thanks Michael, please take a look at the latest commit to see if everything is covered.

drc38:
I wasn't sure whether it was necessary to include `StandardMetrics.ms_v_gen_kwh` in `MetricModified()` to trigger periodic updates when exporting power or not. Please remove if it is not required. Thanks

dexterbg:
Whatever trigger you add to the `MetricModified()` checks will lead to an immediate update transmission (i.e. within a second), so we should only check for changes there, clients need to be informed about ASAP (like mode changes).

Thinking about the regular updates now I think we should also make that transmission depend on the generator metrics actually having data. I.e. send one initial transmission as done now, and additionally if, say, `ms_v_gen_power` or `ms_v_gen_state` has data.

That way we won't send updates if the vehicle doesn't use the new metrics, but get the standard update frequency if it does.

drc38:
Something along these lines?

PR #639, opened 2021-06-27
CAN: Second of four espressif TWAI/CAN errata fixes: SW workaround for TX interrupt lost

leres:
For reference, here is Michael's original issue:

    https://github.com/espressif/esp-idf/issues/4276#issuecomment-548753085

The commit of interest is here:

    https://github.com/espressif/esp-idf/commit/2f5806092135e3d991057bc06225bdcf536e93a5

Here is a reformatted description of this fix:

    Errata workaround: TWAI_ERRATA_FIX_TX_INTR_LOST

    Add SW workaround for TX interrupt lost

    On the ESP32, when a transmit interrupt occurs, and interrupt
    register is read on the same APB clock cycle, the transmit
    interrupt could be lost. Enabling this option will add a
    workaround that checks the transmit buffer status bit to
    recover any lost transmit interrupt.

The fix involves keeping track of when the tx buffer is in use; look for TWAI_HAL_STATE_FLAG_TX_BUFF_OCCUPIED in:

    esp/components/hal/twai_hal_iram.c

My change adds a place to keep track of the tx buf state (m_state) and looks for the possible lost tx interrupt in ESP32CAN_isr().

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/can.h
+++ b/vehicle/OVMS.V3/components/can/src/can.h
@@ -329,12 +329,15 @@ class canbus : public pcp, public InternalRamAllocated
     CAN_frame_t m_tx_frame;       // saved copy of last TX frame to be used in txcallback
     uint32_t m_status_chksum;
     uint32_t m_watchdog_timer;
+    uint32_t m_state;             // state bitset
     QueueHandle_t m_txqueue;
     int m_busnumber;
 
   protected:
     dbcfile *m_dbcfile;
   };
+
+#define CAN_M_STATE_TX_BUF_OCCUPIED   BIT(0) // transmit buffer is in use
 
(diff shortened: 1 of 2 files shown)


dexterbg:
Two comments:

1. The variable `uint32_t status` is superfluous. I guess the compiler will eliminate it anyway, but I also don't think it adds to code readability.
2. `MyESP32can->m_state |= CAN_M_STATE_TX_BUF_OCCUPIED` is outside of the critical section lock. That's a potential race condition. Simply move it up, directly after the transmit request `MODULE_ESP32CAN->CMR.B.TR=1;`

Regards,
Michael

leres:
I guess I've always put registers into variables before use. Sometimes reading a device register can be slower than processor memory or registers; and sometimes they must be  accessed as a certain word size.

Issue #287, opened 2019-11-20
Possible to add stop charging on Nissan Leaf?


dalathegreat:
0x1BD contains a section to stop charging, by modifying this bit you can stop it.
frame.data[1] = (frame.data[1] & 0xE0) | 2; //request charging stop

I am using muxsan CAN bridges, so I'm unsure if this is possible to transfer over to OVMS. But just as an FYI to you guys more knowledgeable of the OVMS system :)

markfinn:
I couldn't get that (the failsafe request from the HVBAT) to work from OVMS.

But,

`can can1 tx standard 1f2 30 71 60 b0 00 3c 03 88` did work.  It's a Charge_StatusTransitionReqest with a value of 3 -- "stop". 
I'm a little hazy on what device is supposed to be sending this normally.  The reaction of the car was to turn off the EVSE.  I was too far away to hear if any relays in the car flipped.  The charge lights kept showing that a charge was in progress.  I didn't try to re-start the charge from software yet, but pressing and releasing  the EVSE disconnect (not actually unplugging it) did NOT restart the charge, even though the charging status lights went out while the button was pressed.

I generated my message by msg[2] := msg[2] & 0xcf | 0x60, then updating the checksum.

The last byte seems to be a crappy checksum.  for the messages I've seen it's ((sum(all the nibbles in the first 7 bytes) + 0xa) & 0xf) | 0x80

I actually don't think this is the right way to do it, I think maybe faking the HVBAT "charge complete" signal or the (EVC?) Keep_SOC would be better.

@dalathegreat your leaf messages repo 078291e809295311bb3517e872e718576c320f21 has been very helpful.

grnbrg:
I was able to get charging stopped with:

      `can can1 tx standard 1db 00 1a 00 00 00 00 e3`

On the down side, I've collected a P31E1 DTC -- "EV/HEV HV System Interlock Error", but I also set a couple of flags that seemed interesting.  :)

The last byte (For 0x1DB, anyway) appears to be a CRC-8, with a polynomial of 0x85, and an initial value and final XOR of 0x0.  See http://www.sunshine2k.de/coding/javascript/crc/crc_js.html for a CRC tool.

The good news is that we seem to have proved that turn off an ongoing charge is at least _possible_!!

BTW:  The initial issue post by @dalathegreat appears to have a typo -- based on his CAN bus files, the message to stop charging should be 0x1DB, not 0x1BD.

[...]

PR #651, opened 2021-09-15
NL charge stop

MikMudry:
Merged @markfinn 's 1db method to stop charge of Nissan Leaf into current branch.
Issue #287 Possible to add stop charging on Nissan Leaf?


mjkapkan:
I tested this as well. However, the scheduled charging might not work after the car charging was stoped using this method Unless re-plugged. Anyway still worth having, so should be added.

dexterbg:
I would normally ask for some code cleanup here regarding indentation/block style and member variable naming. Also, what is the purpose of `last1db`? Seems to be unnecessary…?

But @mjkapkan, if you're ok with this, I can merge it anyway.

markfinn:
> what is the purpose of last1db? Seems to be unnecessary

It is.  I must have been planing on using it to send one 1db packet out right away when the charge stop command came in, but that doesn't look like I implemented it.

It still would be worth following up on the idea in #287  to try to avoid the car noticing what we're doing and throwing an error.  I don't remember if I got very far in trying it back when it was discussed.

mjkapkan:
Unfortunately my car is under repair after front collision for over 3 weeks now. Hopefully I can get the car back next week and verify the code. Another question,  has anyone tested this on other MY cars?

dalathegreat:
Could @MikMudry remove the `last1db` and we could maybe get this one merged? :) I have been using this method now for 3 years to stop the charge (every single charge!), so it is fair to say that it is not wearing out the contactors like theorized in the linked thread.

dalathegreat:
This PR can be rejected and closed, functionality merged in #662
@dexterbg

PR #657, opened 2021-12-01
Vehicle: add recuperated power to trip report

sharkcow:
I propose to add to the new trip reports the actual recuperated power (when motor is in generator mode) additonally to the recharged power that goes back into the battery (up to now labeled as recovered power).

[If the used standard metric ms_v_inv_power isn't generic enough we could introduce a new metric for the motor power and just link them in the specific vehicle modules.]

### DIFF
--- a/docs/source/requirements.txt
+++ b/docs/source/requirements.txt
@@ -1,2 +1,3 @@
 m2r
 docutils<0.18
+mistune<2.0.0
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,13 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Server V2/V3: added manual update request
+    Note: the servers normally don't need a manual trigger to perform
+    data updates, they listen to metrics changes and events. Use this
+    only if you need extraordinary single updates as fast as possible.
+  New commands:
+    server v2 update [all|modified]     -- Request V2 data update
+    server v3 update [all|modified]     -- Request V3 data update
 
 2021-11-22 MWJ  3.2.018  OTA release
 - Vehicle: added optional automatic trip report generation
--- a/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
@@ -2005,6 +2005,25 @@ void OvmsServerV2::Ticker1(std::string event, void* data)
     }
   }
 
+void OvmsServerV2::RequestUpdate(bool txall)
+  {
+  if (txall)
+    {
+    m_lasttx = 0;
+    }
+  else
+    {
+    m_now_stat = true;
+    m_now_gen = true;
+    m_now_gps = true;
+    m_now_tpms = true;
+    m_now_firmware = true;
+    m_now_environment = true;
+    m_now_capabilities = true;
+    m_now_group = true;
+    }
+  }
+
 OvmsServerV2::OvmsServerV2(const char* name)
   : OvmsServer(name)
   {
@@ -2174,6 +2193,21 @@ void ovmsv2_status(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc
     }
   }
 
+void ovmsv2_update(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
+  {
+  if (MyOvmsServerV2 == NULL)
+    {
+    writer->puts("ERROR: OVMS v2 server has not been started");
+    }
+  else
+    {
+    bool txall = (strcmp(cmd->GetName(), "all") == 0);
+    MyOvmsServerV2->RequestUpdate(txall);
+    writer->printf("Server V2 data update for %s metrics has been scheduled\n",
+      txall ? "all" : "modified");
+    }
+  }
+
 OvmsServerV2Init MyOvmsServerV2Init  __attribute__ ((init_priority (6100)));
 
 OvmsServerV2Init::OvmsServerV2Init()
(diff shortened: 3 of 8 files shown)


dexterbg:
sharkcow, some remarks:

first of all, the energy generated from the motor/inverter is an interesting metric.

But I think it's wrong to relate it to the energy used from the battery, for the same reason you'd be taking this measurement: the energy coming from the motor as a generator only partially flows into recharging (and losses are involved in that process), some part also is used directly for lights, heating etc. The same applies to the energy taken from the battery, not all of that is going to the inverter/motor.

So I would rather relate the recuperation energy (= sum of negative power) to the energy actually going into the inverter, i.e. the sum of positive power.

We can then tell the amount and ratio of energy that was recuperated from movement over the energy used for movement of the vehicle.

You can easily get both energy sums using a generalized form of your changes, and include an improvement of your method: take the mean value of the current and the last inverter power recorded, then apply that to the time interval. If the mean is negative, add it to the recuperation sum, else to the driving sum. Using the mean value avoids needing to track the state and gives a better approximation.

To avoid confusing users, we should keep this clearly separated from the battery energy usage statistics. I suggest this layout:

```
Trip 35.3km Avg 57km/h Alt -207m
Energy 140Wh/km; 11% recd
SOC -15.0% = 66.0%
Range -11.0km = 138.0km
Accel +4.2 / -4.3km/h/s
Motor +3.7 / -1.1kWh (30% recup)
```

Another, probably less confusing, option is to introduce a dedicated "energy details" report, and add other energy consumers (heating, climate control, …) and producers (solar cells…?) to that as well. But as long as we don't have these additional details, it's OK to add this to the trip report.

Regarding the implementation:

1. As explained above, use the mean instead of detecting time segments of all negative power.
2. It's bad you're taking `esp_log_timestamp()` twice -- the process may cross a timestamp border in between. Rather do this along the way I've done it in `CalculateAcceleration()`.
3. Avoid divisions whenever possible, especially floating point divisions, here on `m_recuptimedelta`. The ESP32 FPU has no builtin division, and contrary to expectation the compiler will **not** optimize a division by a constant into a multiplication with the inverse. I've created a test case for this: https://godbolt.org/z/xKzx3aMo5
4. Avoid superfluous logging in potentially time critical sections for production code. Any standard log statement like `ESP_LOGD()` needs to always create the log entry and queue it into the system. The test for the actual log level enabled for a component is only done in the logging context, so any logging statement that isn't excluded during compile time has a performance impact.

I'm worried about performance here because non polling vehicles may be able to update the inverter power metric every 10 ms.

Regards,
Michael

dexterbg:
Another remark on the implementation: you also need to check if the motor energy sum is nonzero in the report, as there may be vehicles that cannot provide the inverter power metric.

sharkcow:
Hey Michael,

thanks for your extensive and valuable comments (as always :)! You're right in every respect, I wasn't really happy with the placement in the trip report myself; your suggestions will make it much easier to read.
I'll get to work on it as soon as I can.

Best regards,
sharkcow

sharkcow:
Hey Michael,

thanks for your extensive and valuable comments (as always :)! You're right in every respect, I wasn't really happy with the placement in the trip report myself; your suggestions will make it much easier to read.
I'll get to work on it as soon as I can.

Best regards,
sharkcow

sharkcow:
OK, I tried to make all suggested changes. Just let me know if anything is not up to speed. :)
Do I need to close this pull request and open a new one?

sharkcow:
Another thought: I would also like to show the percentage of energy actually used for movement during a trip. Any ideas how to call that (and where to put it in the trip report)? Or do you think that's too much?

dexterbg:
…oops, I just learned that Ctrl+Return submits a comment ;)

You can avoid mixing the commits by either using a dedicated branch for the pull request that only includes your changes, or by always rebasing your master instead of merging the master.

Regarding your new code:

1. Please rename the new member variables following the code style and using the prefix grouping you already introduced, i.e. rename `motor_energy` to e.g. `m_inv_energyused` and `recup_energy` to `m_inv_energyrecd`.
2. Don't overwrite these member variables for the output formatting in `CommandStatTrip()`. That invalidates the previous scale of the values, so the variable content semantic becomes dependent on the state, which is generally bad style. The trip status command can also be issued any time, also during a trip, which will produce a total mess this way.
3. Don't precalculate float constants, the compiler will do that just fine from a constant expression, which will be much easier to understand and maintain. Also, replacing float divisions in non performance critical sections like `CommandStatTrip()` isn't necessary and should be avoided if it hurts readability.

That's all I can see for now, maybe more if you resubmit a cleaned PR.

Regards,
Michael

dexterbg:
> Another thought: I would also like to show the percentage of energy actually used for movement during a trip. Any ideas how to call that (and where to put it in the trip report)? Or do you think that's too much?

I think it will be too much for a typical user. You can try to design a message layout and ask some non technical users for their opinion first, but I think we're risking an information overload here.

sharkcow:
Oh wow, sorry; I think I should have known at least some of these things... thanks for your patience! I'll try my best :) 
Would you advise to use a different variable for the displayed value scaled in kWh or should I collect the data already in kWh? (I was a bit worried that we might lose too much resolution in the second case.)

dexterbg:
Collecting in kWh would be totally fine and consistent with the other energy values, and it would make creating metrics from these easier if we decide to do so.

You don't lose precision just by shifting the scale on your aggregation values. Shifting the scale essentially just means changing the exponent. A float is internally represented normalized by a mantissa and an exponent. For example, 1.23 is represented as 123E-2 (123 x 10^-2), and 0.00123 is represented as 123E-5 (not taking into account that IEEE 754 floats use base 2 for the exponent, which isn't relevant here).

So as long as you don't exceed the exponent range and don't need to build sums up to a point, the mantissa does not provide enough bits to represent the total plus the fraction to add, you can use any scale you like.

The right question to ask isn't about the scale, it's about the precision vs. total value needed.

A float (32 bit) generally provides ~7 decimal digits for the mantissa. So the highest value you can sum up to is roughly 7 orders of magnitude times the smallest difference you need to distinguish. If you need more, you can use double precision (64 bit) or split into a combination of a total and a differential counter.

Let's check this case: if we define the lowest power we need to count as 1 kW and the fastest rate to cover as 100 measurements per second (10 ms interval), we get a minimum significant difference of ~ 1 kWh / 3600 / 100 = 2.78E-6 kWh. That means the highest total sum we can aggregate in a 32 bit float will be ~ 27.8 kWh, which is arguably not enough for a typical trip. So we can either alter our defined limits or switch to using double precision floats.

Btw, using double values for the internal aggregation does not conflict with using floats for the metrics. The metric only needs to represent a total value, it does not need to serve as an aggregation counter.

sharkcow:
OK, thanks again for the detailed advice and explanations! I have created a new branch, rebased and made the changes as requested (hopefully). Will close this pull request as I have opened a new one for the improved code.

PR #661, opened 2021-12-22
NL: ZE0 charger fix, AC Voltage added, AZE0 cleanup

dalathegreat:
This PR aims to improve three things
- 2011-2012 LEAFs are reported as plugged in all the time (Waiting for ZE0 user to test)
- 2011-2012 LEAFs now report measured AC voltage from CAN message 0x380 (Waiting for ZE0 user to test)
- 2013+ LEAFs now report 220V/110V instead of 200V/100V when charging on L1/L2

For now I have only tested this on my 2015 AZE0. There might be some codechange still needed to display AC voltage properly in app while charging for ZE0, and also we need to see if this fixes the "always plugged in" issue 476

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -351,11 +351,6 @@ void OvmsVehicleNissanLeaf::vehicle_nissanleaf_charger_status(ChargerStatus stat
         StandardMetrics.ms_v_charge_voltage->SetValue(StandardMetrics.ms_v_bat_voltage->AsFloat());
         StandardMetrics.ms_v_charge_current->SetValue(StandardMetrics.ms_v_bat_current->AsFloat());
         }
-      if (m_AZE0_charger)
-        {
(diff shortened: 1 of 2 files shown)


mjkapkan:
Awesome!

dalathegreat:
For any ZE0 testers, I have hosted the .bin here https://dalasevrepair.fi/wp-content/OVMS/ovms3.bin

Just point your OVMS firmware update to the URL :)

mjkapkan:
Tested also on 2013 car. Nice to see 220V, but as mine is 240V, not perfect. How did you find the message for ZE0?

dalathegreat:
Thanks for testing @mjkapkan ! Yeah since the AZE0 cannot measure voltage afaik, static text is the best option. **Should we bump it up to 230V?** Since most of the world that use this voltage either have 230/240V? According to Wikipedia, everything between 208V->240V is a thing 😅

I got the ZE0 AC voltage info from an internal document, all my findings are in the github repo: https://github.com/dalathegreat/leaf_can_bus_messages

drc38:
Here is the output from my charger which has a calibrated meter, hence my view it should be 200/100V as losses will be seen as a voltage drop not a decrease in current. 
Screenshot_20220107-220152_Home Assistant

PR #663, opened 2021-12-28
NL: Add new metrics

dalathegreat:
This PR does some cleanup and adds new metrics for the Nissan LEAF

- m_max_gids (Max number of GIDs the battery can reach (for 2016+))
- m_battery_heatrequested (Battery is requesting heater elements to turn on (for 2013+))
- m_battery_heatergranted (Vehicle is OKing heater elements to turn on (for 2013+))
- m_charge_minutes_3kW_remaining (The 3kW charge time estimate the car calculates)
- m_remaining_chargebars (Remaining charge "bars" on dashboard)

Further development should be done with the battery heating metrics, and alert the user once this is triggered. During colder than -17*C battery temperature, the battery will start to use its own power to heat itself up in the form of 300W heating elements. This will discharge the battery down to 30%SOC if left in the cold for too long, so it would be worthwhile to report this in the app!

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -16,11 +16,11 @@ Open Vehicle Monitor System v3 - Change log
 - Nissan LEAF:
  - Stop charge feature added, works via app
  - 5 new metrics added
-    m_max_gids (Max number of GIDs the battery can reach (for 2016+))
-    m_battery_heatrequested (Battery is requesting heater elements to turn on (for 2013+))
-    m_battery_heatergranted (Vehicle is OKing heater elements to turn on (for 2013+))
-    m_charge_minutes_3kW_remaining (The 3kW charge time estimate the car calculates)
-    m_remaining_chargebars (Remaining charge "bars" on dashboard)
+    xnl.v.b.max.gids (Max number of GIDs the battery can reach (for 2016+))
+    xnl.v.b.heatrequested (Battery is requesting heater elements to turn on (for 2013+))
+    xnl.v.b.heatergranted (Vehicle is OKing heater elements to turn on (for 2013+))
+    xnl.v.c.chargeminutes3kW (The 3kW charge time estimate the car calculates)
+    xnl.v.c.chargebars (Remaining charge "bars" on dashboard)
  - Wakeup command now charges 12V battery if connected to EVSE
  - 2011-2012 LEAF now reads AC voltage from grid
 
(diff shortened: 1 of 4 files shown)


dalathegreat:
@dexterbg Do you have any example code for creating an alert?

mjkapkan:
@dalathegreat wait, so you managed to test the heater activation? 😮 Also I thought that the car will not activate battery heater if the SOC is below 30%. Anyway, great addition. Will be interesting to see how it works (let's hope for some cold nights here in Norther Europe 😅).

dalathegreat:
@mjkapkan Whoops I forgot a '3', well spotted 👍 . I am fairly sure that is how the heating works, but if it is, I think it won't be possible to activate the heating manually. The theorized stateflow is now

- Battery senses -17*C internal cell temperature
- Battery requests permission to turn on heaters
- Vehicle sends either OK or NG
- Battery acts upon the vehicle response

mjkapkan:
Yeah this is what I observed as well. However, it might be possible to spoof the battery temperature. 😈

dexterbg:
> @dexterbg Do you have any example code for creating an alert?

- The fixed string way:
  https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle/vehicle.cpp#L799
- The printf style way:
  https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle/vehicle.cpp#L809
- The internal command result way:
  https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle/vehicle.cpp#L761
- The command call way:
  https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_renaulttwizy/src/vehicle_renaulttwizy.cpp#L392

Regarding the PR, please add the new metrics info to the `changes.txt`, so users will know about this.

Regards,
Michael

dexterbg:
Btw, text & error notifications automatically raise events following the scheme `notify.<type>.<subtype>`, so you don't need to emit additional events if that's sufficient.

…oops, that's actually an undocumented feature, will fix that.

dalathegreat:
Thanks for those examples @dexterbg , this one could now be merged and I'll continue to experiment in a new branch :)

dexterbg:
Actually, sorry, not quite. You described the internal variable names in `changes.txt`, which are irrelevant & unaccessable to users. Please see other additions to the file for reference: this is meant to contain the user level information on changes, so you need to describe metric names, configs, commands etc. as visible for users.

dalathegreat:
Thanks @dexterbg , fixed!

PR #667, opened 2022-01-01
Some tweak on trio

KommyKT:
- Supply new standard metrics
- Persist metrics fix
- Fix charge metrics if crash below charge session


dexterbg:
Tamás,

I can't tell from your source if your new metrics may be relevant to users (and thus should be documented). But at least one of the metrics you removed (`xmi.c.power.ac`) has been documented for users, so you should at least update the documentation accordingly.

Regarding your code: you're checking `has_trip == true`, at some point, but it never gets true. Doesn't look intentional to me.

Regards,
Michael

PR #671, opened 2022-01-03
NL: Reimplement remote heating for 2011-2012 LEAF

dalathegreat:
During the hardware migration from OVMS 2 -> 3, the remote heating/cooling support for 2011-2012 LEAFs broke. This PR re-implements this so that it works with the new v3.X hardware. 

Documents also updated with guides on how to take this into use (with pictures!)

Verified working on my own 2012 LEAF (and verified that nothing broke on my 2015 LEAF).
Test firmware available here: https://dalasevrepair.fi/wp-content/OVMS/ovms3.bin

As a bonus to this PR, when sending 'wakeup' commands to the LEAF, it will now top up the lead acid 12V battery if it is connected to an EVSE.


dexterbg:
Daniel, is this considered to be "just" a fix, or is it a new feature that needs to be mentioned in `changes.txt`?

Also, regarding images for documentation: I'd prefer having them on-site instead of referencing some external forum page, that can be gone any time. It's not hard to integrate images in the vehicle section, see e.g. the VW e-Up section for reference.

Regards,
Michael

dalathegreat:
Thanks @dexterbg , I added two pictures to the documentation and folder. I didn't feel like adding them all since they will clutter the doc easily (and it wasn't done properly with a DSUB26 connector!)

I also updated the changes.txt to reflect what was done. We really need to get a new version released with this, lots of 2011-2012 users will want this asap :)

mjkapkan:
Verified the latest code. Climate Control activation/deactivation works as intended with both 2012 and 2013 model in the config. Also very nice documentation update. Overall very flexible solution for 2012 NL owners. Thanks @dalathegreat !

vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
```diff
@@ -2037,7 +2042,8 @@ OvmsVehicle::vehicle_command_t OvmsVehicleNissanLeaf::CommandWakeup()
   if (!m_enable_write) return Fail; // Disable commands unless canwrite is true
   ESP_LOGI(TAG, "Sending Wakeup Frame");
   unsigned char data = 0;
-  m_can1->WriteStandard(0x679, 1, &data);
+  m_can1->WriteStandard(0x679, 1, &data); //Tops up the 12V battery if connected to EVSE
+  m_can1->WriteStandard(0x5C0, 8, &data); //Wakes up the VCM (by spoofing empty battery request heating)
```
nattgris, line 2046: Noticed while debugging issues preventing climate control, that this seems to transmit garbage data. 8 bytes sent from the single unsigned char. Could this cause the intermittent problems?

nattgris: (referring to the issue #804 )

PR #675, opened 2022-01-04
NL: Restore 2011-2012 charge calculation

dalathegreat:
During the ZE0/AZE0 charge refactor, the ZE0 charge calculation broke and reported wrong values. On top of that, the app started to spam that charge stopped and started. This PR reverts the charge calculation to how it was before (and cleans up some uncommented stuff).


dalathegreat:
Hold on merging this, issue wasn't fixed yet.

Issue #645, opened 2021-07-30
NL: Unable to Start Charge from Android App


mjkapkan:
Sliding the charger plug in the app does nothing in Android app, but works in iOS. Also the charge start command is working as well. So this must be somewhere in the Android app.

glynhudson:
Starting a charge works fine for me with a Nissan Leaf using the Android app. On the Leaf starting a charge only works when the vehicle is plugged in and the charging timer is enabled.

mjkapkan:
Interesting. seems to work now.Might have been some connection issues...

mjkapkan:
Found out what was the problem. The charge port was set to false after 2 minutes even pilot signal is set to true. This caused the slidder to not appear. Now fixed in #676 .

Issue #644, opened 2021-07-30
NL: HVAC metrics incorrect for MY2013


mjkapkan:
It looks like the HVAC gets set to yes every time the car is charging. This was already fixed in #281, but it looks like something broke it afterwards.

glynhudson:
I''ve also noticed this issue

drc38:
The only functional difference I can spot is the code for the fan speed. 
```
int fanspeed_int = ((d[4] & 0x38) >> 3); //use bits 5:3 to read fan speed
m_climate_fan_speed->SetValue(fanspeed_int / 7.0 * 100);
```
versus
```
int fanspeed_int = d[4] / 8;
// todo: actually we need to use individual bits (7:3) here as remaining ones are for something else
if ((fanspeed_int < 1) || (fanspeed_int > 7))
      {
        fanspeed_int = 0;
      }

m_climate_fan_speed->SetValue(fanspeed_int);
```
is the fan speed also showing as >0 when charging?

drc38:
Also in https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/376/files on line 752 should it be a bitwise & or logical &&?
`hvac_calculated = (climate_on & (m_climate_fan_speed->AsInt() != 0));`

mjkapkan:
Yeah, I now remember that CC flag used to be on during charging so I added the fan speed as an additional check (you can never have hvac on if the fan is not running). So this one needs to be logical. I will test how it looks tonight.

mjkapkan:
Checked. So it seems whenever the car is charging we get 57% of fanspeed. We could use this rule+xnl.cc.fanonly metric to mitigate this until we find another solution.

dalathegreat:
I checked this yesterday on my 2015 LEAF running the latest build, fanspeed was at 0%, so no issue here?

mjkapkan:
@dalathegreat Checked just now. Issue still persists. If vehicle is charging, then fanspeed is set to 14%, and hence CC flag is enabled. If you dont't have this issue, then it could be limited to <2015 MY cars.

mjkapkan:
Basically this should be an easy fix. We just need yo find one more metric to validate against.

mjkapkan:
Implemented a hacky fix, le't see if it works for most and then we can close it. Link to beta firmware in PR: #676 .

mjkapkan:
[...]

PR #676, opened 2022-01-05
NL Charge to SOC/Range Feature Enhancement And Rework

### LINKED ISSUE #672: NL: Functionality to Start/Stop Charge Depending on Range/SOC Not Working
Currently the charge is not stopped on Nissan Leaf vehicle even if sufficient range/SOC which is set in the `Battery config` page is reached.

DOD: Implement charge stop command so that it works according to the config page.

mjkapkan:
Resolves #672 and #644. Also fixes the charge substates that now finally should indicate if the charge was stopped by request from app, the auto charge/stop feature or interrupted manually. Also moved charge stop to a remotecommands function along with all others so it's easier to actiavte them from within. Firmware for testing: [click to download](https://mjkapkanovms.s3.eu-north-1.amazonaws.com/ovms3.bin).

Note: It has been quite some changes so let's not merge too quickly. ;)


vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
```diff
@@ -1175,6 +1173,12 @@ void OvmsVehicleNissanLeaf::IncomingFrameCan1(CAN_frame_t* p_frame)
         m_climate_remoteheat->SetValue(d[1] == 0x4b);
         m_climate_remotecool->SetValue(d[1] == 0x71);
         hvac_calculated = (climate_on & (m_climate_fan_speed->AsInt() != 0));
+        //dirty fix until solution is found:
```
dalathegreat, line 1176: Would using CC_ACStatus in 0x54C help? I just found it and added it to the database file: https://github.com/dalathegreat/leaf_can_bus_messages/commit/c36252dd65e5fc8eca57fc8f8bc7b40b3354ad5c#diff-433d4c98e1bbbbe20be37efde34e858531adebe851382b0d331c13c000998450

mjkapkan: Hm, I surely remember dissecting the 54c byte by byte down to bits lol, but you might be right. Will try it out.
image

mjkapkan: Ok @dalathegreat, so now I remembered that this logic is not working while charging:
```
54c byte 2:
10010000 0x90 <-- preheat ON
11111000 0xf8 <-- preheat OFF + 
00000000 0x00 <-- preheat OFF + charging
```
 that's why I didn't use it form the start. However, I managed to make a simpler logic for this to work. So will push an update soon together with some other tweaks and cleanups.

mjkapkan:
Did some more refinement, but ready for testing now. Latest firmware uploaded to the link in the description.

drc38:
The changes look useful. The substate transitions may need some tweaking however. Plugging in to my charger which does not charge until a set time of day no longer transitions out of a powerwait substate.

drc38:
Screenshot_20220107-210126_Open Vehicle

mjkapkan:
@drc38 I see. I think it's best to assign "scheduledstart" substate in that case. would you agree?

drc38:
> @drc38 I see. I think it's best to assign "scheduledstart" substate in that case. would you agree?

The car (and Ovms) only knows it is connected to a charger and has a pilot signal. The scheduling is controlled by the charger hence in my view "powerwait" makes sense until it starts charging then transition to "onrequest", I am assuming "scheduledstart" would indicate charging scheduled by the car or ovms?

drc38:
Also rather than by schedule I can start charging adhoc with a remotestart request to the charger.

mjkapkan:
@drc38 Agreed, that makes sense. @dexterbg I think it would be useful to have a new standard metric called "v.c.initiator" or "v.c.requestor" this could show who has initiated the charging request: server, vehicle, module and so on. But for now will add this as an xnl metric.

drc38:
Another option would be to tag the substates with EV, EVSE, OVMS eg onrequestEV vs onrequestEVSE, stoppedOVMS vs stoppedEVSE

mjkapkan:
Well, that would kind of clutter the metric. Having them in different metric would be more clear.

drc38:
> Well, that would kind of clutter the metric. Having them in different metric would be more clear.

If you add state eg "v.c.state.initiator" then the metric can be for charging or stopping as it could be started by the EVSE and later stopped/suspended by OVMS for example.

mjkapkan:
@drc38 Also not sure if this is related to my change, but I have noticed that the chargeport door is set to false 2 minutes after charging is finished. This in turns leads to not being to start the charge from the app (slider not shown). So have added the logic to not set charge port door to false for MY 2013 and newer cars if the pilot is true, as it correctly shows if a cable is plugged in both for DC and AC.

mjkapkan:
> > Well, that would kind of clutter the metric. Having them in different metric would be more clear.
> 
> If you add state eg "v.c.state.initiator" then the metric can be used for charging or stopping as a session could be started by the EVSE and later stopped/suspended by OVMS for example.

Agreed. Opened new ticked to track this: #677

mjkapkan:
> @drc38 Also not sure if this is related to my change, but I have noticed that the chargeport door is set to false 2 minutes after charging is finished. This in turns leads to not being to start the charge from the app (slider not shown). So have added the logic to not set charge port door to false for MY 2013 and newer cars if the pilot is true, as it correctly shows if a cable is plugged in both for DC and AC.

@dalathegreat you might want to check if this issue exists on MY2012, so it can be handled at the same time ;). Basically after stopping the charge, check back after 2 minutes of the charge port is set to false even when a cable is connected.

mjkapkan:
> The changes look useful. The substate transitions may need some tweaking however. Plugging in to my charger which does not charge until a set time of day no longer transitions out of a powerwait substate.

@drc38 This is fixed now, new firmware uploaded to the same link.

drc38:
> > @drc38 Also not sure if this is related to my change, but I have noticed that the chargeport door is set to false 2 minutes after charging is finished. This in turns leads to not being to start the charge from the app (slider not shown). So have added the logic to not set charge port door to false for MY 2013 and newer cars if the pilot is true, as it correctly shows if a cable is plugged in both for DC and AC.
> 
> @dalathegreat you might want to check if this issue exists on MY2012, so it can be handled at the same time ;). Basically after stopping the charge, check back after 2 minutes of the charge port is set to false even when a cable is connected.

I seem to recall on my 2016 there can be cases where the 0x390 message stops before the charge port is set to false, hence the catch-all after 2min if the substate is not powerwait or timerwait (indicating the cable is still plugged in). I'll let you know how it goes next charge.

mjkapkan:
@dexterbg I see the Range and SOC limit can be set from Android app settings. Can you advise how to update the titles?
Screenshot_20220107-223000

10 is suffsoc
11 is suffrange

drc38:
Using the latest update I seem to be getting a range calculation of 1440min even when the car is not charging, which then shows in the app.
Screenshot_20220108-190859_Open Vehicle

drc38:
> I seem to recall on my 2016 there can be cases where the 0x390 message stops before the charge port is set to false, hence the catch-all after 2min if the substate is not powerwait or timerwait (indicating the cable is still plugged in). I'll let you know how it goes next charge.

Just checked and new lines 1824-1827 need to be removed or amended, otherwise the chargeport is set to false when a charger is plugged in but not yet charging (eg scheduled/delayed start) when it should be true.

dexterbg:
> @dexterbg I see the Range and SOC limit can be set from Android app settings. Can you advise how to update the titles?
> 
> 10 is suffsoc 11 is suffrange

Features besides the V2 standard ones are vehicle specific. The App needs to be extended to label these accordingly, there is no way to define these in the module.

For the Android App, see here:
- https://github.com/openvehicles/Open-Vehicle-Android/blob/master/OpenVehicleApp/src/main/java/com/openvehicles/OVMS/ui/settings/FeaturesFragment.java#L183
- https://github.com/openvehicles/Open-Vehicle-Android/blob/master/OpenVehicleApp/src/main/java/com/openvehicles/OVMS/ui/settings/FeaturesFragment.java#L240

It seems the iOS App does the feature labels here:
- https://github.com/openvehicles/Open-Vehicle-iOS/blob/a81e378e8ad4c585d0deec363743aecdfba4150e/OpenVehicleApp/ovms/ovmsControlFeaturesViewController.m#L229
…but I'm not familiar with the iOS code.

Regards,
Michael

mjkapkan:
> > I seem to recall on my 2016 there can be cases where the 0x390 message stops before the charge port is set to false, hence the catch-all after 2min if the substate is not powerwait or timerwait (indicating the cable is still plugged in). I'll let you know how it goes next charge.
> 
> Just checked and new lines 1824-1827 need to be removed or amended, otherwise the chargeport is set to false when a charger is plugged in, but not yet charging (eg scheduled/delayed start) when it should be true.
> 
> Also when I stop the charge via the charger the substate should change to powerwait as it is still plugged in and charging can be started again. At the moment it is reporting interrupted then changes to stopped when the cable is unplugged.

It looks like, I am unable to replicate this. My vehicle switches to powerwait if the charging timer is set to start at later time. And also if the cable is connected, the pillot is always yes. Maybe this is because the delayed charging is set from the car and not from the charger? (mine doesn't have this feature). Or it could be your model year car behaves differently. Which year do you set in the NL features section?

drc38:
Mine is a 2016, however the pilot signal is controlled by the charger so it is possibly a safety feature whereby the pilot is only sent when the charger is set to charge at a current >0A.

mjkapkan:
> Mine is a 2016, however the pilot signal is controlled by the charger so it is possibly a safety feature whereby the pilot is only sent when the charger is set to charge at a current >0A.

What is the sequence of the lights and sounds when you connect to the charger, but the charger is set not to charge immediateley? Does it do like this?
`beeps once>car HV relay clicks>beeps twice>vehicle HV relay clicks`
Or like this?
`beep once>no clicking of vehicle rellay`
Or something else?

mjkapkan:
@dexterbg Regarding android app, which metric is used to show the slider? Both `v.c.pilot` and `v.d.cp` are 'yes', the `v.c.state` is 'timerwait' and `v.c.substate` is 'powerwait', but the slider is not shown.

drc38:
> > Mine is a 2016, however the pilot signal is controlled by the charger so it is possibly a safety feature whereby the pilot is only sent when the charger is set to charge at a current >0A.
> 
> What is the sequence of the lights and sounds when you connect to the charger, but the charger is set not to charge immediateley? Does it do like this? `beeps once>vehicle HV relay clicks>beeps twice>vehicle HV relay clicks` Or like this? `beeps once>no clicking of vehicle rellay` Or something else?

One beep and no clicks or lights on dash.

dexterbg:
> @dexterbg Regarding android app, which metric is used to show the slider? Both `v.c.pilot` and `v.d.cp` are 'yes', the `v.c.state` is 'timerwait' and `v.c.substate` is 'powerwait', but the slider is not shown.

`v.c.substate` = `powerwait` inhibits the slider if the charge port is open. Substate `powerwait` means "not actually plugged in yet" (waiting for the power cable connection).

See:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_teslaroadster/src/vehicle_teslaroadster.cpp#L373
- https://github.com/openvehicles/Open-Vehicle-Android/blob/master/OpenVehicleApp/src/main/java/com/openvehicles/OVMS/ui/InfoFragment.java#L703

drc38:
> > @dexterbg Regarding android app, which metric is used to show the slider? Both `v.c.pilot` and `v.d.cp` are 'yes', the `v.c.state` is 'timerwait' and `v.c.substate` is 'powerwait', but the slider is not shown.
> 
> `v.c.substate` = `powerwait` inhibits the slider if the charge port is open. Substate `powerwait` means "not actually plugged in yet" (waiting for the power cable connection).
> 
> See:
> 
> * https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_teslaroadster/src/vehicle_teslaroadster.cpp#L373
> * https://github.com/openvehicles/Open-Vehicle-Android/blob/master/OpenVehicleApp/src/main/java/com/openvehicles/OVMS/ui/InfoFragment.java#L703

Powerwait can also work in the context of the external charger controlling charging (even though the cable will be plugged in) as the car cannot initiate a charge only the charger can, hence no reason to show the slider. Note in the app for the NL the charger cable will show greyed out indicating the charge port is open and a cable is plugged in.

mjkapkan:
Thanks, this clears up a lot. It looks like separate method and substate is needed for timed charge. In that case timerwait should do the trick as the substate.

slcasner:
The code comment that @dexterbg referenced is part of code that I recently added for the Tesla Roadster to allow manually starting the charge when the cable is plugged in and the EVSE is ready but the car is waiting for a scheduled start. This Roadster was the car for which OVMS was first implemented so the states and substates were determined by the CAN message values from the car. I needed to infer the scheduled start problem and add the hack to fake the state. For other vehicles the set of states and substates need to be created from other data from the car, so the mapping is more flexible. You may need to similarly infer a particular situation and work around what the car is indicating.

-- Steve

mjkapkan:
> The code comment that @dexterbg referenced is part of code that I recently added for the Tesla Roadster to allow manually starting the charge when the cable is plugged in and the EVSE is ready but the car is waiting for a scheduled start. This Roadster was the car for which OVMS was first implemented so the states and substates were determined by the CAN message values from the car. I needed to infer the scheduled start problem and add the hack to fake the state. For other vehicles the set of states and substates need to be created from other data from the car, so the mapping is more flexible. You may need to similarly infer a particular situation and work around what the car is indicating.
> […](#)
> -- Steve
> On Jan 9, 2022, at 3:01 AM, Michael Balzer ***@***.***> wrote: ﻿ @dexterbg Regarding android app, which metric is used to show the slider? Both v.c.pilot and v.d.cp are 'yes', the v.c.state is 'timerwait' and v.c.substate is 'powerwait', but the slider is not shown. v.c.substate = powerwait inhibits the slider if the charge port is open. Substate powerwait means "not actually plugged in yet" (waiting for the power cable connection). See: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_teslaroadster/src/vehicle_teslaroadster.cpp#L373 https://github.com/openvehicles/Open-Vehicle-Android/blob/master/OpenVehicleApp/src/main/java/com/openvehicles/OVMS/ui/InfoFragment.java#L703 — Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android. You are receiving this because you are subscribed to this thread.

Thanks @slcasner ! Had a hunch it's something like that. I think we have the majority of the states for Leaf, but if we want to add more, we need also to find the logic on how the substates affect the apps.

drc38:
> Thanks, this clears up a lot. It looks like separate method and substate is needed for timed charge. In that case timerwait should do the trick as the substate.

Unless I'm missing something the case in vehicle_nissanleaf_charger_status for CHARGER_STATUS_PLUGGED_IN_TIMER_WAIT should provide the correct states/substates already?

slcasner:
(Sorry, hard to edit replies on the iPhone)

I looked at the iOS app source code to see the effects of the substates. It is a bit tricky because you have to translate into the numeric v2 server protocol values.

mjkapkan:
I will try to update the Android app code in separate [issue](https://github.com/openvehicles/Open-Vehicle-Android/issues/135) on adroid app repository. 
Meanwhile I (hopefully) resolved all the stuff we discussed:

- Charge time estimation: Restored old code and moved that to separate function and just added a way to adjust calculation for different range estimation methods (ideal/est). (Side note: the estimation has never been acccurate for my car, especially now when the battery has degraded quite a bit, so will try to work on better estimation in the future.)
- Powerwait substate: Partially restored old code. Should work as before now, for that specific situation. But new substates will work as well.

Feel free to test. Firmware uploaded to the same place.

drc38:
Just tested and the two issues raised are resolved on my NL 👍

mjkapkan:
Thanks @drc38 . One last thing I did is to set the v.d.cp to true if the pilot is activate, and only then check if it's stale and do the logic you implemented. This way it should work great now for both powerwait (do not show slider) and timerwait (show slider) states.

drc38:
> Thanks @drc38 . One last thing I did is to set the v.d.cp to true if the pilot is activate, and only then check if it's stale and do the logic you implemented. This way it should work great now for both powerwait (do not show slider) and timerwait (show slider) states. This should close #645.

Interesting I wonder why that is necessary as the charge port should already be set to true in the CHARGER_STATUS_PLUGGED_IN_TIMER_WAIT case logic? Is it getting set to false somewhere?

mjkapkan:
> > Thanks @drc38 . One last thing I did is to set the v.d.cp to true if the pilot is activate, and only then check if it's stale and do the logic you implemented. This way it should work great now for both powerwait (do not show slider) and timerwait (show slider) states. This should close #645.
> 
> Interesting I wonder why that is necessary as the charge port should already be set to true in the CHARGER_STATUS_PLUGGED_IN_TIMER_WAIT case logic? Is it getting set to false somewhere?

Yes it does set to true, but if I don't add these lines as in latest commit, somehow here it get's set to false:
```
    std::string str = StandardMetrics.ms_v_charge_substate->AsString();
    if      (str == "timerwait") ;
    else if (str == "powerwait") ;
    else StandardMetrics.ms_v_door_chargeport->SetValue(false);
    }
```
I am not too familiar with this kind of syntax, so will not touch this part in this PR. Most importantly it works now. So as this was not the main part of this PR, I would preffer to not make any more changes here. But later on we can of course refactor this stuff to be more precise.

drc38:
> > > Thanks @drc38 . One last thing I did is to set the v.d.cp to true if the pilot is activate, and only then check if it's stale and do the logic you implemented. This way it should work great now for both powerwait (do not show slider) and timerwait (show slider) states. This should close #645.
> > 
> > 
> > Interesting I wonder why that is necessary as the charge port should already be set to true in the CHARGER_STATUS_PLUGGED_IN_TIMER_WAIT case logic? Is it getting set to false somewhere?
> 
> Yes it does set to true, but if I don't add these lines as in latest commit, somehow here it get's set to false:
> 
> ```
>     std::string str = StandardMetrics.ms_v_charge_substate->AsString();
>     if      (str == "timerwait") ;
>     else if (str == "powerwait") ;
>     else StandardMetrics.ms_v_door_chargeport->SetValue(false);
>     }
> ```
> 
> I am not too familiar with this kind of syntax, so will not touch this part in this PR. Most importantly it works now. So as this was not the main part of this PR, I would preffer to not make any more changes here. But later on we can of course refactor this stuff to be more precise.

It looks like the reason is the removal of line 349:
`StandardMetrics.ms_v_charge_substate->SetValue("timerwait");`

It would be cleaner to reinstate this line rather than include the latest commit.

mjkapkan:
> > > > Thanks @drc38 . One last thing I did is to set the v.d.cp to true if the pilot is activate, and only then check if it's stale and do the logic you implemented. This way it should work great now for both powerwait (do not show slider) and timerwait (show slider) states. This should close #645.
> > > 
> > > 
> > > Interesting I wonder why that is necessary as the charge port should already be set to true in the CHARGER_STATUS_PLUGGED_IN_TIMER_WAIT case logic? Is it getting set to false somewhere?
> > 
> > 
> > Yes it does set to true, but if I don't add these lines as in latest commit, somehow here it get's set to false:
> > ```
> >     std::string str = StandardMetrics.ms_v_charge_substate->AsString();
> >     if      (str == "timerwait") ;
> >     else if (str == "powerwait") ;
> >     else StandardMetrics.ms_v_door_chargeport->SetValue(false);
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am not too familiar with this kind of syntax, so will not touch this part in this PR. Most importantly it works now. So as this was not the main part of this PR, I would preffer to not make any more changes here. But later on we can of course refactor this stuff to be more precise.
> 
> It looks like the reason is the removal of line 349: `StandardMetrics.ms_v_charge_substate->SetValue("timerwait");`
> 
> It would be cleaner to reinstate this line rather than include the latest commit.

Unfortunateley, this would break the logic of setting the scheduled start/stop substates. That's why most of the substate handling has been moved to the "HandleCharging()" function. I only left the setting of the substes here that were tricky to move.

drc38:
> Unfortunateley, this would break the logic of setting the scheduled start/stop substates. That's why most of the substate handling has been moved to the "HandleCharging()" function. I only left the setting of the substes here that were tricky to move.

I couldn't see an if statement that refers to the "timerwait" substate, or have I missed something? Also I noticed line 2035 can never be called as "HandleCharging" is only called when the pilot is set to True, which also means substates such as powerwait and timerwait  cannot be easily set in the function because they occur before or after charging rather than during.

mjkapkan:
> > Unfortunateley, this would break the logic of setting the scheduled start/stop substates. That's why most of the substate handling has been moved to the "HandleCharging()" function. I only left the setting of the substes here that were tricky to move.
> 
> I couldn't see an if statement that refers to the "timerwait" substate, or have I missed something? Also I noticed line 2035 can never be called as "HandleCharging" is only called when the pilot is set to True and ChargingInProgress is True, which also means substates such as powerwait and timerwait cannot be easily set in the function because they occur before or after charging rather than during.

This was like that before my changes. Now "HandleCharging" is called regardless of charging state. The "timerwait" can be both state and substate. But we only use "timerwait" substate if no other state is fit for the description. So when module stops charging due to limit SOC it will have "timerwait" as state and "scheduledstop" as substate. Same way if charging is stopped from app or via shell the substate will be "onrequest". And finally if charging is stopped by car or charger (while pilot still on) the substate will be "timerwait".

drc38:
Ah ok, I missed the change in logic due to the function renaming. It would be good to split out the substate change logic into its own function. At some stage I will look to add the V2X logic in the "HandleCharging" to stop export based on the minimum range or SOC now that charging can be stopped from ovms.

mjkapkan:
This PR is Ready to merge from my side.

dexterbg:
Getting compiler warnings:

```
/home/balzer/esp/Open-Vehicle-Monitoring-System-3/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp: In member function 'void OvmsVehicleNissanLeaf::HandleCharging()':
/home/balzer/esp/Open-Vehicle-Monitoring-System-3/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp:1992:10: warning: unused variable 'charge_power_w' [-Wunused-variable]
   float  charge_power_w     = 0;
          ^
/home/balzer/esp/Open-Vehicle-Monitoring-System-3/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp:2003:10: warning: unused variable 'max_range' [-Wunused-variable]
   float  max_range          = StandardMetrics.ms_v_bat_range_full->AsFloat(0, Kilometers);
          ^
```

mjkapkan:
@dexterbg good catch. Fixed in #681.

PR #682, opened 2022-01-14
Improve wakeup message

dalathegreat:
Noticed while testing on various LEAFs that it doesn't always wake them up. Added another message that helps :)

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -2230,10 +2230,11 @@ void OvmsVehicleNissanLeaf::HandleRange()
 //
 OvmsVehicle::vehicle_command_t OvmsVehicleNissanLeaf::CommandWakeup()
   {
-  // The on board charger wakeup message is the same on all 2011-2017 LEAFs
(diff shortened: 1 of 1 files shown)


mjkapkan:
Yeah previous stuff did not work for me. Will try it out. It might depend on the firmware in the car as well.

dalathegreat:
For future reference, here are all the wakup messages that the LEAFs use. env200 might be different, haven't got enough logs.
x68C CAR-CAN (2013+)
x682 CAR-CAN ? 
x603 CAR-CAN (2011-2012)
x68C EV-CAN (2013+)
x679 EV-CAN (all)
x603 EV-CAN (2011-2012)

drc38:
@dalathegreat what's the logic behind sending x679 twice to the same bus, wouldn't be better to add another message on the other bus eg x68c or same x679 message on the other bus?

mjkapkan:
For me the only way all metrics are updated when vehicle is not charging or plugged in is by sending charge stop message. Will try to check which frames actually trigger it

dalathegreat:
> @dalathegreat what's the logic behind sending x679 twice to the same bus, wouldn't be better to add another message on the other bus eg x68c or same x679 message on the other bus?

The LEAF has some modules that operate only on EV-CAN, and some only on CAR-CAN. It is quite common for the VCM to copy messages from one bus to the other. For example copy ABS info from CAR -> EV.

PR #695, opened 2022-02-03
branding: Fix Bolt/Volt to be consistent in name

kallisti5:


kallisti5:
this one is pretty simple.. it just makes the pretty names of the Volt / Bolt to be consistent.

Yes, VA doesn't match the new standard.  I opted to leave it VA over CVOLT since changing this would break people's configurations.

Issue #344, opened 2020-03-10
GSM Data Saver V3 Server MQTT: post only selected metrics?


glynhudson:
I've been using server V3 to post metrics to an MQTT server for logging using Emoncms, informing OpenEVSE chargepoint of vehicle SoC and logging GPS data to OwnTracks. It works great, I've been running this for flawlessly for the past year or so. 

Currently, when server V3 is enabled all metrics are periodically posted. This adds a *lot* of extra data to the GSM bandwidth. e.g Using hologam my monthly data cost when from $1.9/month ($23/yr) when running only V2 to $17/month ($204/yr) when enabling V3 server. 

I'm aware it's possible to reduce the rate the V3 data is posted, however this would have a detrimental effect on the metrics I'm logging. Ideally what I would like is just to be able to select which metrics are posted e.g just SoC, and GPS co-ordinates. This would significantly reduce the amount of data required. 

Is this possible currently? 

I've been working with @trystanlea, the main developer behind https://emoncms.org to develop a MQTT integration with OVMS to enable users to easily log OVMS metrics to emoncms. We're keen to avoid a huge amount of metrics being unnecessarily posted to the server.

TrystanLea:
If providing per-metric selection is too complicated, I imagine it could be.. perhaps having a reduced metrics option that would contain a collection of metrics would be another approach e.g:

- metric/v/b/soc
- metric/v/b/soh
- metric/v/p/altitude
- metric/v/p/latitude
- metric/v/p/longitude
- metric/v/p/odometer
- metric/v/b/12v_voltage

leres:
My credit card statement tells me I should be interested in this (it's not outrageous but 3-4 times higher than it used to be).

The above list of metrics look good; two more I care about are m/net/mdm/sq (cellular signal strength) and v/p/satcount (number of gps satellites in view).

Can we do something simple such as a web console page with check boxes for all possible metrics and a default subset enabled by default?

markwj:
I think it should be simple to have an optional config to list the metrics that should be sent. If not defined, it will send all. Is that sufficient for this requirement? If so, I am happy to do it.

TrystanLea:
Thanks @markwj that would be perfect and really appreciated if possible.

glynhudson:
[...]

PR #717, opened 2022-05-27
Fix/Workaround OVMS Server v3 streaming bug

CrashOverride2:
If using OVMS Server v3 (MQTT) and location streaming is enabled, every second metrics like gps time and monotonic are pushed to the MQTT server. So the OVMS module has no time for housekeeping an other stuff. If network connection changes, it will crash on network modules. With this workaround the location streaming times will be used to push only updates with selected update time/rate under location.


dexterbg:
Carsten, welcome & thanks for the submission.

Implementing a better streaming logic for the V3 server has been on the list for some time, see issue #344.

So you're very welcome to take care of this. Your current attempt seems to be based on a partial understanding of the issue though and is following a wrong / bad approach.

You extend the metric modification callback in a way it's not meant to be used: the metric modification callback is called for a single metric at the very moment an update takes place, in the context of the task doing the update (which is mostly the vehicle task), blocking that task until the callback returns. It's meant to attach single, atomic & small actions that need to be done ASAP following a metric update, e.g. to derive coupled metrics or signals. It may be used to trigger a single asynchronous transmission, but that's stretching the callback execution time about as far as possible.

You then walk through the metrics list starting at the updated metric -- but that isn't necessarily the first of the metrics to be transmitted. If other metrics residing further to the front of the list still need transmission, they will be blocked afterwards, as you update `m_lasttx_stream` after the walkthrough.

A proper implementation of flexible streaming needs to be done in the general ticker handler. There is already a place reserved for this in the code:

https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/05b69a8a99e4ccbff05c509c3b73b06ad95a35ad/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp#L935-L939

Regarding the actual issue, streaming currently enables _immediate_ transmissions of _all_ metric modifications to the V3 server -- not just per second and just location related. To mimic the V2 semantics of the streaming config, you need to implement a filtering version of `TransmitModifiedMetrics()`, as discussed in issue #344, and call that from the ticker shown above.

You can then remove the single metric transmission, possibly eliminating the overall need for the metric modification handler in the V3 server. Another option would be to keep that code, but introduce a separate dedicated config param for it, as it may have its applications.

Regards,
Michael

CrashOverride2:
Hi Michael,

thanks for your response. I am not a very experienced C programmer and complexity of your code is for me high. You are correct I not yet fully understand all functions in the ovms server v3 component.

I know this is an inefficient workaround, I hoped someone could help me to find a good solution for this.

I have no problems if all changed metrics are transmitted while streaming, the main problem with original state of the code, the timer set under notifications is completely ignored. I can filter metrics like in v2 or sent all changed metrics within location streaming time, because MQTT is fast and data efficient enough.

If a time is set it started streaming metrics permanently. Of course TransmitModifiedMetrics() is called multiple times per second, but the visible result is transmitting every second because of changing metrics like monotonic or gpstime every second in idle.

If I understand you correctly, immediate transmissions of metrics regardless of filtering or not is by design. Like server v2 if a peer is connected send filtered metrics asap.

Okay I will look and try to understand server3 code better, but what is better transmit all or filtered metrics within location streaming time or transmit filtered metrics like position and so on immediately?

dexterbg:
> I have no problems if all changed metrics are transmitted while streaming, the main problem with original state of the code, the timer set under notifications is completely ignored. I can filter metrics like in v2 or sent all changed metrics within location streaming time, because MQTT is fast and data efficient enough.

To get that behaviour on v3, don't enable location streaming, but set the "on" interval instead. That will cause the current v3 code to call `TransmitModifiedMetrics()` in that interval while the vehicle is on.

> Of course TransmitModifiedMetrics() is called multiple times per second, but the visible result is transmitting every second because of changing metrics like monotonic or gpstime every second in idle.

No, `TransmitModifiedMetrics()` is called at most once per second, as it's done from the Ticker1 handler. It will probably be more efficient than doing single transmits on metric updates, as it runs in the event task context and can create a large chunk of data for the network layer.

> If I understand you correctly, immediate transmissions of metrics regardless of filtering or not is by design. Like server v2 if a peer is connected send filtered metrics asap.

You need to ask @markwj about the MQTT design details. I think it's a re-interpretation of the early v2 streaming mode aimed at making use of the advanced capabilities of the v3 hardware. The v2 streaming mode originally didn't support intervals, just on/off, intervals have been added later on.

> what is better transmit all or filtered metrics within location streaming time or transmit filtered metrics like position and so on immediately?

See above, I think it makes sense to make streaming mode work similar to v2 on v3, as we only have one config parameter for this, but to keep the current "high speed" version as well.

Regards,
Michael

CrashOverride2:
Hi Michael,

can you review my latest changes. I try to implement streaming like v2 server.

I know I need to remove logging, I just want your opinion.

I tested it for 2 weeks and its working.

Thanks
Carsten

dexterbg:
Carsten,

I think you should have someone comment on this who is actually using the MQTT client. I am not, but I would probably have an issue with a fixed set of metrics to be defined as being important. Part of what makes MQTT attractive is the fine granularity of update control that can be implemented as each metric is transmitted individually.

Technically, your ticker addition is very inefficient by transmitting a whole chunk of metrics if any of them has changed. That's the downside of MQTT, it's got a lot of overhead due to single metric transmissions. So we only should transmit metrics that need to be transmitted.

That would lead to the same pattern as you turned `TransmitModifiedMetrics()` into (i.e. checking a set of defined metrics), so both could simply apply the same method to a different set of metrics.

That in turn could also lead to a simple means to let the user have some control over these transmission groups: use maps or lists to define the groups, and add some config scheme to support extending these by custom entries.

Please describe your solution in issue #344 and get some feedback from other MQTT users.

Regards,
Michael

CrashOverride2:
Hi Michael,

thanks for your reply, I understand it isnt efficient of course. In server v2 this is similar, if one metric changes a chunk will be submitted. It is not so heaviweight than mqtt ok.

But this is better than original code, because if you have streaming enabled it will publish every second till it crashes. Streaming will be enabled by users who use v2 and v3 simultaneously I think.

Its not the best solution there are other people who can implement individual topics and so on like in #344.

But this is too complicated for me, so I am giving up on this pull request, you can close it.

PR #718, opened 2022-07-03
Update the docs for the ABRP plugin to refer to the official repo

dteirney:
The current content was already a few versions out-of-date and a new 2.0 version has been released that incorporates all of the installation and usage instructions in the README


dteirney:
There was some duplicate content for the documentation and plugin source code. I've deleted most of the information assuming that referring to the external repository that holds the latest content is the most sensible place for people to refer to going foward.

dexterbg:
David, thanks for contributing.

The "plugins" directory structure is meant to serve as the main plugin repository for the plugin management feature (WIP). While it will of course be possible to use other repositories, from a usability point of view we should keep at least the meta data in one place, so users can browse through all plugins available without further configuration.

Currently the metadata syntax isn't defined for external sources though (IIRC). You're welcome to work on this. Until we've got that usable, I suggest keeping the plugin in the central repository.

In theory using a git submodule to include the iternio repository would be an option. I have no idea about the download URLs provided by github for such a setup, but you could sort that out.

Read about the plugin repository draft in this thread:
http://lists.openvehicles.com/pipermail/ovmsdev/2020-July/006942.html

Regards,
Michael

werdnum:
Is there a way we can move this PR forward? I encountered this issue in December, and it seems like a year is too long to leave the plugin in this broken state. Should we just remove the old version from the repository to avoid confusion?

dexterbg:
Well, the PR is unchanged and still broken. But I'll merge it now, as probably more harm is done by the outdated info.

PR #723, opened 2022-08-01
Improve time to charge to Range Limit

GreyPeter:
Adjust time to charge to Range Limit to account for battery SOH


GreyPeter:
Calculated time to charge to the Range Limit was inaccurate due to the battery state-of-health not taken into account. This patch fixes that.

PR #727, opened 2022-09-21
Fix the TCP client CAN logger

vincentbarat:
The implementation of the TCP client CAN logger is broken.
canlog_tcpclient->Open() does not wait for connection to be established and immediatly returns, making canlog_tcpclient->IsOpen() always return false, and thus the logger object to be deleted immediately.
I propose a patch that waits at most 10s until the connection is established.

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/canlog_tcpclient.cpp
+++ b/vehicle/OVMS.V3/components/can/src/canlog_tcpclient.cpp
@@ -34,7 +34,6 @@ static const char *TAG = "canlog-tcpclient";
 #include "canformat.h"
 #include "canlog_tcpclient.h"
 #include "ovms_config.h"
-#include "ovms_peripherals.h"
 
 canlog_tcpclient* MyCanLogTcpClient = NULL;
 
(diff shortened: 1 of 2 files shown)


dexterbg:
Vincent, welcome & thanks for contributing.

Nice find, pointing to a hole in the logger API definition: `Open()` isn't required to be synchronous but the API also offers no means to wait for a connection.

The purpose of `IsOpen()` is to tell the framework if the logger instance is currently accepting messages for logging. Both TCP loggers are technically "open" as soon as they have been instantiated (they'll just discard log messages while not connected). But the TCP client doesn't retry connecting, so would need a manual "stop" command if it cannot connect to the destination port on the first attempt.

Your fix has some issues though:

1. It's got a race condition: on a context switch between `m_connecting = false` and `m_isopen = true`, it will assume a failed connection. Rule of thumb: reach a consistent final state before signaling any "go" to other tasks.
3. A `usleep()` loop with 10 ms is a bit better than busy looping, but still really isn't the FreeRTOS way of doing things like this. Consider using semaphores instead wherever feasible -- scheme below.
4. If you need to use state variables instead of semaphores (not here), you should always make them `volatile` -- don't rely on the compiler to detect a busy wait on a variable.
5. Minor: a fixed maximum blocking time also isn't perfect, but would be acceptable for now.

Semaphore scheme:

1. Replace `bool m_connecting` by `OvmsSemaphore m_connecting` (add the `#include "ovms_semaphore.h"` as necessary).
2. In `Open()`, replace the busy wait loop by a simple `m_connecting.Take(pdMS_TO_TICKS(10*1000))`.
3. In the handler, add `m_connecting.Give()` after (!) each `m_isopen` state change.

The task doing the `Take` operation will be put to sleep by FreeRTOS until the token is given to the semaphore or the timeout is reached.

Regards,
Michael

vincentbarat:
Hi Michael,
Thanks for your judicious comments :)
I fixed my pull request accordingly.
Regards,

Issue #728, opened 2022-09-30
[components/can] 2 small questions regarding naming


llange:
While browsing the source code I have 2 questions - not important things, but as I start to dig I try to understand the way the things are.

* The TAG for [the canformat `canswitch`](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/b70e4e3a8d46245f7564bc5f5479a97739cb1a90/vehicle/OVMS.V3/components/can/src/canformat_canswitch.cpp#L28) is `canformat-crtd` (i.e., the same as the canformat `crtd`), is this expected or a typo ? I can provide a PR to change it (to `canswitch` ? to `cs11` ?)

* Why is the [canformat `lawricel`] named as such ? It looks like the format from [Sweden Company Lawicel AB](http://www.lawicel.com/can232/can232.pdf) (not law**R**icel), is it expected or a typo ?)

Thanks for your help.

(PS: Please tell me also if it's the right place to ask, or would be better in the forum ? in the mailing-list ? somewhere else ?)

dexterbg:
You're probaly right on both, and you'll find some more copy/paste remnants, e.g. in the comments, as CRTD was the first formatter. You're welcome to post a pull request.

llange:
Thanks @dexterbg .
I've open 2 PRs:
* #729 for `cs11` (canswitch)
* #730 (Draft, not ready to be merged) form `lawricel` -> `lawicel`. Being a little bit concerned that it could break existing setup and user habits, I've duplicated the `RegisterCanFormat` call so that the format is registered twice and available as both `lawicel` and `lawricel`. I don't know if it's a good idea or not, neither if it will work... - let me know.

llange:
Thank you !

PR #730, opened 2022-09-30
rename `lawricel` format to `lawicel` (#728)

llange:
Typo - this format name (`lawicel`, a.k.a. `slcan` in the Linux Kernel) comes from the LAWICEL AB company and its product : http://www.lawicel.com/can232/can232.pdf

### DIFF
--- a/docs/source/crtd/can_logging.rst
+++ b/docs/source/crtd/can_logging.rst
@@ -77,7 +77,7 @@ or log specific CAN packets by applying a filter e.g 0x55b the Nissan LEAF SoC C
 
 ``ovms# can log start vfs crtd /sd/can.crtd 55b``
   
-Other CAN log file formats are supported e.g ``crtd, cs11, gvret-a, gvret-b, lawricel, pcap, raw``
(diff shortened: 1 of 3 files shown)


dexterbg:
You missed changing the names of the classes, methods, etc. -- the typo should be fixed everywhere, not only in strings.

llange:
Yes, it's Work In Progress - I was being a little bit concerned that it could break existing setup and user habits, I've duplicated the RegisterCanFormat call so that the format is registered twice and available as both lawicel and lawricel. I don't know if it's a good idea or not, neither if it will work... - let me know.

dexterbg:
Missed your comment -- please always post related comments with the PR:

> Being a little bit concerned that it could break existing setup and user habits, I've duplicated the RegisterCanFormat call so that the format is registered twice and available as both lawicel and lawricel. I don't know if it's a good idea or not, neither if it will work... - let me know.

Generally good to consider compatibility, but if you register both, the command shortcut "law" will no longer be unique, so you'll break habits that way just as likely.

My 2 cents: if there were user habits to consider here, users would have suggested fixing the typo already. Usability (i.e. unique "law" shortcut) has priority.

llange:
Ready to merge AFAICT - please review when you have time. Thx.

PR #734, opened 2022-10-08
2nd gen Cadillac CTS: Leverage spontaneous PIDs to avoid polling when the car is off.

leres:
We only use two poll states, remove unused polltimes from the poll_pid_t list. Minor whitespace.


leres:
Oops, I'm not setting the poll state...

PR #736, opened 2022-10-11
fix negative integer DBC numbers having wrong values

llange:
When non-32-bits negative numbers were processed by the DBC processor, there was an issue where the resulting metric values were incorrect.

As an example:

A 16-bit signed value coming from the can bus as 0xFFF7(-9) is being multiplied by a factor of -1.
When handled by the DBC processor, the incoming number is stored as 0x0000FFF7 ; which is not -9, but 65527.
After multiplication, it comes as 0xFFFF0009(-65527).

(While those should be: 0xFFFFFFF7(-9) and 0x00000009(9))

The root cause is that the sign of the non-32-bit number is not extended to the left, thus making it a different number.

The fix is to sign-extend the number to the left (it should work from 2-bit numbers to 32-bit numbers) when we're dealing with signed numbers.

### DIFF
--- a/vehicle/OVMS.V3/components/dbc/src/dbc.cpp
+++ b/vehicle/OVMS.V3/components/dbc/src/dbc.cpp
@@ -830,8 +830,10 @@ dbcNumber dbcSignal::Decode(CAN_frame_t* msg)
 
   if (m_value_type == DBC_VALUETYPE_UNSIGNED)
     result.Cast((uint32_t)val, DBC_NUMBER_INTEGER_UNSIGNED);
-  else
-    result.Cast((uint32_t)val, DBC_NUMBER_INTEGER_SIGNED);
+  else {
+    int32_t signed_val = sign_extend<uint32_t, int32_t>((uint32_t)val, m_signal_size-1);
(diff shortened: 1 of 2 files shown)


dexterbg (approved):
LGTM

llange:
If anybody is using the DBC component / vehicle, I'd appreciate some feedback on this PR, to ensure there is no regression. Thank you !

dexterbg:
Best to cross post requests like these to the mailing list, as there are many developers not following github notifications.

frogonwheels (approved):
This looks good.  Somebody may have a suggestion as to a .h file that we could put the sign_extend function in, however it is only a compile time issue.

frogonwheels:
>  Somebody may have a suggestion as to a .h file that we could put the sign_extend function in, however it is only a compile time issue.
Having said that, I now know this function is here, and I can have a discussion about where to put this separately and pull it out into a common header at that point! This is a bug so let's get it merged in.

dexterbg:
Please move the general sign extension utility to the utility module (`main/ovms_utils.h`).

PR #744, opened 2022-10-21
Single poll to vector buffer

frogonwheels:
Add 'vector buffer' implementation for single polls.
Also adds some tools to safely extract various data from the vector buffers.


frogonwheels:
This is no longer needed and is too big a change for what it gives.

PR #746, opened 2022-10-22
add more information on the wifi scanning mode - and CLI instructions

llange:
As I found myself not understanding the scanning mode of the wifi interface, I thought it could be interesting to add more details in the doc.


dexterbg:
This is a bit unbalanced and harder to read than before now. The initial section was meant to and should first outline the modes for the user, not go into details. So command details should come after the initial outline section. Also it should first point to the web UI, possibly with a screenshot or two, as normal users will prefer that.

When explaining command details, you should not omit the commands & configs for the access point mode. The section then also needs to explain auto start configuration using the shell, it currently misleads to thinking a manual mode setup would be persistent.

llange:
Hi @dexterbg, here is a new proposal.
Let me know if you prefer this direction ; in which case I'll write the "access point mode" page.
(I converted the issue to draft mode to mark that it is a work in progress)

llange:
hi @dexterbg, just a quick check if you had the time to look at this proposal. Thx.

dexterbg:
Yes, that looks good. I thought it would all still fit on a single page, but it's perfectly ok to introduce separate pages.

llange:
Hi, here is the full update with the 2 wifi modes ready for review (as separate pages).  
If you think they can fit on a page I'll try to do my best to integrate those. I was a little bit concerned that it would be difficult to navigate (especially with a total of 6 screenshots). Let me know if you still prefer it that way.

PR #753, opened 2022-10-24
Vehicle - Add ability to cap the cell count used in Voltage/Temperature

frogonwheels:
This is used when the final cell count depends on readings.
 The idea is you set the maximum readable initially, and then you can cap the count depending on the number of cells read without resetting the data.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.h
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.h
@@ -647,9 +647,11 @@ class OvmsVehicle : public InternalRamAllocated
     std::vector<bool> m_bms_bitset_t;         // BMS tracking: true if corresponding temperature set
     int m_bms_bitset_cv;                      // BMS tracking: count of unique voltage values set
     int m_bms_bitset_ct;                      // BMS tracking: count of unique temperature values set
-    int m_bms_readings_v;                     // Number of BMS voltage readings expected
+    int m_bms_readings_v;                     // Number of BMS voltage readings allowed
+    int m_bms_readings_cap_v;                 // Number of BMS voltage readings expected
     int m_bms_readingspermodule_v;            // Number of BMS voltage readings per module
-    int m_bms_readings_t;                     // Number of BMS temperature readings expected
+    int m_bms_readings_t;                     // Number of BMS temperature readings allowed
+    int m_bms_readings_cap_t;                 // Number of BMS temperature readings expected
     int m_bms_readingspermodule_t;            // Number of BMS temperature readings per module
     float m_bms_limit_tmin;                   // Minimum temperature limit (for sanity checking)
     float m_bms_limit_tmax;                   // Maximum temperature limit (for sanity checking)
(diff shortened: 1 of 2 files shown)


dexterbg:
Michael,

I still have major doubts about this. You're proposing we should have metrics carrying partially filled vectors, which up to now have been meant to use their size as the only info necessary to correctly interpret and process the content. So you're making things more complicated for applications using these metrics, as these then need to take into account the vectors might have been filled partially by intention. We would need to add accompanying "size used" metrics, and all applications would need to read and apply these instead of the intrinsic vector size.

Statistical metrics on the cells will also differ in reliability and applicability, possibly affecting threshold levels and conditions, although the vectors they are assumed to describe appear to be the same.

This all could be acceptable (with some further changes) if it was necessary or beneficial, but I don't see this. The BMS API was designed to allow dynamic reconfiguration of the pack layout, I don't see any need to support partially filled cell data.

Examples for dynamic pack layouts:
- by configuration: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_vweup/src/vehicle_vweup.cpp#L284
- by live CAN detection: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_renaulttwizy/src/rt_can.cpp#L718

Maybe you can elaborate a bit more on exactly why and where you would like to apply this.

Thanks,
Michael

frogonwheels:
Ok - I will have one last go at justifying it, and then I will fall back on a different solution.

The issue I was having with the dynamic BMS buffer change is that it resets all the values collected.. and this change simply allows you to set the value count at the end without clearing the data. (And I didn't have the count until I got to the end of receiving the voltage data over multiple addresses).

The first idea did allow the data to have gaps.. but this current version no longer supports that (yes, that was definitely overkill),  - it just means you can shorten the number of packs at the end.. That's why that finalise factored out code is private not protected.

What it does allow me to do is to specify the maximum packs that a particular car type will hold (so I5 and all battery variants).. and then when all the batter data is polled for, we get to the end and use the bitmasks in the OBD data to find out what the final pack is and set the actual length of the array at that point. If the packs are deemed to be full at that point, it will do the averaging and if not, it won't. 

Yes, in reality, for the car, the number of packs won't change, so all the statistics and stuff are fine, and apply to the smaller number. The array length  for the temps and voltages in the metric won't  be changing (they will be the smaller number).  It's just the internal BMS arrays that will be initialised to the larger number and won't change.

Ok, so the alternate solution (without this change)
 * Start with the largest number of values supported by the potential data available.
 * Fill the packs as it does now.
 * When we get to the end of the series of OBD poll addresses, and we discover a different number of temps/voltages then update the sizes at that point (which would clear the data ).
 * The next time around we get the voltage/temp data and the sizes are the correct size, then it will finalise properly.

So really the only issue is that after a reset, the voltage/temp count will be wrong and the first set of values will be thrown away.  This could be, I guess, mitigated by having a metric or 2 for the i5 that contains the number of cells counted. Then really it would come down to throwing away one set of data only.

dexterbg:
Skipping the first cell data read until you know the layout is one option.

Another could be a detached metrics loading, i.e. read into an internal buffer before publishing to the system. That way you've got full control, don't need a persistent metric, and can even implement partial scans if that makes sense for you.

frogonwheels:
> Another could be a detached metrics loading, i.e. read into an internal buffer before publishing to the system. That way you've got full control, don't need a persistent metric, and can even implement partial scans if that makes sense for you.

..So you mean effectively replicating all the BMS 'filled bits' and array storage etc in order to achieve what this (now) simple modification to the existing system does?

I guess I could make a function which copies out the values from the existing BMS to a temporary array (if they are filled to the point) and then uses the existing resize and then fills all the values back in again.  I mean that could still eventually go to vehicle.h but wouldn't necessarily need to start there.

dexterbg:
> ..So you mean effectively replicating all the BMS 'filled bits' and array storage etc in order to achieve what this (now) simple modification to the existing system does?

Yes, because the modification breaks the simplicity of the current API & model for readers, and the reason for the modification appears to be a vehicle specific issue that's better solved by the vehicle adapter.

> make a function which copies out the values from the existing BMS to a temporary array (if they are filled to the point) and then uses the existing resize and then fills all the values back in again

That's another option. I wouldn't have any objections to adding this to the framework. Could also be done as a parameter to the `BmsSetCellArrangement…` methods.

Maybe you also should reconsider introducing a user configuration for the battery pack or car model variant. If the models have substantial differences, you'll probably need more case differentiations later on. It may also be possible to determine the model from the VIN or some ECU version.

frogonwheels:
Yeah Just implemented the copy -aside and reapply, and it looks fine, is a lot simpler..  and it's a one off extra memory usage.  I'm pretty sure it will work too.

PR #754, opened 2022-10-24
Vehicle - Add buffer handling tools to ovms utils

frogonwheels:
One of the strengths of these buffer extraction tools is that they check to make sure the extractions are within bounds.  They also make very clear how many bytes are being extracted - something I had found confusing previously.
There are big-endian and little-endian versions, signed (with sign-extension thanks to Ludovic's discovering of the bug) and unsigned versions,  std::string  and uint8_t* versions.
 
There is also a slight improvement to the sign_extend (which I originally converted from macro to template and was credited for by Ludovic). I have added an overloaded sign_extend  that templates the sign-bit, shifting the bit arithmetic to compile-time.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_utils.cpp
+++ b/vehicle/OVMS.V3/main/ovms_utils.cpp
@@ -612,3 +612,19 @@ std::string idtag(const char* tag, void* instance)
   std::string res = buf.str();
   return res;
   }
+
+/**
+ * get_buff_string: Helper function to get at string value from sized buffer.
+ */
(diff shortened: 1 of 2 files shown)


frogonwheels:
I have fixed a missing sign-extension on one of the calls.
  
Also, I have added utils to extract signed/unsigned integers from defined bits - with compile-time (Templated) and run-time specified versions.

The integer extractions tools are now templated (with 32bit default) to the integer type.

This means you could use  64 bit integers if you wanted.

PR #757, opened 2022-10-30
SmartED: CommandStats string fix

Dimitrie78:


dexterbg:
Sorry, that one slipped my attention. I've done another review of PR #731. I think this was the only bug, but you're welcome to take a second look.

Dimitrie78:
I think that was the only error I found. I also checked everything three times.

PR #759, opened 2022-11-01
Add Method to change voltage cell arrangement, preserving existing data

frogonwheels:
This adds a method that copies out existing data - resets to a new configuration, and then copies it back in.


dexterbg:
My review results:
- calling `BmsSetCellTemperature` instead of `BmsSetCellVoltage`
- please explain the bool result in the header comment
- please avoid "this seems…" in comments, be sure & clear about the implications
- please add the corresponding temperature layout method

Regards,
Michael

PR #760, opened 2022-11-01
Enhance BMS shell display

frogonwheels:
- Allows for voltages and/or temperatures columns to be displayed depending on what's available.
- Wraps values for one module over multiple lines (where you have multiple cells per module) making it readable in a smaller display.
-  Have a maximum of 5 columns total
- Add commands  `bms temp` to display just temperature stats and `bms volt` to display just voltage stats.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -2,6 +2,10 @@ Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
 - Renault Zoe Phase 2: Initial support
+- Improved output of bms shell command for narrow windows.
+  New commands:
+    bms volt                            -- Output only voltage info if available
+    bms temp                            -- Output only temperature info if available
 
 2022-09-01 MWJ  3.3.003  OTA release
 - Toyota RAV4 EV: Initial support added. Only the Tesla bus is decoded and just listening so far.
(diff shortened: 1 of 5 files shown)


dexterbg:
Review:
- Minor code style remark: we try to keep `verbosity` & `writer` arguments to shell methods together, as they build their base signature and belong together. For `BmsStatus()` that means the additional arguments would come after `writer`, as with `FormatBmsAlerts()`.
- The local utility `check_max_cols()` should be `static` to avoid later linker issues. (Btw, that could also be a lambda, as only `BmsStatus()` needs it, but a static function is perfectly ok.)
- If commands are actually mode parameters of a base command, it's better to use the same function for all and query the command name within. See `vehicle_charge_mode()` for an example. That way it's immediately clear from the command registry these are just modes, not independent commands.
- If changes are made that are relevant for users, a brief note needs to be added to the `changes.txt` history. See there for examples on new command announcements, add the note in the open "???" section for the coming release.

Regards,
Michael

frogonwheels:
Thanks Michael,
My knowledge of c++ was quite good at the time but I'm still catching up on
lambdas; I've been using primarily Delphi for about 18 years now. (In
delphi 'check-max-cols' would be a function local to the function).  We'll
go for a lambda.

I'll switch those two bools to an enum for readability and move it to the
end of the function.  I did check for that virtual function being
overridden and didn't find any.

//.ichael

PR #763, opened 2022-11-02
dbc : fix a regression on sign extension

llange:
Fix a small regression accidentely introduced in ecec210


frogonwheels (approved):
Ooh. I'm so sorry about that.

PR #764, opened 2022-11-05
Metrics - Improve accuracy/consistency km <--> mi and other length conversions

frogonwheels:
- Use functions for km <--> mi conversions for int and float conversions that have better precision.
 - Adds some extra conversions between all length metrics  (feet, metres, miles, km)


frogonwheels:
This branch has errors that are being addressed.

PR #765, opened 2022-11-06
Add commonly used power consumption units to Metrics

frogonwheels:
- Adds kWh/100km, km/kWh, and miles/kwh to metrics.. and conversions between them all.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_metrics.cpp
+++ b/vehicle/OVMS.V3/main/ovms_metrics.cpp
@@ -1577,8 +1577,8 @@ int UnitConvert(metric_unit_t from, metric_unit_t to, int value)
         {
         case WattHoursPM: return pkm_to_pmi(value);
         case kWhP100K: return value / 10;
-        case kPkWh:    return value?static_cast<int>(1000.0 / value):0;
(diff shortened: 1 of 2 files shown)


dexterbg:
Michael,

- for consistency with the `WattHoursP…` and `kWh` schemes, please rename the new unit codes as follows:
  - `kWHP100K` → `kWhP100K`
  - `kPkWH` → `KPkWh`
  - `MPkWH` → `MPkWh`
- for code style & readability, please add spaces around ternary operators `?` and `:`
- the `WattHoursPM` to `kWhP100K` conversions use `km_to_mi`, should be `mi_to_km`

Thanks,
Michael

frogonwheels:
Yes to everything except that  WattHoursPM to kWHP100K  is functionally correct, except that I actually created an alias for km_to_mi of pmi_to_pkm  and vica versa for readability, and I missed putting it here.  (The conversion from  per km to per miles is the same as from miles to km).  So for clarity I will definitely use the pmi_to_pkm and pkm_to_pmi functions!!.  It was doing my head in as well which is why I created the alias.

dexterbg:
Ah yes, you're right.

Just one thing missing: the second rename is incomplete, note the upper case 'K' for 'Kilometers'. That's for consistency with '…PK' and 'Kph…'.

PR #769, opened 2022-11-11
add a recursive ls `vfs rls` command

llange:
When you have a deep directory structure it can sometime be useful to have a `tree`-like command to show all files / directories at once.

This implementation tries to do that, and adds the `vfs rls` command for that:
```
OVMS# vfs rls /store/
[DIR]     01-Jan-1980 00:00  /store/ovms_config/
Error: protected path
[DIR]     01-Jan-1980 00:20  /store/plugin/
    9.8k  01-Jan-1980 00:20  /store/plugin/metrics
    4.0k  01-Jan-1980 01:22  /store/plugin/metrics-table
    5.8k  01-Jan-1980 01:23  /store/plugin/commands
   29.9k  22-Oct-2022 23:13  /store/plugin/dashboard
[DIR]     01-Jan-1980 01:00  /store/dbc/
[DIR]     08-Nov-2022 23:35  /store/scripts/
[DIR]     08-Nov-2022 23:38  /store/events/
[DIR]     08-Nov-2022 23:40  /store/events/sd.unmounting/
     27   09-Nov-2022 21:47  /store/events/sd.unmounting/10-canlog
[DIR]     08-Nov-2022 23:44  /store/events/sd.mounted/
     82   09-Nov-2022 21:45  /store/events/sd.mounted/10-canlog
```

At first I used a classic, recursive approach but as expected it had a huge impact on the stack and crashed often.

I took the opportunity and challenge to write a non-recursive approach - which unfortunately is not that easy to understand.

Looking forward to your comments in order to simplify it / make it more readable.

Note: github has a hard time making the diff readable. It's easier to review it commit by commit.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_vfs.cpp
+++ b/vehicle/OVMS.V3/main/ovms_vfs.cpp
@@ -51,65 +51,6 @@ static const char *TAG = "vfs";
 #include "vfsedit.h"
 #endif // #ifdef CONFIG_OVMS_COMP_EDITOR
 
-void vfs_ls(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
-  {
-  DIR *dir;
-  struct dirent *dp;
-  char size[64], mod[64], path[PATH_MAX];
-  struct stat st;
-
-  if (argc == 0)
-    {
-    if ((dir = opendir (".")) == NULL)
-      {
-      writer->puts("Error: VFS cannot open directory listing");
-      return;
-      }
-    }
-  else
-    {
-    if (MyConfig.ProtectedPath(argv[0]))
-      {
-      writer->puts("Error: protected path");
-      return;
-      }
-    if ((dir = opendir (argv[0])) == NULL)
-      {
-      writer->puts("Error: VFS cannot open directory listing for that directory");
-      return;
-      }
-    }
-
-  while ((dp = readdir (dir)) != NULL)
-    {
-    snprintf(path, sizeof(path), "%s/%s", (argc==0) ? "." : argv[0], dp->d_name);
-    stat(path, &st);
-
-    int64_t fsize = st.st_size;
-    int is_dir = S_ISDIR(st.st_mode);
-    const char *slash = is_dir ? "/" : "";
-
-    if (is_dir) {
-      strcpy(size, "[DIR]   ");
-    } else {
-      if (fsize < 1024) {
-        snprintf(size, sizeof(size), "%d ", (int) fsize);
-      } else if (fsize < 0x100000) {
-        snprintf(size, sizeof(size), "%.1fk", (double) fsize / 1024.0);
-      } else if (fsize < 0x40000000) {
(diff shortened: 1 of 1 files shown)


dexterbg:
Ludovic,

a `find` command equivalent like this is a valuable addition, and I like you've taken it as a challenge to do this without recursion.

But your approach is quite complex and makes adding another important missing feature – sorting entries by name – very challenging. And… I really don't want to let you know this after all your effort, but I don't think it is actually necessary to avoid recursion here in the first place.

I assume the high stack requirement you saw was coming from the local fixed size char buffers, especially `path[PATH_MAX]` with `PATH_MAX` being 1024. Replacing these by dynamic char buffers or simply `std::string` (or a `StringWriter` to have printf style formatting) would eliminate most of the stack requirements.

Sorting entries btw can be done using a `std::set`, see `OvmsScripts::AllScripts()` for this pattern.

If you're still willing to continue on the command: the protected path error shouldn't come in a recursive listing, only if you try to access a protected path initially. Within a listing, I'd rather add a tag on the directory line.

Proposal: tag protected paths with a `[P]`, omit file size on protected files:
```
OVMS# vfs rls /store/
[P][DIR]  01-Jan-1980 00:00  /store/ovms_config/
[P]       01-Jan-1980 00:20  /store/ovms_config_file
[DIR]     01-Jan-1980 00:20  /store/plugin/
    9.8k  01-Jan-1980 00:20  /store/plugin/metrics
    4.0k  01-Jan-1980 01:22  /store/plugin/metrics-table
…
```

We don't have any protected files yet, but may have some in the future. Placing the `[P]` in front of `[DIR]` automatically makes the path inaccessable for readers looking for a `[DIR]` prefix.

What do you think?

Regards,
Michael

llange:
Hi @dexterbg,

Thanks for your comments ! It was actually a fun challenge to write (my programming skills are kind of rusty) but I absolutely agree it's not practical at all.
I'll be trying a simpler approach, with recursion and less use of stack, and your suggestions - thanks for those !

OK for the protected tag on files, it's the right way to signal it.

llange:
Hi @dexterbg,

here is another take at it, taking into account your suggestions.
It is certainly naive and will need some back and forth before being OK.

I took the liberty to rewrite `vfs ls` with the common code, if not wanted you can drop the last commit.

While I'm at it, I found an issue with the protected path concept - there is a way to bypass the protection and list / cat / edit / ... the protected files / path - be it from the console, or via the remote shell.
Should I open a GitHub issue to track it, or is there a specific process for reporting security issues ?

Thanks for your feedback.

dexterbg:
I actually don't find much to change here. I only suggest renaming the methods as follows:
- `list_entries` → `read_entries`
- `list_entries_and_display` → `list_entries`
…and making both either `static` or moving them into the utils module.

> While I'm at it, I found an issue with the protected path concept

I'll send you an email about this.

dexterbg:
Oops, just saw this is missing an entry to `changes.txt` – never mind, I'll do that.

llange:
> Oops, just saw this is missing an entry to `changes.txt` – never mind, I'll do that.

Sorry, I missed it. Thanks for that.

PR #770, opened 2022-11-11
Use new power metrics in Hyundai kia/ioniq 5

frogonwheels:
The base modules had the power consumption doubled up with different metrics and using capitalisation and / characters.

This commit removes that from the base and inherited classes.  (An upcoming commit will allow access to metrics with different units)

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -7,6 +7,9 @@ Open Vehicle Monitor System v3 - Change log
     bms volt                            -- Output only voltage info if available
     bms temp                            -- Output only temperature info if available
 - Hyundai Ioniq 5: Initial support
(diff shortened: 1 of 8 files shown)


dexterbg:
The change needs to be documented as well, at least the Soul manual has a list of custom metrics.

@goev Geir, can you please review this as well, as this affects your vehicle modules?

Thanks,
Michael

goev:
Seems like sensible changes to me. I’m glad someone is doing something with these modules 🙂Best regards,GeirSendt fra min iPhone11. nov. 2022 kl. 12:41 skrev Michael Balzer ***@***.***>:﻿
The change needs to be documented as well, at least the Soul manual has a list of custom metrics.
@goev Geir, can you please review this as well, as this affects your vehicle modules?
Thanks,
Michael

PR #771, opened 2022-11-12
Add Unit Conversions to shell and duktape scripts 

frogonwheels:
I have left this as a block of commits as a single pull-request as
* they really form a single feature of unit conversions in scripts.
* there as already been feedback on the output and form of the functions/commands.

The inclusion of FeetPSS is really a part of the metric/imperial conversions.  (needing a 'short distance' version of metres per second squared).

I think I've got all the documentation done.

### DIFF
--- a/docs/source/userguide/scripting.rst
+++ b/docs/source/userguide/scripting.rst
@@ -827,11 +827,17 @@ OvmsMetrics
 
 - ``bool = OvmsMetrics.HasValue(metricname)``
     Returns whether the specified metric has a defined value.
+    Returns undefined if metric is un-registered.
 - ``str = OvmsMetrics.Value(metricname [,unitcode] [,decode])``
     Returns the typed value (default) or string representation (with ``decode`` = false)
     of the metric value optionally converted to the specified unit.
-- ``num = OvmsMetrics.AsFloat(metricname)``
-    Returns the float representation of the metric value.
+    Invalid ``unitcode`` or ``metricname`` will return invalid.
+    Mismatched ``unitcode`` will be ignored.
+- ``num = OvmsMetrics.AsFloat(metricname [,unitcode])``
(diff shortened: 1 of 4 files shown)


dexterbg:
Nice work, Michael!

Review results:
- Typo in `GrpPreassure`
- `GrpSound` should be called `GrpLevel` or `GrpSignal`
- The documentation needs to explain when exactly the Duktape API functions return `undefined`, i.e. `HasValue` when the metric does not exist, `Value` when either the metric or the unit does not exist, but `GetValues` ignores undefined units.
- Typo in the documentation example: `GetValue`
- The documentation should explain the effect of applying a non-matching unit to (a group of) metrics.
- There's at least one indentation inconsistency (used a TAB instead of spaces on line 642), please check for more.

Nice to have for the future:
- Split the metrics module, it's grown over 2.000 lines of code… maybe move the general unit support into a separate module?
- Auto completion for unit names, maybe also for metrics names

Regards,
Michael

frogonwheels:
Already found GrpPressure type. will move that fix forward.
Ditto  GrpSound. At the time I new it was wrong .. and I realised yesterday what it should be. I'll move that fix forward too.
Ah yes. HasValue - currently returns  Undefined when the metric does not exist-  should it just return false?
I'll add it to doco.. but applying a non-matching unit will 1) Not change the value and 2) If the units are printed.. print the original and correct unit.
I'll check the tabs thing. I was looking at it yesterday .. and I did see one.  There's a bit of a mix of <tab> and space indentation in that file.. which is 'fun'.   I should set up my 'listchars' in vim.. which I haven't yet.

dexterbg:
`HasValue` returning `undefined` for an unregistered metric name is good and consistent with the previous `Value` behaviour, we should just document it.

PR #776, opened 2022-11-18
Add user-configurable unit support

frogonwheels:
This series of patches adds the ability for a user to configure the units they want for the various types of metrics.  It extends the simple km/M configuration to Speed, acceleration, Pressure etc.

The general 'metric list' as an option to show user-configured units.
The TPMS shell modules will now use the user-configured units.  I think this should be the way it works, however I'm also prepared to have the default as it was and to have a (-u) option to show TPMS in user-configured units (or vica versa).

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -33,6 +33,7 @@ Open Vehicle Monitor System v3 - Change log
 - Add user configuration for groups of metrics
   Adds the 'ToUser' unit that converts to the user specified unit.
   Add -u to 'metrics list' to view metrics as  user units.
+- Add completion for metrics set/get as well as units
 
 2022-09-01 MWJ  3.3.003  OTA release
 - Toyota RAV4 EV: Initial support added. Only the Tesla bus is decoded and just listening so far.
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -29,6 +29,10 @@ Open Vehicle Monitor System v3 - Change log
   New events:
     vehicle.charge.type                 -- Vehicle charge connection type has changed (e.g. ccs/type2/…)
     vehicle.gen.type                    -- Vehicle generator connection type has changed
+- Add units Bar, Permille
+- Add user configuration for groups of metrics
+  Adds the 'ToUser' unit that converts to the user specified unit.
+  Add -u to 'metrics list' to view metrics as  user units.
 
 2022-09-01 MWJ  3.3.003  OTA release
 - Toyota RAV4 EV: Initial support added. Only the Tesla bus is decoded and just listening so far.
--- a/vehicle/OVMS.V3/components/ovms_location/src/ovms_location.cpp
+++ b/vehicle/OVMS.V3/components/ovms_location/src/ovms_location.cpp
@@ -290,9 +290,10 @@ void OvmsLocation::Store(std::string& buf)
 
 void OvmsLocation::Render(std::string& buf)
   {
-  char val[32];
-  snprintf(val, sizeof(val), "%0.6f,%0.6f (%dm)", m_latitude, m_longitude, m_radius);
-  buf = val;
+  metric_unit_t user_length = OvmsMetricGetUserUnit(GrpDistanceShort, Meters);
+
+  buf = string_format("%0.6f,%0.6f (%d%s)",
+      m_latitude, m_longitude, UnitConvert(Meters, user_length, m_radius), OvmsMetricUnitLabel(user_length));
   bool first = true;
   for (ActionList::iterator it = m_actions.begin(); it != m_actions.end(); ++it)
     {
@@ -409,12 +410,35 @@ void location_radius(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int ar
     writer->printf("Error: No location %s defined\n",name);
     return;
     }
+  metric_unit_t user_length;
+  if (argc < 3)
+    user_length = OvmsMetricGetUserUnit(GrpDistanceShort, Meters);
+  else
+    {
+    user_length = OvmsMetricUnitFromName(argv[2]);
+    if (user_length == UnitNotFound)
+      {
+      writer->printf("Error: Invalid Metric %s\n", argv[2]);
+      return;
+      }
+    user_length = OvmsMetricCheckUnit(Meters, user_length);
+    if (user_length == UnitNotFound)
+      {
+      writer->printf("Error: Metric %s is not a length unit\n", argv[2]);
+      return;
+      }
+    }
 
   std::string buf;
   OvmsLocation* loc = *locp;
-  loc->m_radius = atoi(argv[1]);
+  int base_value = atoi(argv[1]);
+  int radius_m = UnitConvert(user_length, Meters, base_value);
+  loc->m_radius = radius_m;
   loc->Store(buf);
-  writer->puts("Location radius set");
+  if (user_length == Meters)
+    writer->printf("Location radius set to %dm\n", base_value);
+  else
+    writer->printf("Location radius set to %d%s = %dm\n", base_value, OvmsMetricUnitLabel(user_length), radius_m);
   }
 
 void location_rm(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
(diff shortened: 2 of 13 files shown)


dexterbg:
Michael,

one issue: `OvmsMetricGetUserUnit()` doesn't return the `units.distance` derived defaults for speed, accel etc., so without a default configuration for these, output of "stat trip" would be inconsistent after the update. Instead of adding the previous logic to the method, I suggest to update `cfgversion` and init the new config instances accordingly in `OvmsConfig::upgrade()`.

A minor code style remark: the web server component uses Stroustrup bracket / indentation style.

The shell is the text UI, I also think user messages generated by commands should probably apply the user configured units -- as did/does "stat" & "stat trip" already. Taking this concept further means, "metrics list" should also by default output in user units, as well as any other commands creating reports from metrics data. From a quick check, besides "metrics" it seems only the "location" commands are also affected currently (except vehicle specific commands), these also take a radius argument.

I think config instances could remain unitless for now, users normally don't manipulate these directly.

I noticed you added "To Check" to my comment about '°' not being SMS safe. That's still true, SMS encoding is GSM-7 by default. Characters outside that encoding need UCS-2, which limits the SMS size to 70 characters, pretty much useless. OTOH, we don't use SMS at all in V3 now, and if (!) we reintroduce it some day, messages most probably will need an automatic split/recode mechanism anyway. IOW, I think we no longer need to restrict ourselves to GSM-7 for shell outputs.

Regards,
Michael

frogonwheels:
Thanks! Comments inline.

On Fri, 18 Nov 2022 at 23:54, Michael Balzer ***@***.***>
wrote:

> Michael,
>
> one issue: OvmsMetricGetUserUnit() doesn't return the units.distance
> derived defaults for speed, accel etc., so without a default configuration
> for these, output of "stat trip" would be inconsistent after the update.
> Instead of adding the previous logic to the method, I suggest to update
> cfgversion and init the new config instances accordingly in
> OvmsConfig::upgrade().
>
Ooh... excellent.  I didn't know about/notice the upgrade() - will do
that.. and yes .. an upgrade step to update anything with distance in it
makes absolute sense! Hmm.. on that note - I'll probably only change them
(from blank) if they setting is miles! This also means I can change the
setting for units.distance to the unit names rather than trying to maintain
backwards compat?  Or do other plugins rely on that?

> A minor code style remark: the web server component uses Stroustrup
> bracket / indentation style.
>
AHh crap. I've really be trying to be conscientious about maintaining
styles. That was a weird one to not notice. I much prefer the Stroustrup
style. *sigh*.

> The shell is the text UI, I also think user messages generated by commands
> should probably apply the user configured units -- as did/does "stat" &
> "stat trip" already. Taking this concept further means, "metrics list"
> should also by default output in user units, as well as any other commands
> creating reports from metrics data. From a quick check, besides "metrics"
> it seems only the "location" commands are also affected currently (except
> vehicle specific commands), these also take a radius argument.
>
Ok - I can get behind that.   What I might  do then is to add an option for
'native' unit display for metrics list.
I'll look into the 'location' commands as well.

> I think config instances could remain unitless for now, users normally
> don't manipulate these directly.
>
Agreed.

> I noticed you added "To Check" to my comment about '°' not being SMS safe.
> That's still true, SMS encoding is GSM-7 by default. Characters outside
> that encoding need UCS-2, which limits the SMS size to 70 characters,
> pretty much useless. OTOH, we don't use SMS at all in V3 now, and if (!) we
> reintroduce it some day, messages most probably will need an automatic
> split/recode mechanism anyway. IOW, I think we no longer need to restrict
> ourselves to GSM-7 for shell outputs.
>
Ah yes, this makes sense.  I also noticed that various other commands
seemed to be using characters outside the normal range. Good to know.  I
might clarify that comment though.

dexterbg:
Another thought: the suggestion by @jollyjinx Patrick to provide a simple choice between "metric" & "imperial" would make sense for the setup wizard (module `web_cfg_init`).

> This also means I can change the
setting for units.distance to the unit names rather than trying to maintain
backwards compat?  Or do other plugins rely on that?

We cannot know if there are unpublished V3 user plugins or forks relying on the current codes.

The codes `K` & `M` are legacy from OVMS V2 and used by all V2 clients (e.g. the Apps). They are defined by the MP and V2 parameter specification so need to be kept in the MP/V2 communication. `server_v2` could of course do the translation.

My general approach is, if we don't need to change codes for technical reasons, we should avoid that.

Agreed, having a mix of legacy and new unit codes in the config is ugly, but – again – users don't normally interact with the config directly. If we provide standard APIs & commands to query the user unit configuration, these can translate the legacy codes as well.

So, as there are no hard indications for the config code change (or are there?), I'd rather keep `K` & `M`.

@markwj Mark, what's your opinion on this?

frogonwheels:
Yep. I'll stick with what I've got which leaves it as K and M.

Michael.

frogonwheels:
I have a commit ready that could either be added to this or as a separate p/r that adds metric/unit completion to the metrics commands as well as unit completion for radius of location.
The completion for metrics get/set is sensitive to the unit being get/set iow it only allows/completes units of the same (collapsed) group. 
In the radius - the full list would be   native, metric, imperial, km, miles, metres and feet.

dexterbg:
Michael, I don't have a preference for the PR, do as you think fits best. I'll wait for your info to avoid reviewing twice. Please also resolve the conflicts in this PR – thanks!

dexterbg:
Michael, I've built your branch here and it works very nicely, especially the metric & unit completion improves usability substantially. :+1: 

I found only one new & minor side issue, the new `string_format()` utility has bugs:
- Using a negative `vsnprintf()` result as a memory resize indication
- Not validating the memory allocation result
- Copying the `fmt_str` into the output buffer before the `vsnprintf()` call -- remainder from a previous version?

I'd btw rather use `vasprintf()` for this, see `StringWriter::printf()` for a typical implementation.

If sticking with `vsnprintf()`: the attempt to avoid an initial size scan run for `vsnprintf()` with an assumed sufficient size of twice the format string will in most cases actually result in a necessary second allocation. The initial size should either be at least factor 4-5 or maybe simply a fixed size fitting most typical uses -- or could also be a parameter.

Regards,
Michael

frogonwheels:
Hopefully that will work - vasprintf does make more sense.. found an example of it and added in the check for -ve result. 
I hope the error handling makes sense (log the error and return the original string).

dexterbg:
I see the problem with error handling in this case. Logging from utility methods is bothersome though, as is returning a format string as a result.

Generally, logging should always be done by (or at least in the name of) the component in charge, and it must not be done by a low level utility (or needs to be optional). Logging isn't allowed in all contexts (e.g. not in timer callbacks), because it involves locking potentially blocking mutexes (for the system log queue).

Returning the format string in case of an error means the caller would need to compare the result with that to detect an error. If the caller fails to do so, the format string may end up being sent as a value to some processor, possibly causing all sorts of issues. IMO the method should return an empty string in case of an error.

Regarding format string errors, there's a nice compile time solution provided by gcc: the [format attribute](https://gcc.gnu.org/onlinedocs/gcc-6.1.0/gcc/Common-Function-Attributes.html).

This requires the method to use `const char*` for the format string, but that's actually more appropriate anyway, as these are normally fixed strings. So that would be:
```
std::string string_format(const char* fmt_str, ...) __attribute__ ((format (printf, 1, 2)));
```
Like this, gcc will emit errors when detecting an invalid format string or mismatching varargs.

Regards,
Michael

PS: just saw we actually didn't use the format attribute for our other printf style methods up to now, I'll add that.

frogonwheels:
Thanks for the explanation Michael.  I nearly left it as const char * .. so happy to put that there and add the attribute- -that's cool.

dexterbg:
Adding the attribute to our methods actually revealed a couple of missing `.c_str()`. Oops, I produced a new merge conflict, sorry… ah, yes: you simply need to replace the `%lu`s by `%u` in that block.

frogonwheels:
Ah, all good.  Getting format strings right is a Good Thing!

frogonwheels:
Ahh. Thanks.  I will fix the label issue - I think I might have seen this
in a different context just yesterday.

//.ichael

frogonwheels:
Ahh, I actually picked up this same thing in temperature display elsewhere
and added a default unit parameter to OvmsMetricGetUserUnit() for exactly
this purpose.  Missed it here though.

I just need to provide the default unit. Will get a p/r going for that.

//.ichael

frogonwheels:
The core problem here is that the 'Native' unit of a metric is defined by
the metric itself... ie it's the value the metric is stored in.  Currently,
I have not introduced the concept of a 'Default' unit for a Unit Group -
Currently 'Default' means whatever unit the individual Metric itself is
stored in.  I think what I have done to fix it is an acceptable solution..
but it does highlight the underlying issue.

//.ichael Geddes

PR #778, opened 2022-11-20
add configurable metrics to CAN logs, and allows configuration of events

llange:
This PR adds the ability to log some metrics to the CAN logs ; and allows the configuration of which events are going to be logged (event logging was already existent)

# Metrics
A new configuration item `can log.metrics_filters` allows to choose which metrics will be sent to the can logs.

The default configurations is NOT to log any metric.

The metrics are sent to the can log subsystem as events. When using the CRTD format for example, those will appear as `1CEV` lines.

The metrics are sent as a JSON object prefixed with `metric: `, having the following fields:
* `name` : the name of the metric
* `value`: the value of the metric in the default unit
* `unit`: the default unit of the metric

Example:
```
metric: { "name": "v.p.gpstime", "value": 1668730982, "unit": "Sec" }
```

Example in a CRTD file:
```
1668730982.038759 1CEV Event metric: { "name": "v.p.gpstime", "value": 1668730982, "unit": "Sec" }
```

A possible configuration to log all GNSS-related metrics:
```
config set can log.metrics_filters "v.p.gps*, v.p.latitude, v.p.longitude, v.p.altitude, v.p.direction, v.p.satcount"
```

# Events
A new configuration item `can log.events_filters` allows to choose which events will be sent to the can logs.

To maintain compatibility, the default value is `x*,vehicle*`, which is going to allow only the events starting with `x` or with `vehicle`.

The default value is equivalent to entering the following command:
```
config set can log.events_filters "x*,vehicle*"
```

# Format of filters
The format of filters is a comma-separated list of metrics filters, possibly containing wildcards (`*`):
* Any filter starting with `*` is an `endsWith()` comparison.
* A filter ending with `*` is a `startsWith()` comparison.
* A filter without any wildcard is a `==` comparison.

(A wildcard in any other position in the filter is invalid)

Values are trimmed, so the presence of leading / trailing spaces is not an issue.
Invalid or empty filters are skipped (and a warning is sent).

### DIFF
--- a/docs/source/crtd/can_logging.rst
+++ b/docs/source/crtd/can_logging.rst
@@ -1,3 +1,5 @@
+.. highlight:: none
+
 ====================
 CAN Bus Data Logging
 ====================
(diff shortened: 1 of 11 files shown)


dexterbg:
Ludovic, nice idea & implementation.

One major thing: metrics data should not be logged as events, these need a dedicated log type & tag. Filtering events from a log is normally done by the tag code `CEV`, this should remain unique.

I suggest extending `CAN_log_type_t` by `CAN_LogInfo_Metric` and introducing a new CRTD standard comment tag `CMT` or `CMD` (metrics data) (doc: https://docs.openvehicles.com/en/latest/crtd/index.html#comment-records).

You could also log the new filter configuration using type `CAN_LogInfo_Config`, as is done with bus/id filters.

Minor code remarks:
- Please also use `CAN_PARAM` on getting the queue size.
- The `canlog` modules use Whitesmith bracket/indentation style.
- The number of operators supported (currently 3) should get a readable name.

Regards,
Michael

llange:
Hi @dexterbg , thanks for your comments !
I've implemented most of the changes if not all - I'm still not satisfied with the use of a constant for the number of operators but here it is until I find something better.

llange:
Hi @dexterbg , thanks for your comments. Let me know if this update is in the right direction.

PR #782, opened 2022-11-25
OvmsHttpClient : case-insensitive comparison of `Content-Length` header

llange:
I had some issues fetching OTA firmware from some servers (S3-compatible buckets), because of the headers that were all lower-case.
In that case, the OvmsHttpClient returns a size of 0, and the OTA fails with:
```
Error: Expected download file size (0) is invalid
```

As HTTP Header fields names are case-insensitive, having a case-insensitive comparison means that we can be compatible with more servers.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_http/src/ovms_http.cpp
+++ b/vehicle/OVMS.V3/components/ovms_http/src/ovms_http.cpp
@@ -136,7 +136,7 @@ bool OvmsHttpClient::Request(std::string url, const char* method)
         {
         // ESP_LOGI(TAG, "Got response %s",m_buf->ReadLine().c_str());
         std::string header = m_buf->ReadLine();
-        if (header.compare(0,15,"Content-Length:") == 0)
+        if (strncasecmp(header.c_str(), "Content-Length:", 15) == 0)
           {
           m_bodysize = atoi(header.substr(15).c_str());
           }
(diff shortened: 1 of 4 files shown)


dexterbg:
Ludovic, again nice find :+1:

Your solution isn't strictly wrong, but this is a nice example of how trying to do everything "the C++ way" sometimes leads to unnecessary complex and inefficient, potentially even dangerous code.

The solution could be a simple…
```
if (strncasecmp(header.c_str(), "Content-Length:", 15) == 0)
```
…which would also be more readable IMO.

Besides the need to introduce a new method, `std::tolower()` is locale aware (unnecessary overhead), and `str_tolower()` needs to construct a copy of the string to compare first -- `strncasecmp` does an inline comparison and stops at the first mismatch -- and in this case also copies the full header line. It could be cut down to copying the first 15 bytes, but that introduces another temporary string and makes the comparison even harder to read.

The inefficiency of the comparison really doesn't matter in this case, and injecting a maliciously oversized "Content-Length" header field would be stopped by the `OvmsBuffer` before. Code patterns get copied though, and may end up at a place where performance and attack vectors do matter.

`str_tolower()` still may be a useful utils addition for other uses.

So… TL;DR: I suggest keeping `str_tolower()` as a utils addition (with some caveat warning in the comment), but changing the header comparison to use `strncasecmp()`.

Btw, there may be other place we do case sensitive HTTP header comparisons, may be worth checking if you didn't do so already.

Regards,
Michael

llange:
Yes, I was under the (false) impression that a C++ approach was preferred as much as possible. As I have very little experience with C++ / Standard library I will certainly make some of these mistakes again.
I'll adapt and improve :-)

Here is an updated version. You're right that there are some other places that have this kind of case-sensitive comparison, those I found I fixed in a similar way.
However, I only tested the `ovms_http.cpp` so it is possible that I've introduced issues here - please double check.

PR #783, opened 2022-11-28
Bolt EV: Cleanup m_type layout, don't statically be Chevy Ampera-e

kallisti5:


kallisti5:
Fixes m_type to be dynamic per manufacturer.

* VBE = cheVy Bolt Ev
* OAE = Operl Ampera-e (Ev)  (kinda stealing the E lol)
* VBU = cheVy Boly eUv (SUV version of Bolt EV)

Also fixes some weird + incorrect logic around level 1 max charge rates.

PR #786, opened 2022-11-29
remove some of the warnings regarding unitialized variables

llange:
GCC 8.4.0 throws some warnings and we try to fix them.

Not sure if this is always the right way, please review and comment.


llange:
You're right, those changes are completely wrong.
What's more puzzling is that I cannot reproduce the error message now, I may have had a confusion of compiler flags when switching between ESP-IDF versions.
Let's forget about this PR ; if I face those issues again I'll find a proper way to fix them.

PR #789, opened 2022-11-30
Support for Mini Cooper SE

vincentbarat:
Support for Mini Cooper SE added.
It is directly derived from the BMW i3 support, which has the same BMS.
BMS Monitor, cells voltages and modules temperatures are available for this vehicle (which was the main purpose of this PR).


dexterbg:
Awesome :+1: 

Only missing are an entry to the `changes.txt` and the documentation link in the toplevel `README.md`. I can add these later on, or you can do so if you like. Ah yes, and `sdkconfig.default.hw31` should be extended by the option.

vincentbarat:
I did all changes but changes.txt.
Pull request to come...

-
Vincent Barat - Woleet CTO
***@***.***

PR #795, opened 2022-12-09
Add support for user defined units in web interface 

frogonwheels:
This series of commits:
  * adds a new 'units' stream that defines what the current user defined units, labels and native units are for each metric
  * adds a clearer way of generating the gauge information based on user-defined units.  This also will allow easier transition over to where the gauge ranges could be recalculated within the web-page.
  * fixes a bug where sending multiple pages of 'metrics' value changes was not resuming from where it left off. (It's not critical as over a few iterations it would have sent everything anyway). It's here because it depends on changes I made for the 'units' stream.
  * Adds new proxy collections metrics_user and metrics_label to allow modules to get at units and labels for individual metrics.


frogonwheels:
Fixes coming:
 * Properly translate the incoming units message to a metrics message
 * Fix issue with dbm/sq conversions
 * Support automatically updating form values with Unit and Value elements with converted values.

dexterbg:
Michael,

my comments so far on the current (?) state:

- Subscription now refers to the units dictionary, my original idea for a subscription here
  referred to the additional `…#…` user metrics, which would have added substantial load.
  But the dictionary is mostly static, only tranmitted once on a new WebSocket connection.
  IOW do we actually need an option to explicitly enable the dictionary transmission?

- If (!) we keep the explicit subscription for the dictionary, the `.receiver` should
  get a `data` attribute that `$.fn.sendUnits()` checks instead of generally doing
  `metric units on`.

- The unit conversion functions are now defined as jQuery plugins (`$.fn.…`) -- this is
  misleading and doesn't fit their applications, except for `sendUnits()` (if we keep
  it and it looks for the `data` attribute proposed above).

- Please generally keep using the `;` command separator in our javascripts (ovms.js).
  Omitting the separator is bad, as it cannot always be omitted, but depends on the
  context of the previous & next lines. IMO that unnecessarily introduces a potential for
  bugs resulting from merges / cherry-picks / code block copying.

- Typo/misspelling: "guage" (consistently, even in comments)

- `WSTX_MetricsUnitUpdate` should be called `WSTX_UnitsUpdate`, as the units dictionary
  is meant to cover configs as well…

- …but it now doesn't have the configured group units as proposed.
  These are needed to do non metrics related conversions, e.g. for configurations.
  But as these are missing, and with possible later addition of individual config instance units in mind,
  maybe better add separate structure levels for metrics & configs, i.e.:
```
units: {
  config: {
    "vehicle": {
      "units.distance": { native: … },
      "units.speed": { native: … },
      …
    }
  },
  metrics: {
    …
  }
}
```

- Good & correct: `data-scale` present disables auto unit conversion & replacing.
  Keep in mind this needs to be documented.

- On the `m_sent` fix: nice find, please also synchronize the comments.

- You can save some whitespace in the `units` message.

- There's indentation by tabs instead of spaces in the encoder utils in `web_framework.cpp`.

- `OvmsMetricSetUserConfig()` should not be responsible for triggering metrics units updates,
  instead you should introduce a config listener, so any configuration change on the group
  units will be catched, not only those being done by `OvmsMetricSetUserConfig()`.

- It may be a good idea to refactor `UserMetricSnapshot` into a generally available
  unit cache that's automatically used by `GetUserUnit`.

Regards,
Michael

frogonwheels:
I have been working through all of these. Using the Snapshot object as a
global cache actually works quite well along with the message sink in only
triggering the units that actually changed to be sent (for eg if the user
changed 'distance' to 'miles' only those miles units would be sent).
I've now got  units = { metrics: {}, vehicle: {} } as a dictionary.  The
message is like that but always sent separately because otherwise you have
to keep track of where you are up to and in which collection.

I've changed it to a 'subscription'  of units/metrics and units/vehicle,
primarily so that any other users of the websocket can subscribe to it
separately.
For the moment, I explicitly subscribe to units/#  in the 'onopen' event of
the socket.  The next point where I'm trying to introduce subscribing to
only the metrics required may see that change perhaps.

The one thing I haven't done in vehicle: {} is provide a 'native' unit,
since we currently do not have a concept of 'native' as applied to groups
of units.
I have at least provided helper functions to get at the vehicle config,
allowing the plugin implementor pass in what _they_ consider to be their
default.  I have certainly considered it, but without enforcing it somehow
in the creation of metrics.. I don't see quite how to work around it.

I've done most of the implementation now but still have to get to the
documentation which I will do before pushing up.

//.ichael

dexterbg:
Michael,

I haven't checked the code yet.

> I've now got  units = { metrics: {}, vehicle: {} } as a dictionary

The idea about the `config` object was being able to replicate whatever config parameters we have in that object, with `vehicle` just being the first. Technically `vehicle` is special, as it doesn't describe a config instance unit, so you may be right in separating this. In that case I would opt for another name though, to reflect the meaning, e.g. `default` or `prefs` (as these are not the actual "vehicle" units), and possibly also remove the `unit.` prefix from the name, as we're already in the `units` object.

Regarding the native units for these, I thought this would simply reflect the system defaults if nothing is configured. But you're right, these depend on the actual use case. It seems to be able to let the user set configuration instances in their preferred units, we actually need the individual config instance native unit.

Regards,
Michael

frogonwheels:
So this is what that object now looks like in the web-page javascript:
units:
{
  metrics: [
    ..
    m.hardware:{native: 'Other', code: 'Other', label: ''}
    m.monotonic:{native: 'seconds', code: 'seconds', label: 'Sec'}
    m.net.mdm.iccid:{native: 'Other', code: 'Other', label: ''}
    ...
    v.b.12v.current:{native: 'amps', code: 'amps', label: 'A'}
    v.b.12v.voltage:{native: 'volts', code: 'volts', label: 'V'}
    ..
  ],
  prefs: [
    accel:{unit: 'Native', label: ''}
    accelshort:{unit: 'Native', label: ''}
    consumption:{unit: 'Native', label: ''}
    distance:{unit: 'Native', label: ''}
    distanceshort:{unit: 'Native', label: ''}
    energy:{unit: 'Native', label: ''}
    power:{unit: 'Native', label: ''}
    pressure:{unit: 'Native', label: ''}
    ratio:{unit: 'Native', label: ''}
    signal:{unit: 'Native', label: ''}
    speed:{unit: 'Native', label: ''}
    temp:{unit: 'Native', label: ''}
  ]

  convertMetricToUserUnits:ƒ (value, name)
  unitLabelToUser:ƒ (unitType, defaultLabel)
  unitValueToUser:ƒ (unitType, value)
  userUnitLabelFromMetric:ƒ (name)
}

//.ichael

dexterbg:
Michael,

sorry for the delay, I've been busy. I'm now testing your latest PR state on my bench module, it seems to work quite well now (besides a small issue with the cache, see below). The units can be changed nicely, and a second session updates dynamically on the configuration.

On your changes since my last review, I mostly have naming consistency issues:

- The event name `msg:vehicle_units` breaks consistency with the new object scheme ("prefs"). I suggest using e.g. `msg:unitprefs` or `msg:units:prefs` -- and on the latter, as the members of the units object are sent, the `msg:units` could also be changed to `msg:units:metrics` to make this a consistent naming scheme. That would make a later addition for `msg:units:config` (or others) fit seamlessly.
- The naming scheme on the server side is also inconsistently using "vehicle" instead of "prefs".
- The subscription topic also should now be `units/prefs`, not `units/vehicle`
- `docs/metrics.rst` also still mentions `units.vehicle` and does not yet reflect the splitting of the events for metrics & prefs. It also should mention that having `data-scale` on a number metric widget disables user unit conversion.

- `UserMetricMap` for the units cache is a misnomer, it's not a map of user metrics but of user unit prefs. These are not strictly bound to metrics, so we should avoid confusion here. I'm also a bit unsure if this cache justifies a separate singleton, it can also be seen as a specialized cache extension of the `OvmsConfig` class.
- If we keep it separate, the units cache needs to listen to event `config.mounted` for the initial `Load()`, instead of doing that in the constructor. Not doing so currently leads to a race condition with the config store mounting process, with the effect that the cache may be empty/invalid after a reboot, and remain invalid until some vehicle parameter is changed.

- The config upgrade logic for `units.preasure` looks strange: it only keeps the previous value if there also is an instance value for `units.prssure`…? Also, I think the upgrade should delete the `units.preasure` instance in any case.

Btw, I saw multiple comments referring to the `$.fn.` plugin registry as being NodeJS -- jftr, it's jQuery, not Node. A jQuery plugin is meant to iterate over an array of DOM elements passed to it, and return that same array for chaining.

Regards,
Michael

frogonwheels:
Oh dear - a bit of a brain f*rt on the jQuery thing. Sorry about that.  Only in git comments it seems? I guess I can't do much without force pushing ... so  *sigh*. I understand.. and it's just that I'm not totally familiar with either.. though if I think about it, I kinda know.

 The subscription topic was already renamed.. not sure what your were looking at there .. did I miss it somewhere? I did a git grep and couldn't find anything.

On the upgrade thing  units.prssure is meant to be units.pressure  .. and I managed to miss DeleteInstance() when I went looking .. so I was just setting it to blank. That will be fixed.

All the rest:  absolutely.  I'll go with msg:uni uts:prefs and msg:units:metrics.  And you're right; it doesn't really need a separate singleton so I'll merge it into OvmsMetrics.

dexterbg:
You're right, the subscription topic is fine already -- seems I somehow messed up the github diff view parameters, can't reproduce it now.

frogonwheels:
I might keep the singleton as-is (just renamed to reflect its use as User
Unit prefs), and use the mounted event.

The problem is that putting it into the config singleton adds depency on
ovms_metrics.h in there and also would need some extra functions to access
the Group info.
It's also a bit weird as I would still need the event to get at the
'vehicle' changes..

Michael

vehicle/OVMS.V3/main/ovms_metrics.h
```diff
@@ -1110,7 +1111,7 @@ class OvmsMetrics
   };
 
 extern OvmsMetrics MyMetrics;
-extern UserMetricMap MyUserMetrics;
+extern UserUnitConfigMap MyUserUnitConf;
```
dexterbg, line 1114: I'd rather remove `User` from both the class and the instance name, and then not abbreviate `Config`. `User` is redundant, as unit configuration implies user configuration. `Conf` as an abbreviation for `Config` is somewhat unusual, as `Config` is already an abbreviation. But that's more a personal style choice.

vehicle/OVMS.V3/main/ovms_config.cpp
```diff
@@ -582,9 +582,9 @@ void OvmsConfig::upgrade()
     auto val = GetParamValue("vehicle", "units.preasure");
     if (val != "")
       {
-      if (GetParamValue("vehicle", "units.prssure") != "")
+      if (GetParamValue("vehicle", "units.pressure") != "")
         SetParamValue("vehicle", "units.pressure", val);
-      SetParamValue("vehicle", "units.preasure", "");
+      DeleteInstance("vehicle", "units.preasure");
```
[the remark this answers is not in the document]
frogonwheels: Yep, makes sense.

vehicle/OVMS.V3/main/ovms_metrics.h
frogonwheels: Consider it done.  Conf is an abbreviation I've use but I've no arguments against making it Config. Especially when making it 'MyUnitConfig' .. and you're right,  'My' makes 'User' redundant.

PR #796, opened 2022-12-09
In vehicle stat, don't show charge mode if metric undefined.

frogonwheels:


dexterbg:
Michael,

this is wrong, a defined charge mode isn't meant to be mandatory to enable charge state & detail information.

The charge mode itself already isn't shown with the original code if undefined, so I also don't understand the issue this is supposed to fix.

Regards,
Michael

frogonwheels:
Ack. Closed

PR #797, opened 2022-12-09
Config Wizard - Construct the status checklist from a list.

frogonwheels:
This patch was made when I thought I was needing to add a page to the wizard, which I ended up not doing, so figure I'd submit it.

This generates the config wizard headers from a list rather than having copy pasted versions of them. It also allows for the pages reliant on a `#define` to be removed from the wizard display.


dexterbg:
Michael, as this is incomplete / WIP, I've tagged this PR as a "draft".

Missing are the corresponding changes to the assumed fixed step sequence by `HandleCfgInit()` / `CfgInitSetStep()` et al.

PR #798, opened 2022-12-11
Mgev trip

GreyPeter:
Add code to produce the Trip Metric for the MGEV.


dexterbg:
Peter,

that will result in a rather coarse trip distance measurement, as your odometer resolution is 1 km. If you drive 1.9 km, all energy statistics will be done for a distance of 1 km. So energy data will be wrong/misleading for all short trips.

My solution for this on the VW e-Up and Hyunda Ioniq vFL (just pushed) is to derive the trip distance from the speed integral. If your speed measurement is accurate, you might consider copying that:

https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/914a5bb1b8811fbacabde3da1e76856ed27aba2f/vehicle/OVMS.V3/components/vehicle_hyundai_ioniqvfl/src/vehicle_hyundai_ioniqvfl.cpp#L460-L474

Should be simple to adapt. Do you want to give that a try, or shall I merge this PR now?

Regards,
Michael

GreyPeter:
Thanks Michael,

Yes I will try to adapt the your code and test it out.
I will remove the pull request for now.
 Kind regards,
Peter

PR #811, opened 2023-01-31
Improve Ioniq 5 Metrics and displays

frogonwheels:
Fix the range and charge estimation algorithms.
Make sure the sdcard is mounted/available before writing trip stats to it.
Mirror the average cell temp as the battery temperature.
Tweak the 12V battery monitor to increase sensitivity.
Add Ah / coulomb count to trip metrics.
Add RPM 
Improve responsiveness of RPM and Speed.

Adds an alert if the vehicles is unlocked and off (I meant to separate this one out but didn't).


frogonwheels:
There's a compile issue. Fix coming.

frogonwheels:
Fixed compile issue where I was referencing a metric group that  I was going to leave off for later.. including it.

dexterbg:
Michael, I'll merge this now as is, but please try to avoid mixing framework PRs with vehicle specific PRs. Split these into two successive PRs. The reason for this is to be able to revert each part individually, also to have a clean documentation of changes to the framework.

Thanks!

frogonwheels:
Understood. I initially did that but My mistake was not to compile the
rebased portion without framework changes... There was a metric group enum
I had used from the excluded commit. I'll be more careful next time.

Michael

PR #831, opened 2023-02-20
cleanup warnings about switch case fallthrough

llange:
With the new warnings activated in ESP-IDF 4+, a `switch`-`case` must now properly and explicitely identify when a fall-through is wanted or is a forgotten `break`.

One way to mark those wanted fall-through is to mark them with a [GCC-specific attribute](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough), a [C++ attribute](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) or even with a specific comment as [documented in the GCC compiler](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough).

We took inspiration from the [Linux Kernel approach](https://elixir.bootlin.com/linux/v5.10/source/include/linux/compiler_attributes.h#L208) to create a macro `FALLTHROUGH;` that should do the proper decoration (either CPP attribute, compiler attribute or comment) depending on the capabilities of the compiler - when the fall-through is wanted, and explicitly decorated those.
For the other cases, we added a `break`.

⚠️ We may have mistaken a missing break for a wanted fall-through, or vice-versa - please double / triple check during code review.... thanks !

### DIFF
--- a/vehicle/OVMS.V3/components/simcom/src/simcom_5360.cpp
+++ b/vehicle/OVMS.V3/components/simcom/src/simcom_5360.cpp
@@ -91,6 +91,7 @@ void simcom5360::StatusPoller()
         //  +COPS: 0,0,"vodafone.de Hologram",0
 
         // done, fallthrough:
+        FALLTHROUGH;
       default:
         m_statuspoller_step = 0;
         break;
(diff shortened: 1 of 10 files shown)


dexterbg:
Nice example of compiler warnings actually helping find bugs. :+1:

PR #841, opened 2023-02-24
ESP-IDF v5+: update WatchDogTimer initialization

llange:
We introduced the dependency on the (official) configuration item `CONFIG_ESP_TASK_WDT` to decide whether (or not) we want to initialize the WDT. In ESP-IDF v5+, it's not possible any more to (re)initialize and already initialized timer, and if `CONFIG_ESP_TASK_WDT` the WDT would have been initialized already during boot.

Note: For some reason, in ESP-IDF v5+, we need to define `CONFIG_ESP_TASK_WDT=n` otherwise it won't work as expected.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms.h
+++ b/vehicle/OVMS.V3/main/ovms.h
@@ -40,6 +40,7 @@
 #include <string>
 #include <sstream>
 #include "ovms_malloc.h"
+#include "esp_idf_version.h"
 
 #if ESP_IDF_VERSION_MAJOR >= 4
 #define CONFIG_CONSOLE_UART_NUM CONFIG_ESP_CONSOLE_UART_NUM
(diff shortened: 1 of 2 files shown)


dexterbg:
That may be partially wrong. Quoting myself from the list:

>  Doing the watchdog init ourselves avoids the issue.
> To do so, we need to change line 38 in ovms_main to…
> `#if !CONFIG_ESP_TASK_WDT_INIT`
> …and disable that config option.

`CONFIG_ESP_TASK_WDT` controls the overall WDT support, `CONFIG_ESP_TASK_WDT_INIT` can be used to disable just the auto WDT init during boot. I still need to do that to get the watchdog to work with IDF5, so I suggest doing it that way.

llange:
You're certainly right, I believe the situation regarding the naming of this configuration variable is a little bit complex:

According to documentation, the name of the variable:
* Was `CONFIG_TASK_WDT` [in 3.3.6](https://docs.espressif.com/projects/esp-idf/en/v3.3.6/api-reference/kconfig.html#config-task-wdt)
* Has been renamed to `CONFIG_ESP_TASK_WDT` in both [4.4.4](https://docs.espressif.com/projects/esp-idf/en/v4.4.4/esp32/api-reference/kconfig.html#config-esp-task-wdt) and [5.0](https://docs.espressif.com/projects/esp-idf/en/v5.0/esp32/api-reference/kconfig.html#config-esp-task-wdt)
* Has been renamed to `CONFIG_ESP_TASK_WDT_INIT` in [5.0.1](https://docs.espressif.com/projects/esp-idf/en/v5.0.1/esp32/api-reference/kconfig.html#config-esp-task-wdt-init)

So let's target 5.0.1, use the 5.0.1 name, and we'll see if there are other changes in 5.0.1 that could impact us.

llange:
For the moment I'm handling multiple version: < 5, == 5.0.0, >= 5.0.1+  
I'm still validating 5.0.1 but it seems to work OK.

PR #849, opened 2023-03-03
Valet Mode - add default valet mode with simple geofence

frogonwheels:
- Simply allowing 'Valet mode' without hardware support gives door detection anyway..
 - This adds the ability to have a Valet mode geofence radius from start location.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_location/src/ovms_location.cpp
+++ b/vehicle/OVMS.V3/components/ovms_location/src/ovms_location.cpp
@@ -700,6 +700,11 @@ OvmsLocations::OvmsLocations()
   m_park_distance = 0;
   m_park_invalid = true;
   m_last_alarm = 0;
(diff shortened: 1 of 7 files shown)


dexterbg:
Michael, nice addition. This needs a bit of user documentation though, at least an entry to the `changes.txt` file.

frogonwheels:
Ah yes.  Seeing as this is merged, how about I add a web config for it
(under vehicles) and add the doco?

Sorry about forgetting the doco. I'll get that done.

//.ichael

dexterbg:
Reminded me of: https://docs.openvehicles.com/en/latest/userguide/locations.html

…so, the documentation for the standard flatbed alert system is also still missing. Would be appreciated if you could cover both.

frogonwheels:
Will do, and I'll add them both to the ui if that is ok.

Michael

PR #852, opened 2023-03-11
Web I/F - Add button on script editor to reload ECU Mappings

frogonwheels:


dexterbg:
Michael,

the button should only be present if the ECU subsystem is actually enabled / in use. To my knowledge, the ECU subsystem is used only by a small fraction of users, and the button will confuse those who don't use it.

Regards,
Michael

frogonwheels:
Yes - ECU Enabled, that would do it. Even if I am using the ECU and it's
not powered on, it only makes sense to reload if the ECU is actually
running. I'll do that.

//.ichael.

PR #853, opened 2023-03-11
Increase ecu responsiveness with 'sub-tick' and events

frogonwheels:
- adds a 'sub-tick' timer to keep the poll moving
 - adds an event for ecu on/off 
 - default to auto increase the 'throttle' value

### DIFF
--- a/vehicle/OVMS.V3/components/obd2ecu/src/obd2ecu.cpp
+++ b/vehicle/OVMS.V3/components/obd2ecu/src/obd2ecu.cpp
@@ -173,10 +173,12 @@ obd2ecu::obd2ecu(const char* name, canbus* can)
   xTaskCreatePinnedToCore(OBD2ECU_task, "OVMS OBDII ECU", 6144, (void*)this, 5, &m_task, CORE(1));
 
   MyCan.RegisterListener(m_rxqueue);
+  NotifyStartup();
   }
 
 obd2ecu::~obd2ecu()
   {
+  NotifyShutdown();
   m_can->SetPowerMode(Off);
   MyCan.DeregisterListener(m_rxqueue);
 
@@ -184,6 +186,16 @@ obd2ecu::~obd2ecu()
   vTaskDelete(m_task);
 
   ClearMap();
+  }
+
+void obd2ecu::NotifyStartup()
+  {
+  MyEvents.SignalEvent("vehicle.ecu.start", NULL);
+  }
+
+void obd2ecu::NotifyShutdown()
+  {
+  MyEvents.SignalEvent("vehicle.ecu.stop", NULL);
   }
 
 void obd2ecu::SetPowerMode(PowerMode powermode)
--- a/vehicle/OVMS.V3/components/obd2ecu/src/obd2ecu.h
+++ b/vehicle/OVMS.V3/components/obd2ecu/src/obd2ecu.h
@@ -94,6 +94,9 @@ class obd2ecu : public pcp, public InternalRamAllocated
     uint32_t m_supported_21_40;  // bitmap of PIDs configured 0x21 through 0x40
 
   public:
+    void NotifyStartup();
+    void NotifyShutdown();
+
     void IncomingFrame(CAN_frame_t* p_frame);
     void LoadMap();
     void ClearMap();
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -281,6 +281,9 @@ OvmsVehicle::OvmsVehicle()
   m_poll_entry = {};
   m_poll_vwtp = {};
   m_poll_ticker = 0;
+
+  m_timer200ms = NULL;
+  m_poll_subticker = subtick_init;
   m_poll_single_rxbuf = NULL;
   m_poll_single_rxerr = 0;
   m_poll_moduleid_sent = 0;
@@ -375,6 +378,10 @@ OvmsVehicle::OvmsVehicle()
 
 OvmsVehicle::~OvmsVehicle()
   {
+  if (m_timer200ms) {
+    xTimerDelete( m_timer200ms, 0);
+    m_timer200ms = NULL;
+  }
   if (m_can1) m_can1->SetPowerMode(Off);
   if (m_can2) m_can2->SetPowerMode(Off);
   if (m_can3) m_can3->SetPowerMode(Off);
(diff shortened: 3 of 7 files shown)


dexterbg:
Michael,

please split PRs into system & vehicle specific parts if possible. Aside from this, I see some substantial issues:

- Major: you're introducing `ECUStatusChange()` as a general implementation changing the poller behaviour in a potentially incompatible or even hazardous way: the framework must not change the poller throttling without explicit grant by the vehicle. The default is one (1) request per second, as some vehicles may have an issue with high frequency polling. Also polling lists can currently be designed to be run at exactly the speed configured by the vehicle / poller client.
- Breaking: the apparently new method `HasPollEntry()` isn't included in the PR.
- Conceptual: the introduction of a "sub tick" conflicts with the overall ticker frequency concept of the poller, dilutes the poll ticker concept and adds complexity. All comments documenting the tick/second coupling are unchanged, so I need to assume you at least didn't review them & potentially didn't take this into account. For example, all poll time intervals are currently given in seconds, but with subticks enabled, will be sent at 5x the defined frequency.

I think what this actually needs is either (a) some kind of overall poll tick frequency control, i.e. `PollSetTickInterval(milliseconds_per_tick)`, or (b) introduction of a fully decoupled polling task with an option to install multiple polling lists with different tick intervals each, for concurrent processing by the poller task.

Option (a) would still lead to poll times being interpreted differently depending on the poller setup, but at least this will be transparently controlled by the client and be consistent -- no ticks and sub ticks, all ticks are still equal.

Option (b) would probably be a better solution, as it would allow to install a standard poll list for per second intervals, and to dynamically extend or replace that by a high frequency polling list on demand. It also solves the issue of single requests needing to temporarily replace the current polling list.

Maybe there are other options.

Regards,
Michael

frogonwheels:
Thanks Michael,

I might cancel this P/R and rethink a bit, especially with your suggestions.

Somebody had suggested that there be some advantage for general vehicles
while ECU was on, so I will leave the ECUStatusChange virtual fn blank but
then  I might actually submit that commit first as a separate P/R.

FYI The subtick didn't  change the primary poll tick frequency (the
frequency of m_poll_ticker increments remains variable with 1 second
granularity) or the throttle (events per second) even, since only
VehicleTicker1 calls the PollerSend(true) that enables the list to restart
- that bit is unchanged, and yes I understood that; no doco change
necessary.

The subtick just means that the poll list is kept moving. Usually that is
done by a successfully complete response (that calls PollerSend(false) )..
So if that does not happen the subtick will keep it moving (by calling
PollerSend(false) if we aren't throttled). A failed op without sub-tick
enabled would cause the next op to be deferred to the next '1 second' tick
delaying the poll-tick increment.  So maybe I agree it adds a little
complexity however it is aimed at stopping hang-ups caused by
PollerSend(false) not being called.
Possibly the sub-tick should have checked there were no outstanding
responses being waited on.

The missing HasPollEntry was careless, sorry..

I actually wonder if the changes I've got coming might have more impact.
I'll try it out without subtick enabled.  (though there is potential that
we could have poll lists that reset on sub-tick rather than just primary
tick),

I do like the idea of (a) tweaking the poll frequency..   And using a value
(set dynamically by the car) like 600ms during ECU,  to drive the main
PollerSend(true) tick frequency instead of using the timer event. We could
still make sure the throttle was only reset every second (rather than every
tick)... keeping the throttle in events per second. This would fit in with
the other change set that I have coming.

Would we just move the call to PollerSend(true) that happens in
VehicleTicker1 to what is currently the 'subtick' call (the NULL
frame.origin event)?  This would mean that all the Poller calls happen in
the polling thread which could be a good thing.

//.ichael

dexterbg:
Michael,

> FYI The subtick didn't  change the primary poll tick frequency (the
frequency of m_poll_ticker increments remains variable with 1 second
granularity) or the throttle (events per second) even, since only
VehicleTicker1 calls the PollerSend(true) that enables the list to restart

`PollerSend(false)` is currently meant to be called only on response reception, to immediately send the next due request if allowed to do so. `PollerSend(false)` finishes with increasing `m_poll_ticker` when it doesn't find any more due requests for that second, assuming it won't be called again until the next actual ticker event. When you call `PollerSend(false)` again now, it will thus begin processing the polls for the next second.

> Would we just move the call to PollerSend(true) that happens in
VehicleTicker1 to what is currently the 'subtick' call (the NULL
frame.origin event)?  This would mean that all the Poller calls happen in
the polling thread which could be a good thing.

Just to be clear: _currently_ there is no dedicated polling thread. Polling is initiated by the event task (ticker) and continued by the vehicle task (CAN RX). Polls can also be initiated from any other context by calling the single poll interface (shells, duktape, web commands, …).

Extending the vehicle task to become the polling initiator is a good option to avoid introducing another task -- these always need fixed stack space. It would also take some load from the event task, which already has to do quite some work including slow script execution.

To move _all_ poller calls into that thread, you need to do a full separation of all control API methods as well (wrapping them into command messages sent to the task). That's very clean, but probably introduces some critical changes -- and may not be necessary as we've already got a concurrent access locking scheme for the poller.

Regards,
Michael

frogonwheels:
> PollerSend(false) is currently meant to be called only on response reception, to immediately send the next due 
> request if allowed to do so. PollerSend(false) finishes with increasing m_poll_ticker when it doesn't find any more due
> requests for that second, assuming it won't be called again until the next actual ticker event. When you call
> PollerSend(false) again now, it will thus begin processing the polls for the next second.

Oh, yeah indeed it does.  I fixed that in the next piece of work. Damn. That would explain extra improved throughput I initially experienced though.  I had intended to maintain the 1 poll tick per second maximum.  My understanding of the whole process has improved since I wrote this original sub-tick commit.

I'm going to try reworking this just to see where it might go.  I'll use the sub-tick timer there to create the 1 second (potentially adjustable) injection into the main RxTask() that calls PollerSend(true) rather than using the Timer1.  I'll make sure that m_poll_sequence_cnt is reset exactly every second regardless.  If we start with that we immediately get most of PollerSend(true) passed onto the RxTask() thread.. with the exception of the blocking polls.  I might also pass in an enum to PollerSend() with values something like   { Main, SubTick, Response, OnceOff } .  I might keep this work-in-progress thread but then rebase it into a new P/R.

frogonwheels:
Re-worked and separated the Secondary Poll tick concept as https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/859 but moved it all to the Rx thread

PR #854, opened 2023-03-12
Add ability/events to respond to ECU start/stop

frogonwheels:
- Adds events for ECU start/stop
 - Adds 'ECUStatusChange' virtual method to vehicle
 - ECU Start/Stop also responds to vehicle start/stop change.

### DIFF
--- a/docs/source/userguide/events.rst
+++ b/docs/source/userguide/events.rst
@@ -173,6 +173,8 @@ vehicle.charge.timermode.off                  Vehicle charge timer mode has been
 vehicle.charge.timermode.on                   Vehicle charge timer mode has been switched on
 vehicle.charge.type                 <type>    Vehicle charge connection type has changed (e.g. ccs/type2/…)
 vehicle.drivemode.<n>                         Vehicle drivemode has been set to profile <n> (vehicle specific)
+vehicle.ecu.start                             Called after the ECU is started.
+vehicle.ecu.stop                              Called before the ECU is stopped.
 vehicle.gear.forward                          Vehicle has been put in forward gear
 vehicle.gear.neutral                          Vehicle has been put in neutral gear
 vehicle.gear.reverse                          Vehicle has been put in reverse gear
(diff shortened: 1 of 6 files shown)


dexterbg:
Michael,

thanks for separating this.

Please change the event naming: `vehicle.*` is meant to be bound to / emitted by the vehicle context. Put the events into the component's already defined namespace, i.e.:
- `obd2ecu.start`
- `obd2ecu.stop`

Also, when introducing component status communication it's normally a good idea to also introduce a status metric along with the status events. Think about a client connecting to the system after the component has already been started -- it won't receive the start event and will not be able to query the status.

Regarding the vehicle framework, please follow the scheme of the `Notify…/Notified…` methods/hooks, and place the declarations and definitions near those. Do not combine multiple status sources into the hook call, reduce this to the obd2ecu status change. Do not attach to the events but use the metric listener, so the hook is called synchronously on the metric change (like all other vehicle hooks).

Please remove the accidental `PollerNextTick()` method declaration inclusion.

Regarding documentation, be clear about using the term "ECU" in a way to avoid confusion: ECU = Engine Control Unit, so users may think this is an actual vehicle/hardware component. In this case it's a virtual ECU, an ECU-like OBD2 communication emulation created by the OVMS, to interface devices expecting standard OBD2 communication (like HUDs).

And please reduce the `changes.txt` entry strictly to user related info, that file is meant as a quick overview of changes from the user's perspective. Users normally don't want to read about virtual class methods ;-)

Regards,
Michael

frogonwheels:
Understood.  I think I got all that.  Would a standard metric "m.obd2ecu.on"  be ok? I would set that in the current place when constructing/deleting the obd2ecu class?
And then I would follow say the 'vehicle.valet.on' approach with both the events ('obd2ecu.enabled') and also a notification 
"info", "obd2ecu.enabled" , "obd2ecu Enabled"  for both enabled and disabled.

dexterbg:
`m.obd2ecu.on` would be ok, but a push notification seems to be too much to me for this -- is there any need to alert the user about the component being activated / deactivated? I think that's normally expected to happen automatically as a background support for some HUD. In that case you can omit the `Notify…` method.

The valet mode push alert originates from that potentially being a state change at the vehicle, by someone else than the OVMS user.

frogonwheels:
Yes, this makes sense. You mentioned Notify methods which is why I just wanted to clarify whether you expected one.  
All good. 
WOuld you prefer me to   1) Make modifications as commits to this p/r  2) Force push a recrafted to this p/r  3) Recraft the whole lot and submit it as a better p/r

frogonwheels:
Redone as 
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/857

PR #857, opened 2023-03-12
Add metric+events for obd2ecu on/off

frogonwheels:
- Adds metric m.obd2ecu.on
 - adds events obd2ecu.start, obd2ecu.stop
 - Add bd2ecu change-notification virtual methods


dexterbg:
Issues remaining:
- The `vehicle.ecu.*` events are still in the user guide.
- The `obd2ecu.*` events documented don't match the actual event names signaled. I suggest keeping the variant documented, i.e. start/stop, as these are actions instead of states. Probably the `Notified…` hooks should also follow that naming scheme.
- Being in the component`s namespace, the events should be signaled by the component, not by the vehicle metric listener. Btw, as you move them, take care to first change the metric, then signal the event. That way, an event listener can simply react to both events the same way by reading the metric value.

frogonwheels:
Thanks. Crappy brain switched between on/off and start/stop.  I had one of those nagging feelings that something wasn't right. I believe that should be all good now.

PR #859, opened 2023-03-17
Move ISO TP poller operations to PollRx Thread with variable time

frogonwheels:
- Add Primary/Secondary ticks
 - Aims to Decrease latency of OBD Poll Loop ( eg for obd2ecu)
 - Always run poller from Rx thread.
 - Allow primary and secondary poll lengths to be varied.
 - Pass in Enum of poll call source
 - Make sure only to increase poll ticker in Primary poll tick


frogonwheels:
Just putting this here so it's not lost and @dexterbg is aware that I'm aware.

In https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/853 @dexterbg wrote:

> To move all poller calls into that thread, you need to do a full separation of all control API methods as well (wrapping them into command messages sent to the task). That's very clean, but probably introduces some critical changes -- and may not be necessary as we've already got a concurrent access locking scheme for the poller.

I think sending messages to the thread to adjust things would be a good idea,  however given the  scheme already in existence because there are multiple threads involved, it's not immediately necessary.

frogonwheels:
The latest commits are quite a change along the lines of separating out the poller implementation and effectively having one thread per bus which then executes all the TX/RX responses as well as the Poll loop and most of the parameter changes.

The critical sections are now limited to the points where a particular poll item is fetched from the poll list for execution.
One reason is the particular cases where the poll list is stored in a dynamic array, specifically setting the list to NULL requires the change to happen before returning to execution.

All other modifications are done by commands to the poll queue.  
Because there are a bunch of member variables that moved from OvmsVehicle to OvmsPoller class, they are no longer available for `IncomingPollReply `which has now been expanded considerably.  I'm aware this might attract some attention.  One of the big changes is that I have changed various implementations that used `m_poll_moduleid_low` to use the actual module received (now being passed in the parameters as moduleid). I'm fairly confident that this is the right call, however there are a few where I'm concerned they might not be.

There is a LOT in this which is why it is now a draft. 

I have a nicer mechanism for handling the poller lists coming up, but I need to move it to this new, separate class.

markwj:
Whilst I appreciate the effort, I am confused where this PR is going.

We already have a vehicle task (handling received CAN bus data). Why do we need another task for polling, and one task per CAN bus? My concern is RAM, as each task requires stack space in precious standard RAM (which is already tight).

This PR now seems to be modifying dozens (if not all) vehicle implementation files. I'm trying to understand the structure, but the changes just seem too extensive now to see clearly what the new arrangement is. In particular, the relationship between the standard vehicle base support class, and the new poller vehicle class - as well as how support for vehicle types that need both functionality (active polling, and passive CAN bus data) is handled.

We seem to have gone from a fairly simple goal (support sub-1second polls) to something massive now, and I can't really see the reason why. Can we take a step back and summarise first what enhancements to the polling system we are trying to make? Then we can see the most non-invasive way of making those enhancements.

frogonwheels:
To be honest, I thought I mentioned before the idea of having a command
thread per bus and that it might be a good idea, and receiving nothing
specific back on that I went with it...
but be that as it may (and I may have had it in my head and neglected to
mention it), I've done what I've done, and  what I _have_  done is:

* Factored out the OBD Polling into a separate class (OVMSPoller).
(Currently each OvmsPoller class is designed to handle one CAN bus - to be
discussed).
* _Replaced_ the Single Vehicle Rx Poller Thread with a Poller Thread per
OVMSPoller instance.  (So  currently 1 bus with polling would mean the same
# of threads as before and yes, 2 or more busses with polling would add  1
extra thread each bus requuring plling).
*  Made ALL OBD Interactions happen on that thread  (This is probably that
main bit you intended - and it took me a while to find all the bits).
    * The per-second timer is queued to the thread  (? Did that used to be
on the TIMER thread?)
    * The Can Tx result is queued to the thread.
    * The Can Rx are all queued to the thread.
    * Most of the adjustments to polling parameters except for changing the
Poll List happen on that thread.
* The interactions with the poller list are now very contained.  Once the
entry is fetched, all interactions are with the copy of the entry (that
mechanism was kind of there already).

The 'extensive' changes a really because
   * some of the members that used to be only 'available' in
IncomingPollReply() were via members on OVMSVehicle now have to be passed
as parameters.
     The main ones that we care about are:
      * m_poll_ml_frame (now passed is as mlframe)  and
      * m_poll_moduleid_low (now passed in as moduleid).
     The changes are not much more than that.  Yes it touches a lot of
modules.. it's just about whether this makes sense. Does passing more
parameters on the stack in IncomingPollReply() affect the stack depth too
much?  It makes for a larger function signature, but imho it makes it
easier to see what is happening.
   * IncomingFrame[1-4]()  now has p_frame as a const parameter (which it
notionally is anyway - modifying p_frame makes no sense).  This is really
not much of a change and has zero impact on functionality. It was only
necessary because I took the address of a const reference to the frame.
   * In order to contain the operations on the poller, and allow us more
control over it, I want to remove all reference to the poll list
variables.  From this we get the Pause/Resume mechanism.

I still see these as positive changes.
 * I think there's lot of good in having the Poller functionality factored,
but it DOES mean that you have to pass the few extra parameters in a
different way (maybe as a single const struct reference?). It seems to make
much more sense to use the actual received moduleid than
m_poll_moduleid_low which seems to be almost-but-not-quite-right.
 * I think the command thread speaks for itself, and is what you intended
me to implement(?) - (just not one-per-bus).

So - back to what we want I guess
  *  A single thread + queue that all the polling and related actions
related to it take place on (replacing the old Rx Queue)
  * Contain all actions relating to the queue behind well-defined
interfaces.
  *  A separate, modifiable, timer that keeps the primary poll_state
ticking over and provides secondary calls to the queue to keep responses
and some retries happening, trying to keep the poll_state ticking over once
per second-ish.
   * A mechanism that allows the poller to keep up-to-date (say 3-4 times
per second) certain key metrics required by OBD2ECU.

If you let me keep the OVMSPoller factoring and associated module changes,
I  could change the OVMSPoller class back to handle all CAN busses and then
just have a single OVMSPoller class instance with its single Thread
(replacing the Rx Thread queue).

Thoughts?

//.ichael

frogonwheels:
One other thought which doesn't help a whole lot with the perceived
complexity but which fixes the whole threading issue:
Still have one OvmsPoller class per CAN bus, but have  a single
thread/queue that handles all the OvmsPoller instances.
This would mean that each bus could respond individually/concurrently to
OBD requests (Is this something you might have suggested previously?),
but do it in a single thread.

Thoughts?

//.ichael

frogonwheels:
Notice I turned this P/R into a draft with that last big change.  I'm happy
to truncate it back to where it was at any point for merging.  The problem
was that I received no feedback on the original thread  on my question
about what the simpler version changed (sent on 8th April - waited > 3
weeks), so I moved forward with the idea of a proper command queue as per
your suggestion at some point that was included in the comments on the P/R.

This last push to the draft p/r fixes the issue with the number of
threads.  We are back to the original number of threads/tasks.

What this whole P/R gives us is
 * Reduce the exposure of internal bits of the Polling List and ISOTP
mechanism to vehicle implementations.
 * Puts all the polling Rx / Tx Response etc onto a single thread
 * Allows individual busses to have commands sent to them concurrently
(processed back through a single thread.. but still) - with the
ISOTP/VWTP protocol states being per-Bus.

//.ichael

frogonwheels:
I can also push up a commit that starts the poller thread with the first CAN bus, thus allowing passive monitoring.  (I had missed that and saw that you mentioned it when I re-read the comment).

markwj:
I think I understand what you are trying to do now, and will review. The issue is that the DIFF patch for this is now 7,232 lines (against master) so will take some time to go through.

frogonwheels:
Thanks Mark,

Understanding how long it will take, if you could let me know of any
concerns you have as you go, I will try and address them straight away.

One outstanding is spinning up the thread as soon as the Can is enabled for
passive listening. Patch is small and in the pipeline.

Michael.

markwj:
I am still working through this. In summary, we seem to have three change sets here:

1. A change from 'CAN_frame_t* p_frame' to 'const CAN_frame_t* p_frame' in vehicle prototypes and a few other places (dragging in a large number of files not related to OBD style polling).
2. A change to the OBDII polling prototype (passing extra parameters).
3. A change to the OBDII polling object model.

In normal circumstances, I would not suggest this, but is it possible to simplify this by breaking it up so it is not so inter-dependent? Perhaps do this in three easier stages:

1. Take out #1 and incorporate that first, trivially.
2. Then perhaps #2 could be postponed (just have #3 use the existing prototype format). That would reduce the extent of the changes, and bring us down to just this #3 needing a more extensive review.
3. Then finally #2 could be brought in as a simple change to prototypes and the call (similar to #1).

Overall, the current approach seems ok to me. But we need a way of simplifying this to make it more reviewable and easier to incorporate. The current patch of 116 files and 7,279 lines is just very hard to review properly. The risk of breaking something in those 72 vehicle support files is just too great.

frogonwheels:
One of the reasons I just went for it with a draft of the full changes was
to just put it out there with the (nearly) end Game. Maybe I should push
the last bits to this draft p/r now that we know it will need some
reworking.

#1 is easy and can be done separately. Agrees

#2 could be  split into:
2a - Prototype change only.
2b - Changes to function bodies making use of the new prototypes.

The problem is that the change #3 requires #2 since the couple of obdii
protocol related members being regularly referred to are not in scope any
more (which is why they are passed as parameters) as they only exist in the
OvmsPoller class.
#2 could be done in the 2 parts and I could make the obd protocol members
private. This would be then a good review point. Thoughts?

There are also the couple of commits that allow me to make the poll list
entry and pointers private. I can push these other changes up so you can
see where that is going.

Michael

markwj:
> #1 is easy and can be done separately.

Please do this first (in a separate PR). It should be relatively simple and trivial to accept. It will remove all the changes to broadcast/passive CAN traffic vehicles that don't use polling from this PR.

We can then see how many are left affected by #2 and #3 and tackle that.

frogonwheels:
On its way.  Trying to separate it all out into manageable chunks... Making
sure they all compile of course.
I'm hoping that seeing the end goal will help.

Michael

PR #861, opened 2023-03-26
Add support for Maxus Euniq 5 6-seat

Jockz0rz:
I have added support for Maxus Euniq 5 6-seat. 
The code is mostly a copy of the code from Maxus eDeliver3 with minor edits. 

Also this is my first time I submit a pull request so I hope I doing everything right.


dexterbg:
Jock,

welcome, very nice first contribution :-)

This is mostly perfect, except:
1. You must not include generated files, in this case `charts.js.gz` in a PR, without the corresponding source file change that caused the change. I assume you actually didn't change the `charts.js` file, but accidentally got a slightly different `.gz` from the source by means of your build tools. You need to remove the `.gz` file from the PR / replace your file by the original version.
2. You've kept the eDeliver config/metrics namespace `xmg` for the Euniq. You need to introduce a unique namespace for each vehicle, to keep config parameters and metrics distinct. Otherwise a user switching between these two vehicles (maybe because she/he uses the same OVMS unit for both) may run into conflicts, because configuration values and metrics have a different meaning for each. While that may not be the case yet, vehicles may evolve into different directions. You could maybe use `xme` or `xmq`, but I haven't verified these are unused.

Regards,
Michael

Jockz0rz:
Hi Michael,
Thank you for the input :-)
I think I have done everything according to you comments now. 
But as I said it is my first time using git and contributed to a opensource project so any input is really appreciated.

dexterbg:
When you're ready to announce the new vehicle support to all users, add an entry to `changes.txt` and a link to the vehicle in the `README.md`. It's also recommended to update the sdkconfig defaults in `support`.

Regards,
Michael

PR #867, opened 2023-04-09
WIP: Implement Server V3 metric transmit include/exclude filter

### LINKED ISSUE #344: GSM Data Saver V3 Server MQTT: post only selected metrics?

ericyeargan:
This PR implements filtering of metrics to be transmitted by Server V3.

The system is configured with the `metrics.include` and `metrics.exclude` config instances under `server.v3` (currently by OVMS console only). If those instances are undefined or empty strings, the current behavior should be preserved - all metrics will be transmitted. Otherwise, both items should contain comma metric names. Names may be prefixed or suffixed by '*' to indicate a wildcard match.

Both include _and_ exclude lists may be provided to allow broad inclusion of metrics by wildcard and subsequent, more specific exclusion (See example below).

**Examples**

Only send the `s.v3.connected` and `v.b.soc` metrics:
```
OVMS# config set server.v3 metrics.include s.v3.connected,v.b.soc
OVMS# config rm server.v3 metrics.exclude
```

Send all metrics _except_ v.b.soc:
```
OVMS# config rm server.v3 metrics.include
OVMS# config set server.v3 metrics.exclude v.b.soc
```

Send all v.b metrics except for v.b.soc:
```
OVMS# config set server.v3 metrics.include v.b.*
OVMS# config set server.v3 metrics.exclude v.b.soc
```

Any feedback on design and/or implementation are welcome!

Resolves #344

### DIFF
--- a/vehicle/OVMS.V3/components/id_filter/CMakeLists.txt
+++ b/vehicle/OVMS.V3/components/id_filter/CMakeLists.txt
@@ -1,0 +1,4 @@
+idf_component_register(SRCS "src/id_filter.cpp src/id_include_exclude_filter.cpp"
+                       INCLUDE_DIRS src
+                       PRIV_REQUIRES "main"
+                       WHOLE_ARCHIVE)
--- a/vehicle/OVMS.V3/components/id_filter/component.mk
+++ b/vehicle/OVMS.V3/components/id_filter/component.mk
@@ -1,0 +1,12 @@
+#
+# Main component makefile.
+#
+# This Makefile can be left empty. By default, it will take the sources in the
+# src/ directory, compile them and link them into lib(subdirectory_name).a
+# in the build directory. This behaviour is entirely configurable,
+# please read the ESP-IDF documents if you need to do this.
+#
+
+COMPONENT_SRCDIRS := src
+COMPONENT_ADD_INCLUDEDIRS := src
+COMPONENT_ADD_LDFLAGS = -Wl,--whole-archive -l$(COMPONENT_NAME) -Wl,--no-whole-archive
--- a/vehicle/OVMS.V3/components/id_filter/src/id_filter.cpp
+++ b/vehicle/OVMS.V3/components/id_filter/src/id_filter.cpp
@@ -1,0 +1,135 @@
+/*
+;    Project:       Open Vehicle Monitor System
+;    Date:          8th April 2023
+;
+;    Changes:
+;    1.0  Initial release
+;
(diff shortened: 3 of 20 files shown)


dexterbg:
Eric, welcome & thanks for the submission!

From reviewing your code I don't see anything technically wrong, maybe except using `std::regex`, which probably adds a lot to the overall firmware size (didn't try).

From the perspective of system design and usability though, I would rather like to see you applying the same filter scheme & implementation @llange Ludovic Lange has already implemented for the CAN logging. Ludovic has designed the filter scheme in a very similar way, but his version already has basic wildcard support (for the most relevant prefix/suffix "*" wildcard type). General regex support isn't needed here, in fact would be an issue for most normal users.

See:
- [User Guide: CAN Logging Events & Metrics](https://docs.openvehicles.com/en/latest/crtd/can_logging.html#logging-events-and-metrics)
- [Pull Request #778](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/778)

Reusing / refactoring his filter implementation should be straight forward. Regarding the `blacklist`/`whitelist` scheme, that's probably more user friendly (easier to read & maintain) than introducing a negation syntax, so that's OK for me.

Btw, you can also easily copy & adapt Ludovic' user guide parts for the server v3 configuration user guide section. Please also add an entry to the `changes.txt` history, which is the brief change history for users.

Regards,
Michael

ericyeargan:
Thanks @dexterbg for the pointers. I'm going to close this PR and create a fresh one.

PR #872, opened 2023-04-14
Implement Server v3 metrics filtering

### LINKED ISSUE #344: GSM Data Saver V3 Server MQTT: post only selected metrics?

ericyeargan:
Implement Server v3 metrics filtering with configurable include/exclude list.  See "OVMS Server v3" section of docs/source/userguide/components.rst for details.

A new id_filter component has been factored out of components/can/canlog for use by canlog and the new server v3 metrics filtering functionality.

(see PR #867 for an earlier attempt)

Resolves #344

### DIFF
--- a/docs/source/userguide/components.rst
+++ b/docs/source/userguide/components.rst
@@ -98,6 +98,31 @@ OVMS Server v3
 
 The OVMS Server v3 protocol is MQTT. This is an industry standard protocol, and means an OVMS v3 module can communicate with any standard MQTT server. While this is the future of OVMS, support for this is experimental at the moment and production users should use OVMS Server v2 protocol.
 
+By default, OMVS Server v3 transmits all metrics. This can be adjusted by modifying the ``server.v3`` ``metrics.include`` and ``metrics.exclude`` configuration parameters.
+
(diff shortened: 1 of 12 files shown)


dexterbg:
Perfect :-)

PR #873, opened 2023-04-14
Network: New 'network ping' command to ping (ICMP) hostname or IP address (ESP-IDFv4+ only)

llange:
_Note: This PR won't add any visible feature to current builds, as the ping/ICMP support is only available since ESP-IDFv4.
So the feature will not be enabled for the current (ESP-IDFv3.3.4) builds._

Just like a simplified `ping` command, it will send 4 ICMP echo packets and wait for the reply.

It's not compiled by default, you'll need to enable it in `menuconfig` / `Developer Options` / `Enable netmanager ping support`.

It's a straightforward (and simplified, no handling of any optional arguments) port of ESP-IDF's ICMP echo example: [protocols/icmp_echo](https://github.com/espressif/esp-idf/tree/f404fe96b1/examples/protocols/icmp_echo)

Note: an ugly hack has been introduced (a 4200 ms delay) to prevent the `OVMS#` prompt to interfere with the output.

Without this delay, the output looks like this:
```
OVMS# network ping 127.0.0.1
64 bytes from 127.0.0.1 icmp_seq=1 ttl=-1 time=0 ms
OVMS# 64 bytes from 127.0.0.1 icmp_seq=2 ttl=-1 time=0 ms
64 bytes from 127.0.0.1 icmp_seq=3 ttl=-1 time=0 ms
64 bytes from 127.0.0.1 icmp_seq=4 ttl=-1 time=0 ms

--- 127.0.0.1 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 0ms
```

and with the delay it's:
```
OVMS# network ping 127.0.0.1      
64 bytes from 127.0.0.1 icmp_seq=1 ttl=-1 time=0 ms
64 bytes from 127.0.0.1 icmp_seq=2 ttl=-1 time=0 ms
64 bytes from 127.0.0.1 icmp_seq=3 ttl=-1 time=0 ms
64 bytes from 127.0.0.1 icmp_seq=4 ttl=-1 time=0 ms

--- 127.0.0.1 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 0ms
OVMS# 
```

Let me know the proper way to prevent this.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -97,6 +97,7 @@ Open Vehicle Monitor System v3 - Change log
     obd2ecu.start                     -- Called after the OBD2ECU process is started.
     obd2ecu.stop                      -- Called before the OBD2ECU process is stopped.
 - Web UI: Add configuration for Valet and Flatbed geofence to the Locations config page.
+- Network: New 'network ping' command to ping (ICMP) hostname or IP address. (ESP-IDFv4+ only / needs to be enabled in menuconfig - Developer Options)
 
 2022-09-01 MWJ  3.3.003  OTA release
 - Toyota RAV4 EV: Initial support added. Only the Tesla bus is decoded and just listening so far.
--- a/vehicle/OVMS.V3/main/Kconfig
+++ b/vehicle/OVMS.V3/main/Kconfig
@@ -875,4 +875,11 @@ config OVMS_DEV_DEBUGNOTIFICATIONS
     help
         Enable to show notifications raised
 
+config OVMS_DEV_NETMANAGER_PING
+    bool "Enable netmanager ping support"
+    default n
+    depends on OVMS && LWIP_ICMP
+    help
+        Enable to add 'network ping' command
+
 endmenu # Developer Options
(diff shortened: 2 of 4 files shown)


dexterbg:
Ludovic,

the proper way to prevent the prompt is to not return from the command until it's done. With subtasks / callbacks you normally do this by waiting on a semaphore, pattern:

```
  OvmsSemaphore taskdone;
  // start task, pass &taskdone to the task, do taskdone.Give() on task finished
  taskdone.Take();
```
(See `OvmsConfig::Restore()` or `OvmsVehicle::PollSingleRequest()` for examples, the latter also has a timeout)

In the case of a ping command, maybe there's an even better option. I know the MS ping command defaults to 4 pings, the Unix counterpart can do a specifyable number of pings and by default runs until stopped by the user (Ctrl-C). If you intend to support that, you can use the `OvmsWriter::RegisterInsertCallback()` to listen to user input while waiting for the ping task to finish.

For an example on how to do that, see my utility class `OvmsCommandTask`. I implemented that for the `vfs tail` command. Another task is not needed in this case, but the class shows how to use an insert callback to detect Ctrl-C, and how to detect if the execution mode is interactive.

Regards,
Michael

llange:
Michael,

Thanks for the help. I took the easy way with the semaphore. At the moment I'm not implementing a fully linux/unix compatible ping, just a debugging tool for which I think a count of 4 pings should be enough.   
Were I wrong about it, the [ESP-IDF sample program](https://github.com/espressif/esp-idf/tree/f404fe96b1/examples/protocols/icmp_echo) from which I sourced the code has handling for many more options, and we could certainly implement a few of those, in particular the "count" of ping sent.  
If you agree, we'll try to integrate this poor man's version of ping first, then if there is enough demand we certainly can improve it.

Issue #752, opened 2022-10-23
Explore possibility of wireguard (VPN) support


llange:
I was toying with the idea of having a network-enabled OVMS auto-registering in a dedicated network ; and being "locally" available, mdns working, web and ssh reachable... while exposing no local service on its main interface.

I've seen discussion around a "firewall", and with this approach there is no more need to have one if no services are exposed.

Don't know if it's feasible, a project like https://github.com/smartalock/wireguard-lwip is certainly interesting to look at (and also https://github.com/ciniml/WireGuard-ESP32-Arduino).

This issue is to discuss, gather interest and coordinate ideas.

llange:
some notes: I tried as a first step to integrate https://github.com/trombik/esp_wireguard which is an ESP-IDF component, but unfortunately our version of lwip seems to old for that (missing an udp function etc...)
Maybe #360 could help here.

werdnum:
Would love to see this! Actually what I'd *really* love to see is Tailscale (based on Wireguard), but standalone Wireguard I could live with too.

llange:
Hi @werdnum,
if you're feeling adventurous and not afraid of having your hands dirty compiling, please note that on this branch https://github.com/llange/Open-Vehicle-Monitoring-System-3/tree/752-wireguard you should be able to build an ESP-IDFv5 version of OVMSv3 with my Wireguard patches.
Once you manage to compile it, Wireguard works out of the box - I'm only testing SSH and Web server on it but no reason that other protocol do not work.
Let us know.

werdnum:
> Hi @werdnum, if you're feeling adventurous and not afraid of having your hands dirty compiling, please note that on this branch https://github.com/llange/Open-Vehicle-Monitoring-System-3/tree/752-wireguard you should be able to build an ESP-IDFv5 version of OVMSv3 with my Wireguard patches. Once you manage to compile it, Wireguard works out of the box - I'm only testing SSH and Web server on it but no reason that other protocol do not work. Let us know.

This looks really interesting, thanks for sharing. It looks like this branch is mostly about getting ESPIDF updated, which I understand the importance of. I'd be keen to try it out, but not 100% clear on what kind of a state things are in right now. I see a bunch of TODOs - most of which don't bother me as I'm not actively hacking on OVMS - but I do see there might be some TLS/SSH issues remaining.

[...]

PR #882, opened 2023-04-24
networking: add support for WireGuard VPN

llange:
⚠️ **Not ready for merging - for discussion only**

ℹ️ _This PR is based on other branches that are needed, and not merged in master yet - so it appears huge. In particular it's based on my [experimental-esp-idf-build-workflow](https://github.com/llange/Open-Vehicle-Monitoring-System-3/tree/experimental-esp-idf-build-workflow) branch. However, only the last commit `networking: add support for WireGuard VPN` is the important part of this PR._

This VPN allows your OVMS module to "connect" itself to your own private network, make it visible and reachable, whatever the connexion it is using (WiFi, Mobile, ...)
(Note: 3G untested at the moment)

It should also (unverified) be able to roam from one connexion to the other, or even among different WiFi networks, without any specific setting to do.

A documentation has been added. (Note: The schematics, drawn with [draw.io](https://get.diagrams.net/) is embedded in the resulting PNG file, so you can open the PNG file with the editor to edit the diagram. Please keep it that way (by checking the 'Include a copy of my diagram' check box))

Support for wireguard on ESP provided by https://github.com/trombik/esp_wireguard, cf https://github.com/trombik/esp_wireguard/blob/main/LICENSE

Cf #752


fowi4hjte:
@llange 
This would be great to have in OVMS. It can also solve some basic security concerns when always enabled.

PR #887, opened 2023-04-28
convert wolfssl to a git submodule

llange:
wolfssl has been forked (https://github.com/openvehicles/wolfssl) and the subdirectory converted to a Git submodule.
Additionally, it has been put one directory level below (in vehicle/OVMS.V3/components/wolfssl/wolfssl) to better separate the upstream code from our build glue (component.mk, CMakeLists.txt and configuration files)

This way it's easier to track any changes made to the module, upgrade when upstram changes, or propose patches to upstream.

We kept the existing version (v4.7.0-stable) - it will be upgraded later


llange:
Depends on openvehicles/wolfssl#1 and openvehicles/wolfssl#2 so that we can reference the [tip of the (patched) 4.7.0 branch](https://github.com/openvehicles/wolfssl/tree/v4.7.0-stable-ovms)

llange:
Review and merge is now possible.

neomilium (approved):

neomilium:
It was easy to review code but hard to know if it works as expected. I rely on @llange 's own build and tests.

markwj:
This seems ok to me. Apart from some parts of it used by wolfssh, we don't seem to use this module at all. So long as ssh connections can still be established to the box, I think this is ok.

markwj (approved):
This seems ok to me. Apart from some parts of it used by wolfssh, we don't seem to use this module at all. So long as ssh connections can still be established to the box, I think this is ok.

PR #890, opened 2023-05-01
Upgrade wolfssl 5.3.0

llange:
WolfSSL can be updated to v5.3.0-stable with a few changes to the build recipes.

It's not the latest version (which would be v5.6.0-stable) but I experienced crashes in the SSH console so I'll stay here at the moment while trying to debug those crashes.


fowi4hjte:
Maybe change Wolfssl version to current 5.7.0 that have fixed some critical security issues. https://github.com/wolfSSL/wolfssl/releases

Is there a reason why this project use Wolfssl and not mbedtls?

Issue #724, opened 2022-08-29
MG ZS EV (2020) SOC and estimated km wrong after BMS update


pascalsaul:
During the maintenance they updated the software versions as well as the BMS to 1027630EU4 which adds delayed charging and battery heating. Very nice but OVMS can't handle the changes all though the "Updated BMS Firmware" is checked and gave OVMS a reboot. The latest 3.3.002-46-g89963cee/ota_0/edge (build idf v3.3.4-848-g1ff5e24b1 Aug 26 2022 21:55:38) does give the same issues. MG changed the max voltage to be 449.75 instead of 456.25 and SOC changed to 98.9071 which is the new 100%. 

```
Before:
v.b.range.est                            255.5km
v.b.range.full                           263km
v.b.range.ideal                          263km
v.b.range.speed
v.b.soc                                  100%
v.b.soh                                  97.18%
v.b.temp                                 28°C
v.b.voltage                              456.25V

After:
v.b.range.est                            17km
v.b.range.full                           263km
v.b.range.ideal                          260.126km
v.b.range.speed
v.b.soc                                  98.9071%
v.b.soh                                  97.07%
v.b.temp                                 27°C
v.b.voltage                              449.75V

Softver:
DCDC F01R0A7355
ATC REL_DC_BEV_C036     
GWM 0026011080
BCM 2210 4176 
PEPS ZS11SW01
SCU 194136858 C0807
PLC 1770099611
ABS 1019000000
EPS Zs11ESVf18
SRM ZS11.E1.02
EPB       S001
FVCM Z5.0RC07  
TC F01R0A7354
TBOX SNMX007Z1E
FDR ACC_SW:5.0
VCU   blrv0450
RDRA 00000SW026
IPK S200425-09
ICE S160630.01
BMS 1027630EU4
TPMS QY1105A812
```

mossc001:
@pascalsaul , I've just had the BMS update on my MG ZS 2020 and have exactly the same problem with the estimated range, it just stays at 18km regardless of power cycles or distance. The "BMS Update" box in the features seems to have no affect; odd.

I haven't noticed a big change in SOC, but the estimated range is the very annoying one.

mossc001:
@pascalsaul , I currently have all the data logging to my own MQTT server, then into Home Assistant, so I might build my own estimation calculation until it's fixed.

GreyPeter:
[...]

PR #898, opened 2023-05-12
Trip consumption

GreyPeter:
Added xmg.p.trip.consumption averaged over 5 minutes for the current trip.
Added xmg.p.avg.consumption  an average of consumption over 5 trips.
Estimated range is now calculated from xmg.p.avg.consumption weighted by xmg.p.trip.consumption.
Add extra firmware selections in MG/Features menu as some vehicles with updated firmware not showing 100% on full charge. Adjusted SOC calculations to reflect the new firmware selection.
Fixes issue #724


dexterbg:
Peter,

it must have slipped my attention previously, that you register car specific shell commands as top level commands.

That needs to be changed, especially with command names like `auth` that can easily get in conflict with future top level system commands.

As with metrics & configs, car specific commands also shall be put in their respective namespace, i.e. `xmg`. See other vehicles on the scheme: register `xmg` as a top level command, then register all MG commands as sub commands for that entry.

You'll probably need some migration policy, as users will need to adjust scripts etc., so I'll merge this PR now, as it doesn't add new top level commands.

Regards,
Michael

PR #901, opened 2023-05-13
ESP-IDF v4+: we do not have uxMutexesHeld any more

llange:
In ESP-IDF 3.3.x builds of OVMS, we're using a fork of ESP-IDF. In this fork, FreeRTOS has been patched with openvehicles/esp-idf@95e43fc2c4360f8126a0985bf037a293bebe3767 to add a `uxMutexesHeld` field in `TaskStatus_t`.

However, in the ESP-IDF >= 4 build we're using mainstream ESP-IDF (for the moment), which does not include this field.

So we adapt the display to handle both cases.


markwj (approved):
We lose some functionality here, but it is minor and the benefit of not having to patch the IDF outweighs the loss.

PR #902, opened 2023-05-14
ESP-IDF4+ : crash handler and boot status

llange:
* error handling rework (`xt_set_error_handler_callback` is only available in our ESP-IDF 3.3.4 fork.)
* `esp_task_wdt_get_trigger_tasknames()` not available (**feature removed** from ESP-IDF4+ builds) TODO
* first steps to support other archs than XTensa in `boot_data.crash_data`


markwj (commented):
Changes seem reasonable. Again, lost functionality, but I don't see we have much choice.

markwj (approved):

PR #904, opened 2023-05-19
OVMS Vehicle - Add 'const' to CAN_frame_t and data parameters

frogonwheels:
- These parameters values are both conceptually const anyway and this helps out with reworking.

### DIFF
--- a/vehicle/OVMS.V3/components/dbc/src/dbc.cpp
+++ b/vehicle/OVMS.V3/components/dbc/src/dbc.cpp
@@ -52,7 +52,7 @@ static const char *TAG = "dbc";
 // Helper functions
 
 static inline uint64_t
-dbc_extract_bits(uint8_t *candata, unsigned int bpos, unsigned int align, unsigned int shifter, unsigned int pos)
+dbc_extract_bits(const uint8_t *candata, unsigned int bpos, unsigned int align, unsigned int shifter, unsigned int pos)
   {
   uint64_t val = (uint64_t)candata[bpos/8];
   unsigned int mask = (1 << shifter) - 1;
@@ -60,7 +60,7 @@ dbc_extract_bits(uint8_t *candata, unsigned int bpos, unsigned int align, unsign
   }
 
 static uint64_t
-dbc_extract_bits_little_endian(uint8_t *candata, unsigned int bpos, unsigned int bits)
+dbc_extract_bits_little_endian(const uint8_t *candata, unsigned int bpos, unsigned int bits)
   {
   unsigned int pos, aligner, shifter;
   uint64_t val = 0;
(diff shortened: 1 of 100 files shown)


dexterbg:
Well, conceptionally these are not const, as CAN frames are passed by value; each listener queue transports a full copy of the frame. The only API transporting a pointer to the actual frame within the CAN is the `CanFrameCallback` issued via `can::ExecuteCallbacks()`, which is const already.

Working on the frames received can in theory be used to pass a frame through some processing pipeline applying modifications / data fixes / translations to the frame before passing it to the final decoder, or it can be used to construct a TX from a modified frame without the need to create a copy first.

But apparently noone has made use of the mutable nature of CAN frames received yet (or at least not in public code?), so we _could_ change the API to const. My question is, what do we gain from this?

> this helps out with reworking

Can you elaborate on this?

frogonwheels:
Hmm ok.

I was considering it from the perspective that the top level value is a
temporary copy of the original value anyway (or it will be when it comes
from the queue) and so to me it made sense that it was const all the way
down.  I had checked that nobody modified the value and so that made sense
to me.

I had definitely not considered that an order-sensitive operation might
modify stuff on the way through.

I'm on a plane at the moment so I'll get back to you on whether there was
any other specific reason other than that though I think I pass the queue
struct through as a const& to a consumer function and was then passing  the
frame from that (a union) thus making it const.

If you really think that you want to be able to munge the frame on the way
through I'll  remove const the other way.

Michael Geddes

frogonwheels:
I worry that if there are modifications along the way of the frame data
that it would start to get hard to maintain... But if you really think I am
wrong in my assertion that this is conceptually const and should be
actually const then I will manage and just carry that up a level.

Michael Geddes

dexterbg:
A clean implementation of a processing pipeline is in the responsibility of the actual vehicle sub class.

It boils down to: is there an advantage for the vehicle base class framework in restricting vehicle sub classes to immutable frames, that outweighs the advantage of light weight frame processing?

The only application I can think of would be, if the framework at some time in the future needs to add frame processing _after_ passing a frame to the actual "incoming…" handler method, that depends on retaining the original frame.

The current model is to do all framework/standard processing _before_ passing the frame to the "incoming…" handlers, which are meant to be the final step (no result returned to the framework). That applies to both direct & poller handlers.

I haven't checked your latest code yet -- do you need/intend to add post processing for the poller rework?

frogonwheels:
Nicely summed up.

The future work does stuff before hand and so I think you are right in that
the const does not really add any value and could force unnecessary copies.

This all started passing a const ref to the command pipe and so I'm better
off passing a normal reference or pointer.

Michael

frogonwheels:
I'm satisfied that this commit would take away from rather than add to potential efficiencies.

PR #914, opened 2023-06-17
Use state record as parameters for Incoming poll

frogonwheels:
This makes protocol state information available to IncomingPoll / Error calls without accessing it as members of OvmsVehicle.  This will allow us to separate that state information into a separate class.


frogonwheels:
Note that some structs have been added into OvmsPoller namespace.  OvmsPoller will become a class, and this avoids having to change a whole bunch of files (again) in that commit..

dexterbg:
Michael,

sorry for the late review.

To me this looks like a change that's only half done.

Why the separation of state & poll entry, and why keep the canbus reference as a separate argument? Also, why keep the third callback API `IncomingPollTxCallback()` in the old style?

A clean API redesign would IMO look like this:
- define a `poll_job_t` class/struct that contains all data about the current poll job (definition & state including bus it's being executed on)
- redefine the application callbacks to get only a reference to the job + callback specific data
- move all current separate `m_poll_*` members into an instance of that class, so we don't need to define temporary objects just for the callbacks

I.e. new callback API:
- `IncomingPollReply(poll_job_t &job, uint8_t* data, uint8_t length)`
- `IncomingPollError(poll_job_t &job, uint16_t code)`
- `IncomingPollTxCallback(poll_job_t &job, bool success)`

That also accomodates perfectly to upgrading the whole poller to an actual job/worker service scheme.

Or do I miss something?

Regards,
Michael

frogonwheels:
I mean this is definitely a stepping stone change. The idea to use
poll_state_t came quite late in the evolution of the whole change set which
has undergone many. drastic changes.  I was unsure of how much to include
in it, and decided to keep the bus and pollentry separate.

In the end they all end up in the separate poller class that handles each
bus state separately, so I guess I can merge them. Certainly putting the
pollentry inside poll_job_t would be better.

The scheme that I have implemented for managing poll 'jobs' is really more
controlled by heart-beat message (with primary/secondary ticks) sent to the
polling queue that then manages the 'next job' process through a list of
Poller Entry classes.   The poll_job_t would certainly still work with this
concept.

I don't know why TxCallback didn't change.. I guess it was because of the
way it was used that I kept it like that. I was trying for palatable
changes.

Would you be ok with me keeping those job entries const?  I really don't
thing allowing the state to be messed with works.

   - void IncomingPollReply(const poll_job_t &job, uint8_t* data, uint8_t
   length)
   - void IncomingPollError(const poll_job_t &job, uint16_t code)
   - void IncomingPollTxCallback(const poll_job_t &job, bool success)

I was going to delay moving the pol_job_t struct into members until the
next stage (I had been considering this but was going to do it after the
whole poller re-work).  I can do this now, but it's going to mean another
major rework of the rest of the commits.  Still - i've done it before so i
can do it again.

//.ichael

frogonwheels:
Ok,  Well for the moment I have rebuilt with the following signatures and I
do feel it's an improvement.  I have maintained the namespace (which will
become a class in a later commit if it all goes to plan). I will hold off
on pushing it up until I can make sure it still runs. (My car is a couple
of thousand km away - will be back for the weekend)

    virtual void IncomingPollReply(const OvmsPoller::poll_job_t &job,
uint8_t* data, uint8_t length);
    virtual void IncomingPollError(const OvmsPoller::poll_job_t &job,
uint16_t code);
    virtual void IncomingPollTxCallback(const OvmsPoller::poll_job_t &job,
bool success);


The final does not really have jobs sent as poll_job_t like things.  It
more maintains a list of job generators with  the main queue being
processed by the poller thread becoming more of a 'command' queue.

//.ichael

frogonwheels:
It runs.

dexterbg:
Looks much better.

The general idea of a "poll job" here is, it should allow concurrent processing of multiple instances, even if you don't intend to support that now. Think for example of a large binary upload (e.g. a firmware/config update for some device) running on bus2 while bus1 keeps polling state metrics. All job specific state info needs to be part of the job container to allow such concurrency.

So for this initial step I think you should also move the TX related state variables into the job container, i.e.

```
    const uint8_t*    m_poll_tx_data;         // Payload data for multi frame request
    uint16_t          m_poll_tx_remain;       // Payload bytes remaining for multi frame request
    uint16_t          m_poll_tx_offset;       // Payload offset of multi frame request
    uint16_t          m_poll_tx_frame;        // Frame number for multi frame request
…
    uint32_t          m_poll_txmsgid;         // Poller last TX CAN ID (frame MsgID)
```

These keep track of a multi frame request transmission and also are relevant for a Tx callback to see where a failure occurred.

It's certainly not easy to introduce this separation now, as the poller is quite old and wasn't designed that way in the first place. Your work on this is appreciated.

Btw, the VWTP channel state `m_poll_vwtp` is a bit of a chimera here as it binds to some job parameters, but is more worker/processor state than job state, as polls can reuse an open channel. In theory, VWTP could also allow concurrent channels, but as far as I remember not on the same bus, so that could become part of e.g. a bus specific worker instance.

frogonwheels:
Thanks,

In the current iteration of the OvmsPoller that I have been working on, I
have the single poller thread but all the state info and processing is done
in a separate class per can bus allowing exactly such streamed operations.
So I would be more comfortable leaving the tx state info where it is as
that will be moved into the OVMSpoller class.

Further down the road I've also much improved the 'blocking' calls , and
made it possible to queue one-off calls (like VIN fetch ) without it
blocking.

Maybe if we can leave it like it is with the TX and discuss it further when
it comes to the poller separation pull requests?

We should be able to leave the VW stuff as it is for the time being, and
maybe down the track discuss tidying it up a bit.

Michael

PR #920, opened 2023-06-26
Add digits of precision switch for efficiency

### LINKED ISSUE #919: Trip report should have more digits of precision for mi/kWh
Currently, the trip report only reports (0) digits of precision. It should show more for mi/kWh.

I would propose; Wh/km = 0, Wh/mi = 0, Kwh/100km = 1, Km/kwH = 2, mi/kwH = 2

kezarjg:
Resolves #919.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -1331,10 +1331,18 @@ OvmsVehicle::vehicle_command_t OvmsVehicle::CommandStatTrip(int verbosity, OvmsW
     ;
   if (wh_per_km != 0)
     {
-    buf
-      << "\nEnergy "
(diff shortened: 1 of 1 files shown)


frogonwheels:
I wonder if this could be a more general idea where we could have a function returning the 'default display precision' for a given unit.   This could be then used in either stdout operations like this or  `%.*f` format specifiers.

dexterbg:
Thanks for contributing!

I've got two minor objections:
- Please do not mix code styles within a module/file (here: bracket/block style).
- Please define conditional blocks with a default/fallback (else) branch when possible, especially when testing for symbol sets that may extend in the future. In this case, it would be natural to define precision 0 as the default case.

Regards,
Michael

kezarjg:
Michael, Thanks for the feedback. I made updates from your two points.

PR #927, opened 2023-08-01
Disable continuous metric streaming when vehicle is off

art103:
When enabling the new location streaming option, OVMS continuously sends modified metric updates even when the vehicles is off. 
For a Nissan Leaf, that is 4-6 metrics every second, which quickly uses up LTE data.

Please find a change that checks the caron value, m_streaming value and takes the minimum next update time based on the other state values.
The change also makes the nested (cond)?action1:action2 code more readable.


dexterbg:
Richard, thanks for contributing!

While this change has the potential to break the streaming feature for users that want to stream DC charging metrics, I agree that streaming is generally meant to be used while driving. Also, more detailed transmission control could be added using the `IdIncludeExcludeFilter` utility class if needed.

PR #937, opened 2023-10-08
Fix bms dod limits

GreyPeter:
BMS DoD limits were not being updated when Config changed in MG ZS EV (UK/EU). Added code to fix this.


GreyPeter:
Can you please reverse this branch. I need to add changes for the MG5 and MGEVZS2.

PR #943, opened 2023-10-22
OVMS Unit Conversion - Add breaks to switches

frogonwheels:
There were a number of 'breaks' missing from switch() statements in  the main UnitConvert functions. Mostly this would have been harmless enough but there were probably some strange edge cases introduced.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_metrics.cpp
+++ b/vehicle/OVMS.V3/main/ovms_metrics.cpp
@@ -2363,6 +2363,7 @@ int UnitConvert(metric_unit_t from, metric_unit_t to, int value)
         case MegaJoules:  return (value*3.6);
         default: break;
         }
+      break;
     case WattHours:
       switch (to)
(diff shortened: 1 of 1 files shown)


vehicle/OVMS.V3/main/ovms_metrics.cpp
```diff
@@ -2812,7 +2829,9 @@ float UnitConvert(metric_unit_t from, metric_unit_t to, float value)
         case MetersPS: return mi_to_km(value * 1000 ) / feet_per_mile;
         default: break;
         }
+      break;
     case dbm:
+      break;
```
dexterbg, line 2834: This doesn't look correct.

dexterbg (changes requested):

frogonwheels:
It certainly doesn’t. Surprised I didn’t get a warning on that one. 

Pushed a fix.

//.ichael

PR #945, opened 2023-10-22
Metrics - Clarify which virtual methods are overridden

frogonwheels:
- Some overrides were marked virtual
 - Mark all overridden methods with 'override'

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_metrics.h
+++ b/vehicle/OVMS.V3/main/ovms_metrics.h
@@ -1021,7 +1021,7 @@ class MetricCallbackEntry
   {
   public:
     MetricCallbackEntry(std::string caller, MetricCallback callback);
-    virtual ~MetricCallbackEntry();
+    ~MetricCallbackEntry();
 
   public:
     std::string m_caller;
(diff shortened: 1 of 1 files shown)


frogonwheels:
This P/R is a tidy-up only and is about clarity rather than any fixes.

dexterbg:
Michael,

if we go that path, shouldn't we also tag the destructors as `override`?

Apparently, the C++ core guidelines once suggested not doing so, but the editors have changed their mind:

- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override
- https://github.com/isocpp/CppCoreGuidelines/pull/1448
- https://github.com/isocpp/CppCoreGuidelines/issues/721#issuecomment-598878968

Regards,
Michael

frogonwheels:
Assuming they are virtual destructors in the base then absolutely I think
it's  a good idea.


Michael

frogonwheels:
I have another interesting and related observation:

Virtual destructors should only be used if a class would be destroyed via a
base class.  This absolutely applies to OvmsMetric  but I don't think it
should apply to OvmsMetrics (which is a singleton) or MetricCallbackEntry.
We are unnecessarily creating a vtable for the class just for the
destructor!

Are you OK if I remove 'virtual' from the destructors of these 2?  In the
same P/R? I can do a separate P/R if needed - I'll definitely do a
separate commit anyway.

//.ichael

dexterbg:
Perfectly OK to add here, those are most probably copy & paste errors.

PR #946, opened 2023-10-22
Metrics - add Date metrics with local/utc printing

frogonwheels:
- Adds UnixEpoch, UTCDate and LocalDate metrics as well as a 'Date' group.
- Sets standard metrics m.time.utc to use UTCDate and v.pos.gpstime to use LocalDate
- Conversion between the different Date metrics is unity - it is simply used for displaying and storing the date value.
- JSON Dates now use the standard(ish) Java date format.


frogonwheels:
One discovery I made while implementing this was that time_t is a 32bit number.  It seems when we move to the 5.0+ framework this will be fixed in the esp32 libraries, but we will still need to implement a separate metric class for 64 bit time_t values.  I'm thinking that an easy way to do it, since the underlying persist mechanism is 32bit, would be to have "_high" suffix a store the high bit as a "separate" metric.  Should I do it with this patchset?  I can implement the separate metric class and put in #ifdefs for when the time_t value is 64bit.

dexterbg:
Having support for 64 bit integer & float (double) metrics is a general advantage, you're welcome to implement this. But as this isn't time specific, I suggest doing it in another PR.

PR #947, opened 2023-10-22
Improve Simcom7600 power on off

drc38:
The modem autodetect power on off functions are a compromise to work with a range of modems. When the modem driver is set to SIM7600 by using the shell config command then this code will be implemented to better match the 7600 hardware design power on off timing https://simcom.ee/documents/SIM7600E/SIM7600%20Series%20Hardware%20Design_V1.03.pdf.


leres:
I couldn't find it in the user guide but I **think** this is done via:

> `config set modem auto SIM7600`

dexterbg:
That's nearly correct, it's:
```
config set modem driver SIM7600
```

The driver options are currently:
- `auto` = auto detect model (default)
- `SIM5360` = old UMTS model
- `SIM7600` = current LTE model
- `SIM7000` = unfinished driver for the Cat-M/NB modem

The power timing modification applies also with `auto`, but only after the model has been identified.

It's probably a good idea to add the driver selection to the web UI now.

dexterbg:
Uh… I think we need to roll back this commit. I just tested the modification with the driver set to `SIM7600`, and the system now does no longer recover the modem init from a crash.

My module keeps cycling between `PoweringOn` and `PoweredOn` states on the boot after the crash. With `auto`, crash recovery still works.

Craig, can you please test / confirm this? Set the driver, do a normal boot, then use `module fault` to simulate a crash.

drc38:
Interesting, the reason I implemented the change was because my unit wasn't recovering from after a crash. Showing the same symptoms you describe Michael. I'll try the simulation tonight and report back.

drc38:
Can you also try the test above with a prior firmware, to see whether it is the code changes or the problem already existed.

leres:
> Craig, can you please test / confirm this? Set the driver, do a normal boot, then use `module fault` to simulate a crash.

Confirmed, my dev unit is unable to bring up the cell after a crash when running g20235b10.

leres:
> Can you also try the test above with a prior firmware, to see whether it is the code changes or the problem already existed.

I tried 3.3.003-556-gd67fcf6a and comes up ok after a crash. I still had driver set to SIM7600 but likely nothing was looking at it.

dexterbg:
Simulating a crash is crucial for testing modifications to the modem's power state detection. I remember Mark having quite some headache over getting this to work reliably with both models.

Only `SIM7600` after the merge fails to recover. Attached are my logs from `auto` & `SIM7600` mode, each before and after the PR merge.

[crashlog-auto-beforepr.txt](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/13260745/crashlog-auto-beforepr.txt)
[crashlog-sim7600-beforepr.txt](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/13260744/crashlog-sim7600-beforepr.txt)
[crashlog-auto-afterpr.txt](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/13260742/crashlog-auto-afterpr.txt)
[crashlog-sim7600-afterpr.txt](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/files/13260743/crashlog-sim7600-afterpr.txt)

The system recovers after doing a standard `module reset` when stuck in the state loop. I also don't think many users actually configured the driver, as that was an undocumented feature only known to developers up to now.

So I'll keep the merge for now -- @drc38, please fix ASAP.

drc38:
I can't seem to replicate your results, is it simply issue `module fault` in the shell then await reboot and issue `power on cellular`? With mine recovering ok when following that sequence. Based on your crashlogs it looks like the timeout tickers need to be increased. I can only assume the extra identify step when set to `auto` gives extra time so the timeout is not reached. I've increased  timeouts across the main states and am testing on my unit before send through PR.

dexterbg:
I was referring to a crash happening with the modem in full PPP operation. If you need to power on the modem manually, you probably don't have modem auto start enabled. Most users have the modem enabled by default.

dexterbg:
In case you're not aware of this: the difficulty with the modem in PPP operation is, it won't react to AT commands, as it's still in MUX mode. The modem also has no means to detect a lost MUX connection and return to normal mode. So the driver needs to try a series of modem power cycles until the modem responds normally.

drc38:
Ok, thanks for clarifying Michael. Note, the driver changes made do not change the power cycling logic, simply the timing to match the spec sheet for the 7600. From what I can tell the issue seems to be hitting the timeouts before the modem/network communication has been finished.

PR #953, opened 2023-12-02
Metrics - handle registering of array elements properly

frogonwheels:
- Name (as const char *) in map was pointing to stack memory in metric arrays.

moving to std::string does use more memory but don't see a way around this.


dexterbg:
Nice find, thanks for fixing my stupid bug  :-)
And yes, better use `std::string` than `strdup()`.

PR #958, opened 2024-01-07
Vehicle - use moduleid received  instead of 'moduleid_low'

frogonwheels:
This is not necessarily essential since moduleid_low is (now) in the state, but I believe it makes more to switch on the received module than the 'expected low'  even though that should amount to the same thing.


frogonwheels:
One of the reasons I would like it included is for the sake of copy-pasted code I think it is an improvement.

PR #960, opened 2024-01-07
OBD Poller - Cleaner interface to Poll-List and retrieving next entry

frogonwheels:
The aim here is to move away from vehicle implementations knowing anything about the poll-list pointers, and to move towards a system where there is a function call to 'get the next poll entry'. 
The biggest change here is that it moves a while loop from PollerSend() to a factored NextPollEntry()

-  Factor retrieving of next poll entry
-  Pass in Enum of poll call reason Primary/OnceOff/PollSuccess
-  Clean up resetting of ticker.
-  Introduce PollRunFinished() (as a safe place to check for Poll State changes.)

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_jaguaripace/src/ipace_can_handler.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_jaguaripace/src/ipace_can_handler.cpp
@@ -121,7 +121,7 @@ void OvmsVehicleJaguarIpace::IncomingPollFrame(CAN_frame_t* frame)
 
     }
     // Handle multi-line version responses
-    else if (m_poll_plist == nullptr && frameType == ISOTP_FT_CONSECUTIVE)
+    else if (HasPollList() && frameType == ISOTP_FT_CONSECUTIVE)
     {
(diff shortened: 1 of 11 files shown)


vehicle/OVMS.V3/components/vehicle_jaguaripace/src/ipace_can_handler.cpp
```diff
@@ -121,7 +121,7 @@ void OvmsVehicleJaguarIpace::IncomingPollFrame(CAN_frame_t* frame)
 
     }
     // Handle multi-line version responses
-    else if (m_poll_plist == nullptr && frameType == ISOTP_FT_CONSECUTIVE)
+    else if (HasPollList() && frameType == ISOTP_FT_CONSECUTIVE)
```
dexterbg, line 124: Logic reversed (seems to be an unused if branch, but nevertheless…)

dexterbg (changes requested):
I hope I didn't miss a control flow flaw. Found just a minor issue, but please fix so we've got everything together.

frogonwheels:
Yeah, I thought I'd reviewed those. Doh.

PR #962, opened 2024-01-08
Internal GPS - Handle speed/direction missing from data

frogonwheels:
- Blank direction will leave the value as it was
 - Blank Speed only zeros speed after 10 seconds.

Currently the '10 seconds' is hard-coded and rather arbitrary... but I think better than just always clobbering it as soon as there is a blank speed (which is 'unspecified' rather than 0)

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_cellular/src/gsmnmea.cpp
+++ b/vehicle/OVMS.V3/components/ovms_cellular/src/gsmnmea.cpp
@@ -243,6 +243,7 @@ void GsmNMEA::IncomingLine(const std::string line)
 
     if (std::getline(sentence, token, ','))
       strncpy(time, token.c_str(), 6);
+    bool speedok = false, directionok = false;
     if (std::getline(sentence, token, ','))
       {;} // Status
     if (std::getline(sentence, token, ','))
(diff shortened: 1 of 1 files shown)


frogonwheels:
I'm not 100% sure this is the right thing to do, but I personally like it. 

 The thing I was seeing was this (it's a small thing), that I found a bit weird was that whenever I parked it was resetting the orientation to 0 degrees (North) in the application.  I though the orientation wasn't working (since I hadn't been following along while driving and seeing it work since I'm the only driver of my car).  This, especially for the direction, was the reason I did this.

The speed was just because I was reading the protocol descriptions that these are 2 of the values that can be blank meaning no data and figured we could do a bit better than just blat it to zero km/h.

So to be clear, this is ONLY when the field comes in as BLANK.

dexterbg:
It's certainly right to not update the metrics when the GPS system signals "no data", but I see an issue with setting the speed to 0 after 10 seconds: our mechanism to check for missing updates is the staleness, which is btw also `SM_STALE_MIN` = 10 seconds for the GPS speed. GPS speed is really just the GPS info, and should IMO not be cleared by a timeout in the GPS reader. By clearing the GPS speed, you also prolong the "freshness" by another 10 seconds. This would also create a need for all vehicles reading GPS from the car instead of the modem to do likewise, or the handling would become vehicle context dependent for no good reason. Likewise, if we introduce that behaviour just for that metric, that would be inconsistent to other similar metrics (e.g. the GPS direction here already).

TL;DR: I'd rather keep the metric without updates running into staleness.

frogonwheels:
Sounds good.
//.

PR #965, opened 2024-01-15
OBD Poller - Implement Pause/Resume of polling

frogonwheels:
This allows us to move the poller list to 'private' and away from access by vehicle implementations.


dexterbg:
Well, basically there's already the `PollSetPidList()` call to keep vehicles from directly accessing the poller list.

I implemented a poller pause command for the VW e-Up, that could also use these calls instead. If this is a feature of general interest, a standard shell command and possibly also a Duktape call would make sense.

PR #966, opened 2024-01-16
OBD Poller - Allow variations to the polling time

frogonwheels:
- Allow primary and secondary poll lengths to be varied. (Time between poll attempts). 
 - Retains the 'poll on success of last poll' concept.  (This could be an option).
 - Adds Secondary ticks option (Secondary tick will attempt the next poll but will not reset to the beginning)
 - Move PollerSend() to Rx Thread.  This takes the burden of sending away from the main vehicle ticker thread.
 - Aims to Decrease latency of OBD Poll Loop ( eg for obd2ecu)
 - ResetThrottle() is done in the main ticker so its timing is not changed by the ticker time

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -30,6 +30,7 @@
 
 #include "ovms_log.h"
 static const char *TAG = "vehicle";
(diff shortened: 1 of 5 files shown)


dexterbg:
Michael, I assume this PR is now obsolete?

frogonwheels:
Absolutely

Michael

PR #968, opened 2024-01-20
Various fixes related to C++20 compiler

frogonwheels:
Fix warnings that have been introduced by the switch to C++20 compiler in ESP-IDF v5+.

* Infinite recursion detection
* Deprecated implicit cast of enum to float
* Transition of implicit capture of this (deprecated) to explicit capture of this
* Enhanced 'unused result' detection in macro expansion.
* A deprecated 'nullptr' conversion issue


frogonwheels:
If there are any questions I'm happy to break this up - or extract problematic commits from this patch set.

PR #977, opened 2024-02-26
VWEUP: rework climate contol, add charge control

sharkcow:
implementation of charge/climate control "profile0" as state machine incl. retries
- fix awake-asleep loop (heartbeat timer not stopping in sleep)
- rework of wakeup function for dependability
- rework of remote climate control (fix deleting of charge timer settings in ECU)
- add charge control including stop on maxSoC
- add charge port detection
- add official VW SOH
- fix: transmit SOH as float


sharkcow:
Sorry, a bug was found. Should be fixed soon, then we reopen

PR #983, opened 2024-03-24
Update README.md - remove fasttech

fowi4hjte:
Fasttech is not a distributor of OVMS any more since a longer time but still listed here.


markwj:
OK. Merged. Thanks.

PR #984, opened 2024-03-25
SmartED 451

Dimitrie78:
fix cell volt offset
add option to set custom cell volt poll time


dexterbg:
Dimitrie, I'll merge this now, but please don't forget to add an entry to the `changes.txt` for all new user level features.

PR #987, opened 2024-03-26
Update README.md - update description to represent hardware version 3.3

fowi4hjte:
The current hardware version 3.3 is the one with the highest cost. Many people report on the internet that they do not get the OVMS because of the high cost. Its not that affordable for many people so just remove the comment about the price because it does not add any benefits. The people see the price at the distributor and decide on their own if they can afford it. Hardware version 3.3 have dual certification and the modem can be used worldwide. Also it supports 2g, 3g and 4g networking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ module and the web frontend can be customized by plugins. The module has three b
 extended by a fourth one](https://github.com/mjuhanne/OVMS-SWCAN).
 
 The **OVMS base component** is a small hardware module that connects to the vehicle **OBD2** 
-port. The standard kit in the current version 3.3 includes a 2G/3G/4G modem to provide **GSM**, **UMTS** and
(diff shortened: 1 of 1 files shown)


markwj:
It is a goal of the project to keep costs down and to ensure that the product is as inexpensive as it can be. Apart from component and manufacturing costs, the most expensive fixed cost at the moment is certification costing tens of thousands of USD each time. That is why the v3 product is more expensive than the uncertified older v2. As always, we publish schematics and layouts so others can build if they want (avoiding the certification issues). Compared to other comparable products, and compared to the price of the car, this is inexpensive.

The readme is very core to the project as it should summarise in a few simple sentences our goals and solution. As such we would like to keep it generic. I agree that GSM is an old term, no longer accurate, so perhaps simplest to change to 'cellular'?

fowi4hjte (commented):
Done

markwj (approved):

PR #1001, opened 2024-04-21
Ioniq5 add ev6 descriptions

frogonwheels:
The Ioniq 5 configuration appears to work fine for EV6


frogonwheels:
Also tightened up support for the vin decoding to include ev6

PR #1007, opened 2024-04-28
Poller tweaks

frogonwheels:
Start with poll sequence 0

Remove unused call-backs 

Ignore setting top-level state to same value


frogonwheels:
Tweaked some logs

PR #1009, opened 2024-05-01
NL: Avoid flipflop of 12v charging metric

drc38:
Float charging can be as low as 12.9V and create nuisance logging going between 12.9 and 13V


dexterbg:
If the code depends on a high 12v measurement precision like here, I recommend adding a calibration note to the manual.

PR #1011, opened 2024-05-11
Poller - Short-cut protocol handling if there is no response pending

frogonwheels:
Looking at reducing unnecessary poller task loop processing to reduce time per incoming payload

### DIFF
--- a/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
+++ b/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
@@ -71,7 +71,7 @@ OvmsPoller::OvmsPoller(canbus* can, uint8_t can_number, OvmsPollers *parent,
   m_poll.moduleid_sent = 0;
   m_poll.moduleid_low = 0;
   m_poll.moduleid_high = 0;
-  m_poll.type = 0;
+  m_poll.type = VEHICLE_POLL_TYPE_NONE;
   m_poll.pid = 0;
   m_poll.mlremain = 0;
   m_poll.mloffset = 0;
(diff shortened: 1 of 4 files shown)


dexterbg:
I think this isn't correct -- wouldn't it break VWTP polling, as the VWTP poller doesn't use `m_poll`?

The previous poller did the running poll test (on `m_poll_wait`) only for the ISOTP poller:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/release-3.3.004/vehicle/OVMS.V3/components/vehicle/vehicle.cpp#L494

frogonwheels:
It still uses m_poll for vwtp.  That's where the poller grabs the
information to regardless of the protocol.
So this bit sets the protocol
    case OvmsNextPollResult::FoundEntry:
      {
      // We need to poll this one...
      *m_poll.protocol = m_poll.entry.protocol;*
      m_poll.type = m_poll.entry.type;
      m_poll.pid = m_poll.entry.pid;

So yeah - it should work fine for vwtp.

//.

frogonwheels:
I'm actually hoping that this will be most of the issue we are having with the Nissan Leaf, which seems to be sending > 100 CAN messages per second on the bus.
I believe @dexterbg  is right when he said we need to look at items in the poller thread that were taking too long, and I'm thinking that if we can short-cut this, then the queue won't get overwhelmed.

frogonwheels:
Confirmed that VWTP will be fine - it only uses the m_poll_vwtp for some of the state.. it's still mainly m_poll for the start.
I added a bit to clear that m_poll.type on confirmed success of either VWTP or ISOTP .. which nicely factors some common code between the two protocols.

PR #1012, opened 2024-05-11
Vehicle Shutdown - Wait for loop shutdowns and close down as nicely as possible

frogonwheels:
- Trying hard to avoid race conditions.
 - Signal loops so that we can shut down.

### DIFF
--- a/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
+++ b/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
@@ -786,8 +786,7 @@ void OvmsPollers::ShuttingDown(bool wait)
     entry.entry_type = OvmsPoller::OvmsPollEntryType::Command;
     entry.entry_Command.cmd = OvmsPoller::OvmsPollCommand::Shutdown;
     entry.entry_Command.parameter = 0;
-    xQueueReset(m_pollqueue);
-    xQueueSend(m_pollqueue, &entry, 0);
+    xQueueSendToFront(m_pollqueue, &entry, 0);
     }
 
   MyCan.DeregisterCallback(TAG);
@@ -1196,8 +1195,7 @@ void OvmsPollers::PollerTask()
   if (task)
     vTaskDelete(task);
   // Wait to die.
-  while (true)
-    vTaskDelay(pdMS_TO_TICKS(10));
+  vTaskSuspend(nullptr);
   }
 
 OvmsPoller *OvmsPollers::GetPoller(canbus *can, bool force)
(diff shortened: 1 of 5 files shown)


dexterbg:
The task seems to have the shutdown handling twice?

Also, the task will `continue` after receiving a shutdown request, which means it will again call `xQueueReceive()`, so the queue doesn't become free to be deleted cleanly. It will work I think, but I suggest exiting the receive loop on either the shutdown request or detecting `m_shutdown` set, and doing the task deletion after the loop.

frogonwheels:
I hear you.

The shutdown(true) should be called by the vehicle factory though before
the delete which waits a bit, giving time for the thread to shut itself
down.  So it really shouldn't be doing anything in the destructor... that's
more a just-in-case.

//.ichael

frogonwheels:
Because of this code which is the only place where m_shut_down is set - it
immediately sends a Shutdown command to the poll task.
So m_shut_down flag just means the ONLY command that it will obey from then
on is the Shutdown task. By the time it gets to the queue being deleted the
task is most definitely shut down.

void OvmsPollers::ShuttingDown(bool wait)
  {
  if (m_shut_down)
    return;
  m_shut_down = true;
  if (m_polltask)
    Queue_Command(OvmsPoller::OvmsPollCommand::Shutdown, 0);

frogonwheels:
This is what happens (Sorry – was not awake so making sure I’m not crossing my wires from what used to happen to what is happening now).

 

In going through this I made some minor adjustments along the lines of your suggestion.

 

When the event  ‘system.shuttingdown’ is fired.

 

*	Calls OvmsPollers::Shutdown(wait = true)

*	Sets m_shut_down =true.  This stops any command from being queued, and triggers the loop to be exited.

*  At some point from here, the loop exits and kills the tasks and then waits until it dies (??) 

*	Clears the queue and Sends a Shutdown command - which has the effect of making sure we stop waiting on the queue.
*	Unregisters events and callbacks
*	Clears the poller lists in the pollers
*	Waits for the poller task variable to be nulled for 2s
*	Deletes the task if it isn’t already .. but really it should be.
*	Deletes the queue
*	Powers down the can busses
*	Deletes the poller instances.

dexterbg:
Minor notes:
- Instead of `xQueueReset()` followed by `xQueueSend()`, it's better to do `xQueueSendToFront()`.
- Suspending a task: `while(true) vTaskDelay(pdMS_TO_TICKS(10));` is a busy loop. You can do `vTaskSuspend(NULL)` or `vTaskDelay(portMAX_DELAY)`.
- `OvmsVehicle::VehicleTask()` is missing that task suspension.

frogonwheels:
Agreed.
Will get that done.

PR #1014, opened 2024-05-21
Poller - Make sure state 0 happens on the first run after being reset

frogonwheels:
The poller state used to get reset to 0 ensuring it fired then.
This restores that behaviour.


dexterbg:
Are you sure this will compile? Where is the `IFTRACE` definition?

frogonwheels:
ooh. damn. Sorry.
Yeah - no it won't compile, it needs the logging P/R
I'll get it compiling (unless you want to merge that).

frogonwheels:
Which is now P/R  #1015 - seems like I didn't create the pull-request. Sorry about that.
Let me know which one you want to include first so I can adjust the commits accordingly.

frogonwheels:
You can grab this one first.
Then I'll deal with #1015.

PR #1024, opened 2024-06-24
Vehicle - Cleaner swap / shutdown of vehicle

frogonwheels:
I'm not sure whether the m_pending list is just way overkill... but given that it's mostly empty it maybe ok?

Feedback welcome.


dexterbg:
Sure looks like overkill, but strictly correct: as the vehicle shutdown is now asynchronous, there may be multiple vehicles shutting down (although I wouldn't bet on a stable system in that case).

But I suggest renaming `m_pending` to a more specific `m_pending_shutdown` or similar, so reading the code makes it's purpose directly clear.

PR #1027, opened 2024-07-13
VFS - Add file completion

frogonwheels:
Add completion for vfs command line.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -6,6 +6,7 @@ Open Vehicle Monitor System v3 - Change log
     add TPMS pressure
     add Internal Temp
     add Doors status
+- Support for completion in vfs commands.
 - Separate Polling from the Vehicle implementation
   New commands:
     poller status
--- a/vehicle/OVMS.V3/components/can/src/canlog_vfs.cpp
+++ b/vehicle/OVMS.V3/components/can/src/canlog_vfs.cpp
@@ -33,6 +33,7 @@ static const char *TAG = "canlog-vfs";
 #include "ovms_utils.h"
 #include "ovms_config.h"
 #include "ovms_peripherals.h"
+#include "ovms_vfs.h"
 
 void can_log_vfs_start(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
   {
@@ -81,7 +82,7 @@ OvmsCanLogVFSInit::OvmsCanLogVFSInit()
           "<path> [filter1] ... [filterN]\n"
           "Filter: <bus> | <id>[-<id>] | <bus>:<id>[-<id>]\n"
           "Example: 2:2a0-37f",
-          1, 9);
+          1, 9, true, vfs_file_validate);
         }
       }
     }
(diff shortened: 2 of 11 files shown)


dexterbg:
I've got the impression you partially "reinvent the wheel" here. Did you have a look at extending the `Direntry` class & `read_entries` utility?

dexterbg:
It's a bit inconvenient this only works on selected commands, it should work on any file/path argument. Telling or tagging which are file/path arguments is complex. Maybe it could react to any argument beginning with "/"?

dexterbg:
Also, completions of directories end with a blank after the trailing "/". Adding the blank on a complete file name is OK, but  directories probably should not do that, to allow tabbing through files from there without the need to erase the blank first.

frogonwheels:
There's a couple of reasons.  I initially used that function but then I
also implemented globbing there and that started to get complex. Also I was
having some speed problems completing a directory with 50 files... and was
trying to avoid all the stat() calls.

If you wanted you could accept the globbing p/r (I'll make some comments on
that) where I added a parameter to stop the stat() when using it for rm
globs.

//.

frogonwheels:
The completion works on all the vfs commands that have paths.

I'm happy to go through and add it elsewhere if needed.

As far as the space after the directory goes, fixing that requires some
changes to the completion functionality.. somehow we need to mark the
completion element as non-terminating so it doesn't add the space.  Ideas
on how to mark that would be welcome.

I could reimplement the completion memory area as a std::list with a struct
that contains a flag?  Or I could hack it by adding a special prefix to the
string....  what do you think?

//.

leres:
An observation: Doing what [tcsh](https://www.tcsh.org/) does is a good model. (I'm old enough to remember the AT&T encumbered csh on Berkeley Unix that did not put a space after completed filenames -- it took a minute to adjust but I see now it's probably better...)

frogonwheels:
The reason I didn't supress adding of the space as this would require modifying microrl/microrl.c (microrl_get_complite function)  which is where all the command-line completion happens.  It always adds a space if there's only 1 completion solution matching.

I think it is worth doing - but as a separate  p/r maybe.

Because this is c code it might require the hack solution.. like maybe if the complete element starts with a char(1) it skips the char when printing but then doesn't add the " ".  Tbh having to press backspace to complete the rest of the file is a little inconvenient but still a whole lot of improvement over having no file completion at all.

As far as completing anything with '/' obviously this would require a whole thing - maybe some kind of default completion that looks at the usage hints or something.  Again, this would be a whole separate p/r.

I'm happy to add it in for particular commands (like I guess in the logging) - there's a base function to call for when you know that parameter should be completing a filename  ( vfs_expand ) that could be made public to be accessed by other modules.  It would also be a good common point to mark items as not requiring a space.

dexterbg:
A very simple patch to microrl would be to not add the space if the completion ends with a "/". There's little chance that's something else than a file path, we don't use "/" for any system object names, and certainly not at the end. MQTT paths could be affected if/when completion is added there, but then in terms of completion that's also a "directory" case.

I agree on the `stat()` performance issue.

In terms of commands, first I tried was script execution and config backup / restore, so outside the vfs realm. `log file` comes also to mind, and the can logging of course. Probably best to scan all command registrations for "file|path|dir".

Do you want to extend the PR, or shall I merge it now?

frogonwheels:
I'll see what I can do this afternoon if it is simple enough.

Michael

frogonwheels:
Pushed up a change that extends the filename completion to all the areas I could find.

Collapse of extra space coming in a separate p/r.

PR #1028, opened 2024-07-13
Add simple '*' and '?' globs to VFS functions  ls, rls and rm

frogonwheels:
Note this uses a third-party matching function provided by Joshua J Baker under the MIT license


dexterbg:
Michael, this PR now has a merge conflict, can you have a look? Thanks!

PR #1031, opened 2024-07-17
Console - Improve completion support for paths

frogonwheels:
- Gets the completion length via the callback to count for dropped entries
- Supports not adding space when a completion item (path) is not 'complete'


dexterbg:
Works like a charm :+1:

PR #1032, opened 2024-07-20
Poller - int32_t format strings changes

llange:
In ESP-IDF v5+, a compiler change has been made for type int32_t and uint32_t : Cf https://docs.espressif.com/projects/esp-idf/en/latest/esp32/migration-guides/release-5.x/5.0/gcc.html#int32-t-and-uint32-t-for-xtensa-compiler

This change forces us to update the `printf. format strings and use, as much as possible, the [PRI... portable format constants](https://en.cppreference.com/w/cpp/types/integer)

In addition, a constant in `Atomic_Increment()` call needed to change in order to compile with ESP-IDF v5+ compiler.

### DIFF
--- a/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
+++ b/vehicle/OVMS.V3/components/poller/src/vehicle_poller.cpp
@@ -292,7 +292,7 @@ void OvmsPoller::PollerNextTick(poller_source_t source)
   {
   // Completed checking all poll entries for the current m_poll.ticker
 
-  IFTRACE(Poller) ESP_LOGD(TAG, "[%" PRIu8 "]PollerNextTick(%s): cycle complete for ticker=%u", m_poll.bus_no, PollerSource(source), m_poll.ticker);
+  IFTRACE(Poller) ESP_LOGD(TAG, "[%" PRIu8 "]PollerNextTick(%s): cycle complete for ticker=%" PRIu32, m_poll.bus_no, PollerSource(source), m_poll.ticker);
   m_poll_ticked = false;
 
   if (source == poller_source_t::Primary)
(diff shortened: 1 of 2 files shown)


dexterbg:
Ludovic, this change doesn't compile:
```
/…/components/poller/src/vehicle_poller.cpp: In member function 'void OvmsPollers::Queue_PollerFrame(const CAN_frame_t&, bool, bool)':
/…/components/poller/src/vehicle_poller.cpp:1716:32: error: no matching function for call to 'Atomic_Increment(volatile uint32_t&, long unsigned int)'
     Atomic_Increment(count, 1lU);
                                ^
In file included from /…/main/ovms_command.h:42:0,
                 from /…/components/poller/src/vehicle_poller.cpp:36:
/…/main/ovms_utils.h:743:5: note: candidate: template<class T> T Atomic_Increment(volatile T&, T)
   T Atomic_Increment( volatile T &variable, T amt)
     ^
/…/main/ovms_utils.h:743:5: note:   template argument deduction/substitution failed:
/…/components/poller/src/vehicle_poller.cpp:1716:32: note:   deduced conflicting types for parameter 'T' ('unsigned int' and 'long unsigned int')
     Atomic_Increment(count, 1lU);
                                ^
```

llange:
I'm sorry... I did my compilation test with `CONFIG_OVMS_COMP_POLLER` undefined :-( 
You can roll it back and I'll provide a fix in a few minutes.

llange:
Puzzling: even this fails:
```
Atomic_Increment(count, UINT32_C(1));
```

with the same error, while I believe it should work - whatever the size of `uint32_t`.

Any idea ?

llange:
It's not beautiful but this works on both ESP-IDF compilers:
```c
Atomic_Increment(count, (uint32_t)1);
```
I'll prepare a new MR for that.

llange:
@dexterbg let me know if it correctly fixes the compilation for you (it does for me), and sorry again for breaking master...

PR #1035, opened 2024-07-20
Vehicle_Vweup - fix a compilation issue

llange:
The compiler in ESP-IDF v5+ is very sensitive to detecting potential issue. Here, an indentation could be interpreted by the human reader as if an `if` block was encompassing a whole indented block.

The fix here is either to remove the indentation, or to add curly braces to explicitely delimit the block managed by the `if` condition.

The indentation was removed.

**Note** : I erred on the safe side by remove the indentation, so that it compiles the same way it does on ESP-IDF v3. However, as I'm note familiar with the code and the intention of the `if` condition, I'd appreciate a full review of this block of code by a more knowledgeable person than me. Thanks :-)


dexterbg:
It's not my code either, but I think your interpretation & change is correct.

@sharkcow Heiko, please have a look as well -- thanks!

PR #1038, opened 2024-07-22
Duktape - Finish Implementing console commands

frogonwheels:
- Command map moved to std::string (by necessity)
 - Implement proper shut-down of duktape


dexterbg:
Awesome :+1:

PR #1043, opened 2024-08-05
add trip time to trip report

sharkcow:
good for comparing different regular routes as to which is the fastest one :)


dexterbg:
Sharkcow,

this addition has some minor code maturity issues, you might like to fix:

- There is no good reason to use a ms resolution (`esp_log_timestamp()`) for a trip duration, and it leads to unnecessary followup issues, e.g. needing to think about unsigned overflows (when the log timestamp cycles back to 0 during the trip). I suggest using `time()` instead, which gives seconds resolution.
- Using `float` arithmetic for the hours/minutes/seconds calculation is overkill as well. Simple & fast integer division ('/') and remainder ('%') operators will do just fine.
- The output formatting is faulty if the trip is between 1:00 and 1:10 hour -- the minute won't be filled to 2 digits then.

Regards,
Michael

sharkcow:
Thanks for the valuable feedback Michael (as always :) !
I'll close and reopen with better quality code.

PR #1046, opened 2024-08-13
Ensure Network Stability Before Server Connection

Jamidd:
This PR introduces a safeguard to prevent the device from connecting to server v3 if there is a cellular or Wi-Fi connection that is not yet fully established.

For each connection type, the device checks whether the OVMS has been assigned an IP address or if the cellular connection (netMode + mux + ppp) is complete. It also takes into consideration thresholds for good and bad connection and does not attempt a connection whenever there is no good signal. 

If the connection is established but remains unready for a specified duration, the device will assume a failure has occurred and will automatically restart. This can also be disable via the config ("network", "reboot.no.connection") and is false by default.

Server v3 does not connect if less than 10 seconds of a "good " connection have passed.

The configs for cellular good and bad signal and automatic reboot if no connection were added to the web server.

### DIFF
--- a/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
+++ b/vehicle/OVMS.V3/components/esp32wifi/src/esp32wifi.cpp
@@ -375,6 +375,7 @@ esp32wifi::esp32wifi(const char* name)
   m_sta_reconnect = 0;
   m_sta_connected = false;
   m_sta_rssi = -1270;
+  m_good_signal = false;
   memset(&m_wifi_ap_cfg,0,sizeof(m_wifi_ap_cfg));
   memset(&m_wifi_sta_cfg,0,sizeof(m_wifi_sta_cfg));
   memset(&m_mac_ap,0,sizeof(m_mac_ap));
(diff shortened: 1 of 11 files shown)


dexterbg:
Jaime, welcome.

Making the info about the network status available to the applications/plugins via metrics is basically a good idea, and the automatic reboot to resolve network issues is also a good idea.

Using a ticker to poll for this info though is cumbersome and going against the system design. There are specific events meant to transport IP status changes to the application, and the netmanager already is subscribed to them, so we should rather handle the IP status tracking from the event handlers. In fact you already kind of did that half way for the Wifi network, just missed the signal quality handling. The same approach can be used for the cellular events.

The Wifi quality issue also reveals a basic shortcoming of your approach: the IP assignment status isn't enough to tell whether there is connectivity. The Wifi interface may have an IP address, but be currently not usable due to a poor signal level.

Also, instead of directly modifying the metrics from the event handlers, it may be a better idea to gather status flags and evaluating the overall status within the `PrioritiseAndIndicate()` method. This is meant to be generally called on any major network status change to decide on the network availability.

A side note: please stick to the code design principles seen in the modules, in this case use the `m_` prefix for class member variables, and stick to the same bracket & indentation style as used by the module.

On the automatic reboot: this needs to be configurable. Users may have situations where network availability is known to be gone for more than 300 seconds, but the module shall not reboot. Speaking of this: there are cases where it's more important the module continues to work, than having a network connection. A very important example of this is when the module needs to control the charge power/current in order to keep the vehicle from blowing a charge socket fuse. That's often used on camping grounds (6A sockets), which coincidentally also often have poor networks.

Regards,
Michael

PS: please also don't mix features in pull requests. The MQTT command throttling needs to be factored out into a separate PR.

Jamidd:
Hi Michael,
thanks for the insights. 
I will modify the code appropriately.

Thanks,
Jaime.

Jamidd:
Hi @dexterbg,
I believe it should be ok. Works ok for me.
Let me know if there is something I should change.

Jaime

dexterbg:
Jaime, nice work, much better overall solution & code quality now.

I'll merge this now so it can be tested, but please don't forget to add the new metrics & config to the documentation & to the `changes.txt` (user relevant change history).

PR #1047, opened 2024-08-13
Handle Flash Memory Corruption with Full Factory Reset

Jamidd:
This PR addresses the issue of internal flash memory corruption. When such corruption is detected, the OVMS will automatically perform a full factory reset. This process allows the device to rewrite the configuration file and resume normal operation.

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_config.cpp
+++ b/vehicle/OVMS.V3/main/ovms_config.cpp
@@ -46,6 +46,8 @@ static const char *TAG = "config";
 #include "ovms_boot.h"
 #include "ovms_semaphore.h"
 #include "ovms_vfs.h"
+#include "ovms_module.h"
+
 
 #ifdef CONFIG_OVMS_SC_ZIP
 #include "zip_archive.h"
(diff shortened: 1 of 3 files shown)


dexterbg:
Jaime,

it's a good idea trying to automatically recover from a `/store` corruption.

But performing an unconditional factory reset won't help users, in fact it will make things worse. As explained in the previous PR, a working module is normally preferred, even if it can no longer update it's config, it normally will continue to operate with the config last written.

If you perform a factory reset, the module will afterwards be unconfigured, unusable and appear as broken to normal users. Users would need to be aware it does that, then remember how to connect to it's initial network. Worst of this is, as long as the user needs to reconfigure the module, it will be open to any hacker knowing the system, as the initial module password is fixed.

A better primary/default option is to simply inform the user in case a corruption is detected by sending a notification, and leave further steps to the user.

An optional second level automation could be to schedule an automatic restore of the latest confirmed working backup, to be performed when the module is idle (i.e. not actively engaged with the vehicle or a user in any kind). To implement that, you need to keep track of backups as they are created. An automatic backup schedule would also help for this kind of function, and that should as well check for a corruption before creating a backup. Of course the module also has to be equipped with an SD card to be able to use that system.

Regards,
Michael

dexterbg:
PS: as with the previous PR, please stick to the code style used by modules you change.

Jamidd:
Hi Michael,

Apologies for the code style; I haven't had the chance to address it in the other PRs yet, but I'll make it a priority as soon as possible.

As for the corruption in /store, I've noticed this issue occurs whenever there's a device reset. When this happens, all configurations are wiped out, and the device becomes unreachable, both under the user and original configurations. The only solution I've found so far is to perform a factory reset.

At the moment, I don't have the time to work on the automatic restore. Given the risks associated with the default configuration, I'm going to close this PR for now. I'll try to work on the automatic restore as soon as I can.

dexterbg:
A corruption on each reset of course isn't normal, might be a sign of a faulty hardware.

Another option could be some process trying to write to `/store` without taking the shutdown procedure into account, so the write somehow interferes with the unmount. None of the standard system components should do that though.

Regards,
Michael

Jamidd:
It doesn't occur after every reset, but whenever it does happen, it's always following a reset. It's not a frequent issue, but it has happened a few times.

PR #1052, opened 2024-08-17
Ioniq 5 - Simplify & Improve 12V monitor

frogonwheels:
- Use short(4) and long(8) smoothed average comparison
- Detects a dip while 12V charging as well.
- Only polls while charging if it detects dip or after a longer period
- Factor out to functions


frogonwheels:
There are some reported issues with the hard-coded values in this patch that need to be resolved.

frogonwheels:
Yep that was deliberate. I was waiting on some info from someone testing it.

I'll make the needed tweaks to the numbers, merge and when it is ready I'll
take it off draft.

//.

PR #1054, opened 2024-08-19
Maxus Euniq 6 - Initializing module.  

atroncoso-tucar:
Purpose: 
Initialize support for Maxus Euniq6 model. 

Features included: 
- Speed
- SOC
- Odometer
- ON/AWAKE status
- Vehicle lock/unlock status. 

Notes: 
Noticed there already is a Maxus Euniq5 with 6 seats model but the can bus communication is quite different. So thought it would be best to propose adding a new model instead of modifying the existing one.


atroncoso-tucar:
@dexterbg,

Thanks! Happy to contribute. 
I'll keep those notes in mind.

PR #1076, opened 2024-10-23
SWCAN support for the Chevrolet Volt and Opel Ampera

Donovan22:
I’ve merged changes from another inactive fork (kssmll/Open-Vehicle-Monitoring-System-3's repo), and this request should bring all SWCAN-related fixes into the main branch for future development. The code is ready for testing and further refinement.

Please review and let me know if any adjustments are needed before merging.


Donovan22:
Hi Michael,

Thanks for the feedback! I see your point, but I'm honestly not sure how to make use of can4 for this purpose, and in GM vehicles, the third bus is typically SWCAN. Since the Volt relies heavily on SWCAN for diagnostics and control, repurposing can3 seemed like the most straightforward solution given the hardware constraints.

Ideally, if there’s a way to configure can4 for SWCAN without additional hardware modifications, that would allow us to keep can3 intact for other CAN functions.

If you have any suggestions on how to enable can4 in this context or know of any workaround that could help, I'd appreciate the guidance. I’m open to exploring options that would make this integration smoother and keep the channels consistent.

Best,
Donovan

dexterbg:
I'm talking about the OVMS API bus naming. What the "third bus" is for the vehicle, is irrelevant for the OVMS software, as different vehicles have different buses. On another vehicle, the primary CAN may need to be SWCAN.

From the OVMS software and API side, we've got the bus naming scheme "can1…4", with "can1" = ESP32CAN (internal), "can2" & "can3" = MCP2515 buses, and "can4" = SWCAN.

Even if "can3" and "can4" cannot be used at the same time due to hardware limitations (?), they still should have distinct names, as their name implies the hardware to be used. A firmware or script addressing "can3", assuming that's SWCAN, will run without issues on a module not equipped with the SWCAN extension. It cannot detect "can3" isn't actually SWCAN in that case, possibly causing strange issues and misbehaviour.

Donovan22:
I apologize for my misunderstandings here. I misinterpreted the OVMS bus naming conventions and the intended use of can4 for SWCAN. Your explanation makes sense, especially the potential complications if scripts assume can3 is SWCAN.

I’ll make the necessary adjustments to assign SWCAN to can4 and keep can3 as originally intended. I’ll close this pull request for now and make these changes in my repository before resubmitting. I appreciate your patience and feedback.

PR #1079, opened 2024-11-09
Server v2 - add charge_kwh_grid to MsgStats - Vehicle - keep charge time as long as charge port aktiv

Dimitrie78:


dexterbg:
Dimitrie, when extending a V2 message definition, you also need to update the documentation on this, and you also should add the new fields to the server's HTTP API (and update the documentation accordingly).

Dimitrie78:
I hope I haven't forgotten anything

dexterbg:
You forgot to update the V2 protocol documentation.

But there are some things to change first:
- Use a single metric for the combined charge date & time, and do not use a string for this, but an Int64 -- see other datetime metrics and the datetime units for reference.
- Optimize the metric explanation in the documentation: "remind last charging date" is lacking precision and has confusion potential. Better: "Date & time of last charge end" (if it's the end time you want to keep there)
- The explanation for "v.b.capacity" also should be clear about the kind of capacity metric it's meant to carry, i.e. is it the calculated (estimated) remaining usable (net) capacity, or the nominal new gross capacity, or something in between?
- We try to keep the "gen" metric set as close to the "charge" set as possible since their introduction, so when adding a new charge metric, also add the respective gen counterpart.
- Describe new features in `changes.txt` for end users, and only describe there what is relevant to them. End users will get confused by internal IDs / tech speak. Also, use English only ("aktiv").
- I'm not glad about yet another non-firmware metric being added to the "F" message. It was an error to allow the service info there. This is clearly rather a candidate for "D".

Regards,
Michael

dexterbg:
Dimitry, thanks for taking my suggestions into account. I've reevaluated your current PR and have some more suggestions and change requests:

- Comments & explanations of `v.g.timestamp` should not be "… charge …" but "generator" / "power generation".
- The V2 message "S" documentation lacks the new capacity field.
- The `changes.txt` still isn't user friendly, please at least fill in the missing metric explanations to the protocol changes.
- Metric `m.net.mdm.mode` needs to be moved into the V2 "firmware" (F) message. "F" is meant to transport OVMS & general system information, it already includes other connectivity related metrics. "D" is meant to transport vehicle environment related data ("doors", temperatures etc.).
- The V2 protocol addition of `m.net.mdm.mode` also needs to be documented.
- The server HTTP API documentation needs to reflect the new message field extensions (and the server PR needs to be updated as well).
- I need to object against creating the new charge/generator timestamps as persistent metrics. Space for persistent metrics is limited, this feature should only be used for important metrics that need to survive a reboot. It's acceptable for the battery capacity, as that may be needed for range calculations (although that should already be covered by the SOH metric, but I understand some vehicles interpret SOH differently). It's generally not acceptable for simple, only "nice to have" informational metrics like these. Both timestamps are technically superfluous, as all details about charge & generation phases can already be stored on the server & retrieved by any App via the grid history log.

Regards,
Michael

dexterbg:
Dimitry, you also need to `mp_encode()` the `m.net.mdm.mode` metric for the V2 message, as that's a string that normally will contain the field separator (comma).

PR #1081, opened 2024-11-16
Fix Server V3 client sending messages every minute

chris8086:
OVMS is a brilliant system, but I noticed that it was sending MQTT messages every minute, even when configured with updatetimes that were much longer than that. This activity consumes Hologram allowance and power unnecessarily.

The first step in fixing this is to filter out "clock" events, on the basis that they are of no use to anyone:
2024-11-10T22:52:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event', ... (10 bytes))
2024-11-10T22:52:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event/clock/2252', ... (0 bytes))
2024-11-10T22:53:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event', ... (10 bytes))
2024-11-10T22:53:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event/clock/2253', ... (0 bytes))
2024-11-10T22:54:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event', ... (10 bytes))
2024-11-10T22:54:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event/clock/2254', ... (0 bytes))
2024-11-10T22:55:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event', ... (10 bytes))
2024-11-10T22:55:00: Received PUBLISH from leaf (d0, q0, r0, m0, 'ovms/event/clock/2255', ... (0 bytes))

The second step is to amend Mongoose's keepalive timer, which defaults to 60 seconds. This timer generates these messages:
2024-11-16T12:11:06: Received PINGREQ from leaf
2024-11-16T12:11:06: Sending PINGRESP to leaf

This PR adds a new config parameter updatetime.keepalive, which is passed through to Mongoose.

I did some testing and identified that Hologram's TCP timeout is 30 minutes. For that reason, in this PR the default keepalive is set to 29 minutes in the hope that will be most sensible for most users. When OVMS metrics are published, Mongoose resets the keepalive timer so if the user has configured updatetimes less than that, PINGREQs will never need to be sent.

I'm happy to make changes to the PR if you'd prefer something a little different.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,21 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- OVMS Server v3: Ability to define less frequent MQTT communication to save data charges and power
+  Events:
+    Excludes clock.* events from being published at the start of every minute.
+  New configs:
+    [server.v3] events.legacy_topic  -- Whether to publish every event on the legacy /event topic as well as
+                                        the MQTT-style topic. Defaults to true to preserve existing behaviour.
+                                        Recommended to set to false if you don't use the legacy topic.
+    [server.v3] updatetime.keepalive -- Max number of seconds to allow the MQTT connection to be idle before
+                                        sending PINGREQ. Should be set slightly shorter than the network's
+                                        NAT timeout and the timeout of your MQTT server. If these are unknown
+                                        you can use trial and error. Symptoms of keepalive being too high are
+                                        a lack of metric updates after a certain point, or "Disconnected from
+                                        OVMS Server V3" appearing in the log. In previous releases this used
+                                        the Mongoose default of 60. New default value 1740 aligns with
+                                        observed Hologram timeout of 1800.
 - Metrics
   add new Standard Metrics
     v.b.capacity          -- Main battery usable capacity [kWh]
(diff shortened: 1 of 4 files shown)


dexterbg:
Chris, welcome.

Three things regarding this PR:
- Compare the full event name space prefix, i.e. including the dot: `clock.` (similar to the `ticker.` event filter).
- Add a user level explanation of the new config parameter to the `changes.txt` file. Include a brief info on how to test the timeout of the connection & MQTT hub used.
- Add the new config to the V3 web configuration UI -- all the other update intervals are already there, so this should be in the UI as well (and easy to add). Add the timeout test instructions there as well.

For later, you might consider adding another `IdIncludeExcludeFilter` as an `m_events_filter`, see the ServerV3's `m_metrics_filter` for reference. That way, users can easily define custom exclusions.

You also might consider adding a config switch to disable (or enable?) the legacy event publishing on the fixed `/event` topic.

Regards,
Michael

chris8086:
Thanks Michael. I'll work on those. Regarding the last part, do you mean disabling legacy event publishing entirely? Happy to add that as well.

dexterbg:
I mean making this …
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/470979369be56fe18eae0df90f1619afdd27e850/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp#L538-L541
… depend on a config.

chris8086:
Ok, requested changes made, except the "for later" part. This work required a number of decisions (e.g. name of config parameter, where to put descriptive text) so I picked what seemed to make sense, but feel free to change them to suit.

PR #1096, opened 2025-01-12
Build fixes, bug fixes and warning eliminations exposed when using CMake and ESP-IDF v5

nattgris:
Tested with ESP-IDF v5.0.7 in-car and build-only tested using the regular ESP-IDF version.


dexterbg:
Welcome, Andreas, thanks for contributing! :-)

A first note: removing the uncompressed web UI assets will break the web UI development environment (https://docs.openvehicles.com/en/latest/components/ovms_webserver/docs/index.html#web-ui-development-framework).

Regards,
Michael

nattgris:
Hi!

Ah, I see. If you serve the files locally they are not available in their concatenated form. Let me see if I can do something about that use case. Would it be acceptable to make a wrapper python script that concatenates the files before starting the http.server?

nattgris:
Or maybe it is acceptable to configure the project to create the files before starting the web dev server? Then I can just drop the one-step changes and let the intermediate concatenated files be generated too.

Otherwise, I'll just keep the concatenated files in the repo. It's primarily the gz-files that I thought was good to get rid of.

dexterbg:
Sounds like introducing an additional place to take care of in case of asset changes, and also having the developer then need to deal with avoiding to accidentally check in the generated files. Why not simply keep the uncompressed files in the repository?

nattgris:
The main issue I had with the gz files in the repo was that they changed depending on how they were built, always dirtying the tree. These changes have also been committed to the repo, se e.g. 7371c6f363f6219f24d961b08d0730536c68a33c

But also, having generated files in the repo, conactenated or compressed, places the burden on the developer to make sure the committed files are actually in sync with the source file and not change only one of them. This has also happened in the repo, e.g. 1811311588f0cdda799f5323d54e0922fcce6340 and 6201fbb88bfc48b291388e07088105b2f50c3897 .

If concatenated files are generated, they should be in .gitignore, I think that solves the problem with accidental check-ins.

Regardless, it's not a big deal and I can change that part to just sync the compression setting between CMake and makefiles if you prefer. That will fix the real issue I had with it.

What do you think about the other changes?

dexterbg (commented):
You've dropped the IPv6 case in `network_status()`. The rest look OK to me, but I didn't try to build it with IDF 3.3 yet. Did you verify that still works?

nattgris:
The IPv6 case was broken anyway since it used IP2STR which only works with IPv4 addresses, so if DNS servers somehow got set to IPv6 address (not sure how), it would just print garbage. I figured since IPv6 addresses aren't printed elsewhere in the network_status() there's no point in doing it here.

Not sure how to print IPv6 addresses and I can't really test it since I don't get one here. Otherwise that might be the best solution if we want to support IPv6 (I have it disabled intentionally in my build). Although the rest of the settings would need to be adapted as well.

I removed the removal of concatenated assets, will push an update soon.

nattgris:
I rewrote the DNS print to use IP version agnostic ipaddr_ntoa_r(). However I don't know if it makes much sense to add it since no other IPv6 config is printed, such as the interfaces' addresses. Not even sure if the DNS address can get set to an IPv6 address. Not manually at least it can't since network.dns config is parsed as IPv4 only. Let me know if you want that patch added anyway.

PR #1102, opened 2025-01-21
SAE bus timing changes

drc38:


chris8086:
These values were shown by Michael Balzer to be incorrect in his list post on 24th January, when he attached a diff that should be used instead. I'd like to see that diff merged.

dexterbg:
Chris, this PR was just meant as a solution for those unable to apply a patch, and it only covered the first timing rework attempt.

I'll commit the final timing changes once I know these do solve the issue. If they don't solve anything, changing the timing, which otherwise has been working fine all the years, is kinda pointless.

I'm currently waiting for some definitive test results, or at least some indication of the kind "no incident since running the new timing".

chris8086:
Understood. I've been running the new timing for four weeks now. Just the one incident which I described on the list, and the working theory is that incident can be explained by low 12v. Today I've added immediate logging of any low voltage conditions detected by OVMS, so will monitor for a few more weeks and report back.

PR #1103, opened 2025-01-22
Can filter fixes

frogonwheels:
- Fix issue with RemoveFilter
 - Add some checks for correctness
 - Use busno integer instead of ascii values for filtering.
 - Improve the parsing

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/can.cpp
+++ b/vehicle/OVMS.V3/components/can/src/can.cpp
@@ -747,14 +747,18 @@ const char* canbus::GetErrorStateName()
   }
 
 
-uint32_t can::AddLogger(canlog* logger, int filterc, const char* const* filterv)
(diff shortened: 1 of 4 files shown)


dexterbg:
Michael,

looks good to me, except your new filter string parser isn't compatible with the defined format, as it doesn't allow the bus to be omitted. The filter spec was introduced for the can logging command, which documents the specification format as follows:

```
OVMS# can log start mon crt ?
Usage: can log start monitor crtd [filter1] ... [filterN]
Filter: <bus> | <id>[-<id>] | <bus>:<id>[-<id>]
Example: 2:2a0-37f
```

The bus may be omitted for convenience because many/most vehicles & applications only use one bus or can normally define the ID ranges of interest across all buses used.

Also, please add your extension of `<id>-` to the command help.

Regards,
Michael

frogonwheels:
Ahh. Yes. Will sort both those out.

//.

PR #1106, opened 2025-02-24
completely overhauled vehicle_smarteq (SQ)

zorgms:
Added:
- timebased heating (Web UI -> Features)
- recharge 12V when 12V alert raised (Web UI -> Features)
- switch GPS off when parking for power saving and reactivate GPS every 50 minutes for 10 min. to get new position (Web UI -> Features)
- maintance data: date, days, km and level (Web UI -> Status)
- fixed negative ambient/HV temperature
- calculate v.b.cac and v.b.capacity
- modem network can locked to LTE only (Web UI -> Features)
- gear shift -> metrics
- charge duration auto switch between OCS and math (Web UI -> Features)

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,13 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Smart EQ 453:
+    add timebased Pre-heat/cool Control 
+      - can set at Website -> Smart ED/EQ 4.Gen -> Climate/Heater 
+      - at App Features #4: on = 1 off = 2, #5 time like 05:15 = 515, #6 5/10/15 Minutes = 0/1/2
+    add GPS powersaving mode at parking
+    add 12V alert raised, the car starts the 12V charging process for 15 Minutes
+    add Network type switch, option to lock LTE only. Needed, when the Provider has no Datastream at GSM
 - CAN: add error flag decoding (included in can status output)
   New commands:
     can <bus> explain [errorflags]    -- Decode error flags into human readable text 
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/docs/index.rst
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/docs/index.rst
@@ -18,8 +18,9 @@ Vehicle Cable               OBD-II to DB9 Data Cable for OVMS (1441200 right, or
 GSM Antenna                 1000500 Open Vehicles OVMS GSM Antenna (or any compatible antenna)
 GPS Antenna                 1020200 Universal GPS Antenna (SMA Connector) (or any compatible antenna)
 SOC Display                 Yes
+SOH Display                 Yes
 Range Display               Yes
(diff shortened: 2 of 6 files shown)


dexterbg:
ZorG,

some remarks:
- If this includes user level changes, please describe them in the `changes.txt` file, probably also makes sense to extend the user manual on new features.
- Some of your new features look like framework candidates, especially the GPS parking auto-off and the modem network control. Ideas for framework additions should be presented/discussed on the developer list first though. Optionally binding the GPS activation to the drive status has been on my list for a while, I currently do that via event scripts. The sense of the modem network control escapes me… why would you want to restrict the network type? Please outline these on the developer list.
- Whenever you access specific components, be sure to check if they have a build switch, and guard the code by the appropriate preprocessor check. This applies for example to `MyPeripherals->m_cellular_modem` (the modem is an option) and to `MyWebServer` accesses (the webserver is also an option).
- You use the `MyWebServer.ExecuteCommand()` method quite a lot. If this isn't meant to create a dependency on the webserver component, you should create your own wrapper, or just call `BufferedShell::ExecuteCommand()` directly -- see implementation.
- I see you use the config store for potentially a lot of dynamic runtime changes, like the GPS state. Be aware any change to a config variable causes a full rewrite of that config param file, causing wear on the flash memory. That is not much of an issue, if it's used for very limited cases. If you use this for potentially high frequency changes, it may cause the flash memory to fail early.

Whenever you can do with just a RAM state, that's to prefer over storing in the config. Custom metrics can be registered as persistent, meaning they will normally survive a crash or reboot.

When you totally need to store something in a way that survives any power down time of the module, keep the working state in RAM and sync the RAM state to the disk state as seldom as possible.

When using the config store to save a bunch of state variables, register your own dedicated param scope (aside the vehicle scope and especially a shared plugin scope like `usr`), and use the `OvmsConfig::GetParamMap()` / `OvmsConfig::SetParamMap()` API to reduce multiple updates to a single write -- this also speeds up your processing substantially, be aware each call to some `OvmsConfig::SetParamValue…` API will rewrite the whole param file on the fly.

For really unavoidable high frequency updates, think about using your own database file and offer the user an option to place that on the SD card. A worn down SD card can easily be replaced.

Please tell me if you want to make changes first, or if this shall be merged now.

Regards,
Michael

zorgms:
The revision and consideration of your arguments is complete, I'm doing a few more tests and then the update will come. 

On the subject of network, in the smart-emotion.de community we were able to verify that, for example, the provider o2 does not have overall a data stream with GSM. This meant that the modem always successfully logged into the o2 GSM network, but had no connectivity. The solution was to set the modem to LTE only. We were unable to detect any problems with GSM connections with Telekom/Vodafone.

zorgms:
I'm finished yet.

dexterbg:
ZorG,

I've merged this now, but your build guards for the cellular peripheral are not correct: when building without the modem enabled, `MyPeripherals->m_cellular_modem` does not exist. You need to guard uses of the modem by…
```
#ifdef CONFIG_OVMS_COMP_CELLULAR
…
#endif // CONFIG_OVMS_COMP_CELLULAR
```

Regards,
Michael

PR #1114, opened 2025-03-16
Add Home Assistant integration documentation

enoch85:


enoch85:
@dexterbg I'm open for discussion on this one.

enoch85:
Here's a good read about MQTT overhead: https://medium.com/@dheeptuck/how-to-optimize-data-usage-over-mqtt-792abebd2cd1

enoch85:
cc @markwj as well. 👍

CrashOverride2:
Interesting! I thought about the same, but I use another approach I get data with integrated REST API from OVMS-Server v2. I use a multi SIM tariff with 150GB traffic so this is not a problem on my modules, but I still use v2 protocol because I like its efficiency.

I grab data from the modules with HA's integrated REST API function to poll these from my OVMS Server. Of course if public servers are used, there should be a limit how many polls per day should be allowed, in this case MQTT might be better.

If you like it and want to add it to your documentation, I attached the yaml file for home assistant. Of course this works with HTTPS encryption too.
[ha-rest.yaml.txt](https://github.com/user-attachments/files/19715302/ha-rest.yaml.txt)

ha_zoe_ovms

enoch85:
@CrashOverride2 That's super neat. 

My first problem as a new user though, I want ALL the entities(!) - and without knowing how to tinker with the OVMS module, I created an integration that does it for me. :+1:

CrashOverride2:
Ah ok this makes sense! In my first days with OVMS (2022) I want all metrics too and created a nice Grafana dashboard, but thats still too many, never looked at the dashboard again execpt for the first weeks of course :-)

enoch85:
I don't really care either, but thought I'd try, and it worked! The integration gives me ~175 entities for my VW e-UP, but I only need like 4. 😂

Anyway, I made it to make the step into OVMS easier, and we'll, it works. :)

MQTT is the future, and in 5 years it may be at the same stage as V2, so that's the reason for my decision. It also integrated nicely into HA.

enoch85:
😲 Thanks for the merge!

PR #1116, opened 2025-03-22
Fixing some smartEQ settings

zorgms:
Removed not needed `MyConfig.SetParamValueBool("xsq", "gps.off", false)`
Set Firmware download link to Dexter's Web EDGE Version, some memeber of smart-EMOTION.de community are confused which one is the latest FW.
Optimized Modem Networktype option auto to no preferred Network
Fixed Cmd Wakeup2

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -2,11 +2,6 @@ Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
 - Smart EQ 453:
-    Set Firmware download link to Dexter's Web EDGE Version
-    Optimized Modem Networktype option auto to no preferred Network
-    Fixed Cmd Wakeup2
-    Removed not needed config xsq gps.off
-- Smart EQ 453:
     DDT4all simple support, a full List of commands at www.smart-EMOTION.de
 - VW e-Up: decode & display battery cell level health status,
     provide v.b.capacity (net capacity in kWh)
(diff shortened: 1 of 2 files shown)


dexterbg:
ZorG,

unconditionally overriding the firmware update source configuration with fixed values is not acceptable. The user must be able to control which firmware branch to load and from where to fetch the firmware.

Also, "edge" is generally not recommended for end users. The edge build is the latest development state, it did and will contain bad bugs that may require direct access to the module, possibly flashing via USB, to recover from.

It's totally ok to ask _willing_ users to test the "edge" build, but it's not ok to force all Smart users into this.

I'm aware the last "main" release is quite outdated now, and there are a lot of updates regarding the Smart support in "edge". The next "main" release is due, I asked for feedback on this on the developer list already in January (http://lists.openvehicles.com/pipermail/ovmsdev/2025-January/016587.html), but shortly after that, a major CAN issue was reported, which is still under analysis.

You may give Smart users an _option_ to let the module change the OTA configuration for them, e.g. by adding a button for this to the Smart web config. But that may then only change the user configuration once, is must not force this on each boot, and it should also offer to switch back to "main". Maybe a better option is just to tell those users which entries to select from the server & tag dropdown in Config→Firmware?

Regards,
Michael

zorgms:
Since the development of the OVMS support for the smart EQ is already complete, it is no longer worth implementing your suggestion if the main version is updated soon anyway.

dexterbg:
Do you want to keep the `changes.txt` line as a recommendation?

Btw, the `changes.txt` should not be filled with development notes (like fixes), the file shall focus on update infos relevant to end users. Developer notes can be given in the commit messages and in code comments.

zorgms:
the `change.txt` can remain at the previous state

Issue #980, opened 2024-03-06
Connecting OVMS unit to running LEAF ZE0 caused orange car symbol on dashboard.


malyjaponec:
Connecting OVMS to running car immediately trigger an orange car symbol on dashboard as it generates some errors and it does not disappear with car restart. these errors are not temporary. I had to manually clear them.

Because of other issue the unit restarts during charging time to time (if it does not freeze completely) and these restarts causing more rerros so in one moment after 6-8th of charging that day the car went to not drivable condition. It did not switched on HV battery at all. I had to take a tool and remove all errors to drive out of charging place.

It seems to me that during startup it must block CAN bus somehow that some messages between car and battery are missing or must ask such requests that it prevents normal traffic on the CAN. 

the symbol on dash is almost instant. I do not know how long it takes unit to send first data to CAN but it looks like it must be before it. 

Car is upgraded ZE0 LEAF with 30kWh battery, using Dala's CAN bridge. There are always some non fatal errors in the car due to not perfect implementation of communication with battery bms but generally it did not caused any long / short term problems than occasional orange car symbol after DC charging that always disappear after car restart as the charger errors were temporary.

Connecting OVMS to the switched off car usually made no problem, but the car must be switched off for longer time when the all units sleeps, the symbol as i can remember lights up when I plug it in less then 5s after switching car off.

What else I can test or measure to find why this is happening?

art103:
Thank you for raising this issue! 
Given the cold weather recently starting to thaw, I thought I had an electrical fault with my OVMS unit causing *all* of the CAN systems in the car to fail simultaneously!
Turns out there was a firmware update whilst I was ChaDeMo charging at 2am (don't ask!). 

I am assuming that when OVMS rebooted, that is when the car's dashboard lit up like a Christmas tree.
It had "---" for range, car wouldn't drive, even the A/C wouldn't turn on, so this is a pretty major issue IMHO.
As you've said, the only way to fix is to unplug OVMS, clear all of the DTCs with LeafSpy, power off the car, then plug in OVMS.

FWIW, this is the firmware version that triggered it for me:
3.3.004-339-gf88eb24f/ota_0/edge (build idf v3.3.4-851-gfa4f07fb3 Jan 12 2025 00:05:24)

michalstral:
[...]

PR #1117, opened 2025-03-23
Add the option to avoid powering off Can busses on powerdown

frogonwheels:
Includes option to suppress power-down of Can busses for the Leave vehicle module


dexterbg:
I'm not sure if this is the correct "fix" for the Leaf issue (issue #980), possibly the controller still may or even must be shutdown, and just the SN65 transceiver needs to stay powered on.

But the extended control for the CAN shutdown behaviour makes sense for other application scenarios as well, so I'll merge this now.

PR #1122, opened 2025-04-02
Minor Leaf ZE1 enhancements / fixes

wayne-love:
Fix: Poll state not set to off when charging stopped by OVMS
Enh: SOH_instrument (xnl.v.b.soh.instrument) now float not int to avoid ZE1 instrument value rounding
Enh: Copy v.b.p.temp.max to v.b.temp so that max battery temp displays in the app


dexterbg:
Wayne, just a minor note: `changes.txt` is for users. Your addition there will sound like technobabble to them:

>    Fix: Poll state not set to off when charging stopped by OVMS
>    Enh: SOH_instrument now float not int to support ZE1 instrument values 
>    Enh: Copy v.b.p.temp.max to v.b.temp so that max battery temp displays in app

Please avoid describing internal fixes there, only add changes with major relevance to users. When adding such changes, describe them for non-IT readers.

PR #1126, opened 2025-04-06
Duktape add objects & commands for scripting polling lists

frogonwheels:
This series of commits adds various objects and commands for dealing with the bus and polled lists from a Duktape script.

It adds: 
* basic Bus management (Register/PowerDown) 
* Adding(/Removing) a poll list to the system poller.
  * The list is represented by a series of 'send' / 'decode' functions
  * The decode functionality leverages the already existing DBC decoding. Either by
    * Transferring a version of the packet consistent with DBC multi-packet decoding to an existing loaded dbc file
    * Create an internal dbc decoding structure that decodes against the payload-only version of the multi-packet ISOTP.  The fact that the DBC decoding is used internally is not important (only in that it avoids duplication).
 * Various actions to control the poll state.
 * Support for once-off polls with a Duktape call-back.

### DIFF
--- a/docs/source/userguide/scripting.rst
+++ b/docs/source/userguide/scripting.rst
@@ -1120,6 +1120,15 @@ OvmsPoller
 
 The Ovms Poller object represents the poller sub-system. It contains the following methods:
 
+- ``isregistered = OvmsPoller.RegisterBus(bus, mode, speed [,dbcfile])``: Register the can bus
+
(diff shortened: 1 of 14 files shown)


frogonwheels:
I will go far as to suggest that this patch series could (should?) be deferred till after a release.  I have some DBC duktape commands coming as well.

frogonwheels:
Rebased to master if we are ready to accept this now.

PR #1129, opened 2025-04-19
Boot crash fix

GreyPeter:
A boot crash was occurring when booting with a MG vehicle selected. Found that m_poll_bus_default was not being initialised.


dexterbg:
Peter, this doesn't look like the correct root cause fix to me.

`m_poll_bus_default` should normally only be changed through a call to `PollSetPidList()`. It's just an internal helper now, so you're of course free to directly manipulate it. But then, setting it to `m_can1` in the constructor before actually doing `RegisterCanBus()` for bus 1 doesn't make any sense, except as a null initialization.

That leads me to `m_poll_bus_default` _should_ actually be null initialized, but isn't. But that should rather be done generally by the `OvmsVehicle` constructor (within the `#ifdef CONFIG_OVMS_COMP_POLLER` block), not by the MG subclass.

Please change the  PR to include the `OvmsVehicle` constructor fix instead of the misleading `m_poll_bus_default = m_can1` line.

PR #1144, opened 2025-05-11
Nissan LEAF: improve v.p.speed accuracy and precision

zhongfu:
two changes:
- update divisor for calculating actual km/h speed value from can1 0x284 message
- cast uint16_t raw speed value to float before dividing (to avoid dropping the decimal part of the speed value)

---

the new divisor for calculating km/h speed value was validated by eyeballing the standard OBD-II PID for vehicle speed from Car Scanner and comparing it with a custom OVMS dashboard with `v.p.gpsspeed` and `v.p.speed` on a straight, flat road with cruise control on.

it *is* significantly lower (by approx 10%) compared to the speedometer speed, but this is likely due to Nissan (and many others) deliberately overstating speedo speed to comply with [UNECE Regulation No.39](https://unece.org/sites/default/files/2024-04/GRSG-127-12e.pdf) (see point 5.4 on page 8).

the vehicle itself is aware of the true speed as well (as evidenced by the OBD-II PID readings), so i believe it may be more reasonable to have `v.p.speed` reflect true speed instead of speedometer speed

this will also allow us to derive an accurate trip odometer value based on speed (like what's being done for the MG EVs) that matches the vehicle odometer

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/nl_web.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/nl_web.cpp
@@ -78,6 +78,7 @@ void OvmsVehicleNissanLeaf::WebCfgFeatures(PageEntry_t& p, PageContext_t& c)
   bool ze1;
   std::string modelyear;
   std::string cabintempoffset;
+  std::string speeddivisor;
   std::string maxgids;
   std::string newcarah;
   std::string cfg_ev_request_port;
@@ -86,13 +87,14 @@ void OvmsVehicleNissanLeaf::WebCfgFeatures(PageEntry_t& p, PageContext_t& c)
     // process form submission:
     modelyear           = c.getvar("modelyear");
     cabintempoffset     = c.getvar("cabintempoffset");
+    speeddivisor        = c.getvar("speeddivisor");
     cfg_ev_request_port = c.getvar("cfg_ev_request_port");
     maxgids             = c.getvar("maxgids");
     newcarah            = c.getvar("newcarah");
     socnewcar           = (c.getvar("socnewcar") == "yes");
     sohnewcar           = (c.getvar("sohnewcar") == "yes");
     canwrite            = (c.getvar("canwrite") == "yes");
-    ze1                = (c.getvar("ze1") == "yes");
+    ze1                 = (c.getvar("ze1") == "yes");
 
     // check:
     if (!modelyear.empty()) {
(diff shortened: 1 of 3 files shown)


zhongfu:
with that said, maybe it's a better idea to expose the divisor as a parameter, so users can adjust it to account for variations in tyre circumference if needed?

i *think* you might be able to adjust the speedo/odo by +/- 2.5% on LEAFs with satnavs (SV, SL), but it doesn't seem like anyone has documented how this works yet, so Leaf S users would be out of luck wrt changing this offset. i'm also not sure if this actually affects what's being reported on the speedometer/odometer or via the CAN buses, so having this be configurable could at least allow for more accurate trip reports

dexterbg:
Probably the value read isn't the effective speed but some RPM number? Rather than making the divisor a parameter, you should then make the tyre size the parameter, so users don't need to calibrate the value themselves.

I'll leave this open unless you want this to be merged already.

zhongfu:
> Probably the value read [is...] some RPM number?

That could be the case -- while the raw value doesn't match the actual motor RPM (based on my previous OBD-II logs), there would probably be a direct relation here anyway.

Unless, of course, we're able to test and see if the offset that can be configured through the satnav head unit affects this value in any way.

Mine's a Leaf S equivalent, so I'm not really able to test this, unfortunately 

> Rather than making the divisor a parameter, you should then make the tyre size the parameter, so users don't need to calibrate the value themselves.

Sounds good to me, though I'm not sure how to handle things like:

- people using tyres that are taller than stock (or with snow chains or whatever?)

- "internal" speedometer being slightly off from the factory (I've seen people complain about this on forums, to the tune of 1-2.5%)

- tyre inflation pressure? though I don't think this would affect the readings by any significant amount

Then users would have to set their rim size to weird sizes like 17.17", or configure their tyre specs too, or whatever. That's one reason I'd prefer an arbitrary configurable divisor instead (perhaps with the addition of some sort of formula or example values in the docs)

Nonetheless, I'd defer to you on this though

zhongfu:
Another possibility: add a page with a tool to estimate an ideal divisor value

- User opens tool
- User drives on flat, straight road with good view of the sky (for GPS), with cruise control enabled
- Tool compares speed reported by VCU (?) against GPS speed, or derived distance against GPS track distance...
- ...and gives the user a more appropriate divisor value

zhongfu:
for now i've made the speed divisor directly configurable for testing

PR #1145, opened 2025-05-11
Nissan LEAF: derive v.pos.trip from speed

zhongfu:
probably best used with #1144 for accurate trip odometer values

this is basically the same as what's being done for the MG EVs, and appears to be reasonably accurate (+/- 1%) when comparing to GPS tracks (recorded on a phone, dual-band GNSS, 1Hz, ~45km across 3 tracks). of course, this depends on having accurate speed measurements, hence #1144


zhongfu:
rebased onto current master

PR #1147, opened 2025-05-11
Nissan LEAF: fix charge time estimates for ZE1

zhongfu:
ZE1 doesn't seem to have `xnl.v.b.soc.instrument` available (defaults to 0%), which causes charge time estimates to always show time from 0% instead of current SoC

this change makes `calcMinutesRemaining()` use `xnl.v.b.soc.nominal` instead, if `cfg_ze1` is set to true


zhongfu:
actually, i see that the SoC limit is implemented in `HandleCharge()` by checking `v.b.soc` and not `xnl.v.b.soc.instrument`, so it might make more sense to calculate based on `v.b.soc` instead

additionally,
- on ZE1, `v.b.soc` ~= `xnl.v.b.soc.nominal`
- on other model years, `v.b.soc` = `xnl.v.b.soc.instrument`, unless `soc.newcar` is true, in which case it's calculated based on (current GIDs / max GIDs for a new battery)

just using `v.b.soc` would fix this bug, and possibly make charge times more accurate (if a limit is enabled)

though on ZE1, `v.b.soc` is the raw SoC of the battery pack (which will never hit 100%), so the full charge time estimate would probably be a bit off. probably same for non-ZE1 models as well, if `soc.newcar` is true?

dexterbg:
I'm not sure if you mean to keep this open to gather feedback from users. I'll merge it now. If you intend to keep a PR open, please tag it as "draft".

PR #1149, opened 2025-05-17
fix(ovms_cellular): update `m.net.good.sq` in `UpdateNetMetrics()`

zhongfu:
this fixes a bug where:
- `m.net.type` starts off as "none", before modem IP link is up
- modem manages to register, and we get an AT+CSQ response w/ the signal strength
- `modem::SetSignalQuality()` is called, but `m.net.sq`/`m.net.good.sq` is not updated because `m.net.type` is not "modem" (yet)
- modem IP link comes up, `m.net.type` gets set to "modem", `modem::UpdateNetMetrics()` gets called
- but `m.net.good.sq` is not updated if signal strength doesn't change (caused by `m_sq != newsq` check in `modem::SetSignalQuality()`)
- until signal strength or network type changes to "wifi", `m.net.good.sq` will remain undefined

in this scenario, if the signal strength stays the same for an extended period of time, then `m.net.good.sq` will remain undefined, causing Server v3 to not connect until the signal strength changes.

Server v2 is not affected because it does not have `m.net.good.sq` as a prerequisite for connecting

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
+++ b/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
@@ -1723,17 +1723,8 @@ void modem::SetSignalQuality(int newsq)
     {
     m_sq = newsq;
     ESP_LOGD(TAG, "Signal Quality is: %d (%d dBm)", m_sq, UnitConvert(sq, dbm, m_sq));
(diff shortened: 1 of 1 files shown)


zhongfu:
applied this change to my module, and Server v3 starts immediately after the network comes up -- or rather, it doesn't stay in the "Waiting for network connectivity" state for an unnecessarily long period of time

Issue #409, opened 2020-08-19
Nissan LEAF P312A DTC & dashboard warning light 


glynhudson:
I've been running OVMS in my 2014 Nissan LEAF for the past couple of years without any issues. 

Recently (possibly after upgrading to 3.2.014) I've started getting an EV warning light on the dashboard caused by DTC P312A. The error seems to happen when starting up the car for the 2nd time on a particular day without any charging taking place. The error does not seem to effect the operation of the car. 

I tried switching off CAN writes, but this did not have any effect. Error still occurred. 

After unplugging OVMS the error does not occur. I will try and testing rolling back to a previous OVMS release this week. Just wondering if anyone else has experienced the same issue?  

There have been a number of changes to the NL code in this release : https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/400 https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/399/ https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/397/ https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/394

What model of Nissan Leaf are you using to test these changes @drc38?

P312A is a CAN bus communication error:

![Screenshot from 2020-08-19 00-54-11](https://user-images.githubusercontent.com/758844/90576599-934fcf00-e1b6-11ea-9195-648f71a811cd.png)

EV warning light on the dashboard:

![Screenshot from 2020-08-19 01-55-02](https://user-images.githubusercontent.com/758844/90579832-1117d880-e1bf-11ea-85a9-5943000430c6.png)

Here is the DTC from Leafspy:

![Screenshot_20200818-233943](https://user-images.githubusercontent.com/758844/90576377-f856f500-e1b5-11ea-9768-69012a10d43b.png)

drc38:
Hi Glyn,
I'm using a 30kWh Leaf, the recent changes that affected can bus transmits were to the polling regime. Previously, all polls were being sent to can bus 2 (CAR can). This meant battery/temp/qc polls would fail when charging as the CAR can is asleep.  Initially, I switched polling to can bus 1 (EV can bus) as battery polls were more important than obtaining the vin etc. That said, polls are not frequent and I've never had an issue with a failed poll on an alternate bus causing an issue with the ovms module or car.

[...]

PR #1152, opened 2025-05-17
feat(nissanleaf): add v.b.12v.current metric (from polling VCM)

zhongfu:
as the title says, this PR adds a `v.b.12v.current` metric for measuring 12V battery current.

it doesn't seem like anyone has figured out if the VCM (which is connected directly to the 12V lead-acid battery current sensor) reports this periodically, so for now, we can only poll

potential improvements/issues:
- poll whenever car is awake or HV relay is closed, instead of polling only while in on/running/charging states
  - this includes when remote climate control and the daily 12V battery top up is happening
  - but it seems like we might only support 4 poll states max (`VEHICLE_POLL_NSTATES` in `vehicle_poller.h` is set to 4, not sure about the implications of increasing this)

- figure out if if it's safe to poll while running
  - I haven't had any issues so far on my ZE1, but #409 seems to imply this might be an issue on ZE0/AZE0 (which is probably why we also have a separate "running" poll state)
  - VCM is also connected on EV-CAN, perhaps it'd be safer to poll there (if even possible)?
 
- figure out if VCM (or e.g. DC-DC converter) is reporting this information on the CAN bus
  - would avoid most of the mess above, since we can simply passively listen and avoid polling

for now, this is a draft because i'll still have to test this (to make sure the resulting values seem reasonable), plus this isn't a great solution IMO

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/src/vehicle_nissanleaf.cpp
@@ -55,6 +55,7 @@ static const char *TAG = "v-nissanleaf";
 #define BROADCAST_RXID            0x0
 // other pairs 743/763 744/764 745/765 784/78C 792/793 79D/7BD
 #define VIN_PID                   0x81
+#define CURRENT_12V_PID           0x1183
 #define QC_COUNT_PID              0x1203
 #define L1L2_COUNT_PID            0x1205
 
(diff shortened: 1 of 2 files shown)


dexterbg:
While it's of course possible to increase the number of poll states supported, it seems a bit odd needing to do so, as all other vehicles do just fine with 4 states.

From a look into the source, it seems you currently waste one state. A comment in the module reads…
```
// Leaf does not respond to polls when car is off
// So there is no point polling when car is off
```
Yet you still have state slot 0 assigned to the "off" state, with all poll entry times being zero. If you can detect the vehicle "on" states without polling, you don't need an "off" polling state, but can instead stop the poller.

Regards,
Michael

zhongfu:
> While it's of course possible to increase the number of poll states supported, it seems a bit odd needing to do so, as all other vehicles do just fine with 4 states.
> 
> From a look into the source, it seems you currently waste one state. A comment in the module reads…
> ```
> // Leaf does not respond to polls when car is off
> // So there is no point polling when car is off
> ```
> Yet you still have state slot 0 assigned to the "off" state, with all poll entry times being zero. If you can detect the vehicle "on" states without polling, you don't need an "off" polling state, but can instead stop the poller.
> 
> Regards,
> Michael
> 

Ah yeah, that makes sense. I'll go check it out, thanks!

PR #1159, opened 2025-06-22
Add canbus hw listenonly

zbchristian:
@dexterbg Here now the pull request for the Hardware listen only mode. 
Even though its only of limited use, its helpful to check intereference problems.
For the Smart EQ it actually works pretty well in listen only mode. Only few informations are missing.


dexterbg:
Unless I'm getting something completely wrong, I think you've now mixed up the controller & transceiver controls, or separated them possibly unnecessarily.

We already have the active/listen mode control, this is given on the CAN bus `Start()` call. I don't see a need to control the transceiver indepentantly of the controller mode (unless for temporary technical test reasons), so I think both should always be in the same mode. What's probably only missing is setting the SN65 `RS` pin according to the controller mode, as the drivers previously assumed that's an on/off control.

Dynamic reconfiguration of the mode is also already supported by calling `Start()` again. You need to pass the other parameters, so an API method to change only the mode would be OK. But that should also change the controller mode. That's currently missing from your implementation: if you start the smart module in listen mode, changing the config will have no effect, as the controller will stay in listen mode.

zbchristian:
Yes, you are of course correct, that I ignore the state of the controller and to reuse the Start/Stop function make sense. 
Need to think about it a bit more and close this PR for now.

PR #1164, opened 2025-07-12
Set CAN transceiver active/listen-only mode according to CAN bus mode

zbchristian:
Enable/disable the CAN transceiver TX driver dependent on the CAN bus mode (CAN_MODE_ACTIVE, CAN_MODE_LISTEN) set by the Registration of the bus.
The RS-pin of the SN65HVD233 is set to the corresponding level low (ACTIVE) or high (LISTEN-Only). 

The code has been tested for ESP32CAN as well as MCP2515 based CAN buses.

In the vehicle specific code, the registration of the CAN bus should set the mode correctly, dependent on the write enable configuration:
````
CAN_mode_t mode = m_enable_write ? CAN_MODE_ACTIVE : CAN_MODE_LISTEN;
RegisterCanBus(can-no, mode, CAN_SPEED_500KBPS);
```` 
This code should be executed again, if the write access is changed in the configuration.

### DIFF
--- a/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
+++ b/vehicle/OVMS.V3/components/esp32can/src/esp32can.cpp
@@ -211,7 +211,7 @@ static IRAM_ATTR void ESP32CAN_isr(void *pvParameters)
         {
         // Freeze TEC/REC by entering listen only mode
         MODULE_ESP32CAN->MOD.B.LOM = 1;
-        MyESP32can->SetTransceiverMode(false);
+        MyESP32can->SetTransceiverMode(CAN_MODE_LISTEN);
 
         // Re-trigger bus-off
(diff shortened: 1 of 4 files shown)


dexterbg:
Looks mostly good to me.

Two comments:
- I'd rather have the `SetTransceiverMode()` API use `CAN_mode_t` instead of a bool. It's more readable, better reflects the purpose, and is future proof in case we some day upgrade the hardware to another transceiver type.
- Indentation style on `mcp2515::SetTransceiverMode()` isn't according to the module, if/else brackets need to be indented.

zbchristian:
Ok, changed the parameter of the function and fixed the format. 
Actual added a missing GPIO definition in case its not the standard hardware pin. 

The pull request for the feature to remap the GPIOs will come very soon.

PR #1167, opened 2025-07-14
Feature: GPIO remapping for alternative OVMS modules

zbchristian:
Implement an optional utilization of the OVMS firmware with other hardware modules.  These modules have to be based on the same or a subset of components, but require a modification wrt the ESP32 GPIO assignment.
This is achieved by the option, to load the actual GPIO assignment from an external file in `ovms_peripherals.h`.

There is no restriction on the location of the GPIO mapping file(s), but to simplify the detection of the entered file, a new folder  
 `components/gpio_maps` has been included. This folder is added to the include path via `component.mk` during the build process. In addition the folder contains the documentation of the required GPIO names and functions.

This PR  includes corresponding GPIO mapping files for different versions of the Lilygo T-Call A7670E in the folder `components/gpio_maps` and an example `sdkconfig `in the `support `folder.

@dexterbg, there are different options, how to incorporate different hardware modules into the OVMS firmware.
- This PR adds everything to the firmware repo
- Another option would be to just add the option to load an external GPIO mapping into this repo. This would effect only `Kconfig ` and `ovms_peripherals.*`. The actual instructions, mapping and sdkconfig files could be published in a separate repo. This would separate the firmware from the specifics for a certain hardware.

I am pretty ambivalent about this and could really understand, if you would prefer the second option.

I just realized, that `support/partion4M.csv` is changed by this PR as well. 
The version in the `master` branch has the problem, that the size of the factory partition is too small to flash the current edge version. In addition is the total size too small by 448kB.


dexterbg:
This is probably more something @markwj should approve.

My main issue would be with raising the firmware partition size, or rather, shrinking the `store` size to 512K. That wouldn't be compatible with the 16M standard module anymore, and `store` needs to accomodate for plugin storage and still have spare room for the backup & restore operations. 512K would be too small for my current setup, and probably for many other users as well.

Where does the additional firmware size come from for the Lilygo build? I thought that's mostly GPIO remapping?

zbchristian:
The size of the firmware does not depend on the Lilygo support. 
When  I build the default firmware, its larger than 3MB. In partition4M the size of the factory partition is defined too small for this. So, thats why I have too squeeze the partition definition for 4MB Flash.
You are correct, that the remaining space is very limited , but in the case of only 4MB Flash there are very few options.
This has no effect on the 16MB version, where the space for the firmware is anyway 4MB.
The other option would be to create a special firmware with reduced features for only 4MB flash.
Anyway, the partition4M file should be corrected, because the total size is short by 448kB.

dexterbg:
I see. The idea was to rather exclude features (e.g. reduce/partition vehicle support) when we exceed 4MB flash size than reduce the `/store` size. I think that's also a better option for the 4MB/Lilygo build now. You can easily create different `sdkconfig` templates for this purpose.

zbchristian:
Ok, I corrected the partition map now only to fit the full 4MB and left the store partition at 1MB.

Are you fine with the Lilygo specific parts of the PR?

dexterbg:
No objections, so I'll merge this now. Nice work, Christian!

PR #1174, opened 2025-07-27
smart EQ can mode switch fix

zorgms:


dexterbg:
Before using `ExecuteCommand()`, always check if there is a direct API call available that does the operation. In this case, you would simply call the `Reset()` method on the bus, i.e. `m_can1->Reset()`.

But… if the reset solves the issue, the even better option should be to first `Stop()` the bus when doing a mode reconfiguration, because `Reset()` simply does a `Stop() → Start()` sequence.

I think the `Stop` should be automatically be done by the CAN driver when an already started bus is started again for a mode change. I've just implemented & pushed that change, can you please test?

That should make this PR obsolete.

zorgms:
m_can1->Reset() was my first think, but this will not work:

> CXX build/vehicle_smarteq/src/vehicle_smarteq.o
> Python requirements from P:/OVMS/home/ZorG/esp-idf/requirements.txt are satisfied.
> P:/OVMS/home/ZorG/OVMS-Firmware-EDGE/vehicle/OVMS.V3/components/vehicle_smarteq/src/vehicle_smarteq.cpp: In member function 'virtual void OvmsVehicleSmartEQ::ConfigChanged(OvmsConfigParam*)':
> P:/OVMS/home/ZorG/OVMS-Firmware-EDGE/vehicle/OVMS.V3/components/vehicle_smarteq/src/vehicle_smarteq.cpp:403:17: error: 'class canbus' has no member named 'Restart'
>          m_can1->Restart(); 
>                  ^
> make[1]: *** [/home/ZorG/esp-idf/make/component_wrapper.mk:290: src/vehicle_smarteq.o] Fehler 1
> make: *** [P:/OVMS/home/ZorG/esp-idf/make/project.mk:561: component-vehicle_smarteq-build] Fehler 2

dexterbg:
It would work if you called `Reset()` instead of `Restart()`. But all of this should no longer be necessary with my latest change.

zorgms:
Works well, thank you

PR #1184, opened 2025-08-23
new metric: v.c.timestamp.start -> date/time at charge starting

zorgms:
v.c.timestamp.start -> date/time at charge starting
v.e.serv.level -> e.g. smart EQ: Service A
v.e.climate.timerdata -> Timebased Climatecontrol App Data

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
@@ -1031,8 +1031,7 @@ void OvmsServerV2::TransmitMsgStat(bool always)
     StandardMetrics.ms_v_bat_range_speed->IsModifiedAndClear(MyOvmsServerV2Modifier) |
     StandardMetrics.ms_v_charge_kwh_grid->IsModifiedAndClear(MyOvmsServerV2Modifier) |
     StandardMetrics.ms_v_charge_kwh_grid_total->IsModifiedAndClear(MyOvmsServerV2Modifier) |
-    StandardMetrics.ms_v_charge_timestamp->IsModifiedAndClear(MyOvmsServerV2Modifier) |
-    StandardMetrics.ms_v_charge_timestamp_start->IsModifiedAndClear(MyOvmsServerV2Modifier);
+    StandardMetrics.ms_v_charge_timestamp->IsModifiedAndClear(MyOvmsServerV2Modifier);
 
   // Quick exit if nothing modified
   if ((!always)&&(!modified)) return;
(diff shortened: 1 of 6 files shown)


dexterbg:
This needs some discussion, possibly refinement.

- Please explain the need to keep the charge start time additionally. A running charge's start time can be determined by subtracting the charge duration from the current time.
- If this metric makes sense: when introducing a new charge metric, it's normally necessary to also introduce the same metric for the generator operation (and add it to the MP `X` record), unless it's really a charge specific metric. The goal is trying to keep these two metric groups as similar as possible.
- The new `v.e.service.level` metric is commented as "Level of scheduled maintenance/service (bitfield, vehicle-specific)". Using a string for this seems appropriate, but that means it's not a "bitfield" but probably a list of code tokens. Instead of tagging this as vehicle specific, I'd like to see some standardized core set of service types/flags here, that will fit most vehicles. The App will need to translate these for the user, so the core set would cover all or most vehicles without additional code / string entries. Codes should be short but still readable, and usable as string identifiers in the App, for example `std` for standard and `ext` for extended service. Add these to the documentation, plus a scheme for custom codes and the separator character for multiple codes. It may make sense to ask for common service types of other vehicles on the developer list.
- Speaking of documentation: all changes to the MP record structures need to be documented in the user guide's V2 protocol section. And all user level extensions need to be mentioned briefly in the `changes.txt`.
- Speaking of MP record structure: the current gear (`ms_v_env_gear`) doesn't suit as part of the firmware message. That's clearly a candidate for the location message (`L`), as that's the streaming message meant to track vehicle movement.
- `ms_v_env_climate_timer_data` = "Climate timer data (json array of timers) (for SQ time based Climate control)" -- this clearly exceeds the scope of a metric. Metrics are meant to transport single dynamic state measurements and indications, these are meant to be atomic values, at most sets/vectors of such. Configuration records are outside the scope of metrics. Think of metrics as data items that can be displayed mostly directly without further processing. The timer mode scheme for charge/generator timers is meant to transport only the current (next due) timer mode entry, so it will inform the user about the timer mode schedule for the next day at any time. You could replicate this metrics scheme for the climate control timer mode, and name the metrics `v.e.cc.timermode` and `v.e.cc.timerstart`.
- The climate control timer mode metrics also do not suit as part of the MP "firmware" message, but clearly are candidates for the "environment" (`D`) record.
- If you still want to provide the full Smart specific JSON timer definition for the App, introduce a Smart specific command for this. The App then can issue the command and process the command output as needed, this allows you to query arbitrary custom record structures.

zorgms:
I need to think about whether I should only implement this for the smartEQ.
Inserting and setting v.c.timestamp at the end of charging will cause confusing charging time displays in the app for non-smart EQ/Ph2 vehicles, as v.c.timestamp is currently used by both vehicles as the basis for the elapsed charging time.

dexterbg:
The newly `v.c.timestamp` has been defined as the time at charge end by @Dimitrie78 (and it makes sense, giving the user a quick info on when the last charge was done). The elapsed charging time has always been available as `v.c.time`.

Issue #1175, opened 2025-07-27
Smart EQ periodically sending climate notifications after update to 3.3.005-1


dexterbg:
@zorgms / @zbchristian

A user just reported another issue/regression on the Smart EQ, occuring on the "main" release:

Since OTA firmware 3.3.005-1-geed58d6f was installed on my Smart ForTwo EQ, I am constantly getting messages from the server stating:
```
Climate values:
  on: no
  time: 5:15 h
  weekly: no
  start Day: Mon
  end Day: Fri
  runtime: 5 min
```
This appears to be giving me no useful information and using up my sim card data.  How do I stop this message?

If you need more details, contact me by mail to get the user's email.

zorgms:
This is a warning: If you use or have enabled time-controlled climate control, you'll see this message because the settings reset to default after rebooting.
You can enable/disable it on the Features website -> Diff Settings -> Enable Climate System

denewbery:
I don't know what you are referring to by the "Features website". I have been using the OVMS app on my phone and the dexters-web.de website.  Where is the "Features" website.

zorgms:
You can also change the setting via shell:
Log in to dexters-web.de
Click the shell button
Enter this:
`config set xsq climate.system no`
Then click send.

Or send it via the app via message page:
`config set xsq climate.system no`

More information about smart EQ app topics can also be found here:
https://www.smart-emotion.de/forum/board/79-ovms-unabh%C3%A4ngiger-appzugang/

denewbery:
Thanks – I’ve done that and ‘ll see if it works.

Thank you.

 

Regards

 

David Newbery

zorgms:
By the way, you're having a problem with the module repeatedly restarting. You should fix this. 
Perhaps your firmware is outdated. An update to firmware 3.3.005-1-geed58d6f (https://ovms.dexters-web.de/firmware/ota/v3.3/main/ovms3.bin) would be advisable.

denewbery:
Hello again.

 

The problem started when I installed firmware 3.3.005-1-geed58d6f.  However, it seems to have sorted itself out now.  I’ll get back to you if I have any more problems.

 

Regards

 

David Newbery

dexterbg:
ZorG, I've got the next user being irritated about this.

This is really messy, here's an excerpt of his notification log:
```
[...]

PR #1185, opened 2025-08-25
smart EQ: disabled climate/heater start timer

zorgms:
Automatic shutdown after two crash boot

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,12 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- smart EQ:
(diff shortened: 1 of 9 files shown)


dexterbg:
Audit results:
- Automatically switching from wifi apclient mode to client mode after a configurable time after boot is no vehicle specific function. Please rework this into a generalized feature of the esp32wifi module and post as a separate PR. Please also solve the following issues with your current implementation while doing so:
  - Switching to client mode must not be done while a station is connected to the AP. Ideally the timeout would restart when the last station has disconnected.
  - Auto switching to client mode also must not be done if the wifi mode isn't `apclient` or has been manually altered by the user in any way.

- Regarding your crash count check, you need be clear about reacting not on the _total_ crash count but on the _early_ crash count -- that is, if that's what you actually intend to do. On the example user from issue #1175 as I wrote most were not early crashes. Just checked again: currently 8 of 110 crashes recorded for that user were early crashes. I also wouldn't expect these to be early by nature, if they correlate to the apclient mode.

Regards,
Michael

zorgms:
Ok, thanks for feedback.

PR #1186, opened 2025-08-30
esp32wifi switch APclient to Client mode

zorgms:
esp32wifi:
The OVMS will switch from Wifi APClient to Wifi Client mode after the timeout reached and a client is not connected. "APClient" will be restored upon reboot/network restart. new config:
- [network] ap2client.timeout   -- Wifi Mode APClient to client timeout in minutes (default 30 minutes), settings at Wifi
- [network] ap2client.enable    -- Wifi Mode APClient to client enable/disable (default enable), settings at Wifi

smart EQ:
- added auto disable time based climate control when crash counter > 1 to stop Notify spamming
- deactivated all Web slider checkboxes

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,11 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- esp32wifi:
+  The OVMS will switch from Wifi APClient to Wifi Client mode after the timeout reached and a client is not connected. "APClient" will be restored upon reboot/network restart.
+  new config:
+    [network] ap2client.timeout   -- Wifi Mode APClient to client timeout in minutes (default 30 minutes), settings at Wifi
+    [network] ap2client.enable    -- Wifi Mode APClient to client enable/disable (default enable), settings at Wifi
 - Vehicle framework: signal generator operation state changes
   New events:
     vehicle.gen.start         -- Power delivery started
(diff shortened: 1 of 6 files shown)


dexterbg:
A good idea, and well done on the integration & web UI.

A general comment first… ZorG, you really need to change your git commit habits to enable cleanly separated & easily revertable pull requests.

A PR shall not mix vehicle specific and non-specific changes. A PR generally shall not mix different functional changes, unless they depend on each other. That's easily achievable with git, as you can select which parts of your open changes you want to include in a commit, down to selecting single lines. If using git gui, check the context menu in the diff view. This works both ways, if you only need to exclude some lines, add the full file to the commit staging, then you can exclude blocks and lines from within the diff view of the added file.

Once you have functionally isolated commits, you can easily create new PR branches and cherry-pick only the commits relevant for the PR to that branch. That way the PR also gets updated automatically when you add more commits to that branch.

Now for the audit results on the actual PR:
- Please change the config variable names to using the prefix `wifi.`, following the established standard within the `network` config area, i.e. `ap2client.timeout` → `wifi.ap2client.timeout`.
- Config `wifi.ap2client.enable` then needs to default to `false`. When introducing new configs that change the behaviour, they shall normally be initialized in a way to keep the previous behaviour unless the user takes action. The only exception to this rule is if the change is necessary to keep the system usable for most users, e.g. due to some external requirements. In this specific case, users of external devices coupled to the OVMS via Wifi, like the WifiConsole (https://github.com/dexterbg/WifiConsole) or generally some Wifi button or display, would experience a seemingly broken system after updating their OVMS firmware, if the new config defaults to `true`.
- Minor: indentation and bracket style of `esp32wifi::AP2ClientCheck()` and `esp32wifi::AP2ClientSwitch()` needs to be changed to that of the `esp32wifi` module.

Regards,
Michael

dexterbg:
Maybe I wasn't clear. Keep the config variables within `network`, just add the `wifi.` prefix to the names to clarify they are wifi related. There is no `wifi` config section, and we shouldn't introduce one.

Issue #1191, opened 2025-09-04
New idea: NetLoss loop fix


zorgms:
New idea:

I activated the data connection in Netwait, which allows the modem to reconnect very quickly.

```
    case NetWait:
      MyEvents.SignalEvent("system.modem.netwait", NULL);
      muxtx(m_mux_channel_POLL, "AT+CGATT=1\r\n");
      if (GPS_SHALL_START())
        StartNMEA();
      break;
```

```

I (705866) cellular: State: Enter NetLoss state
I (705866) cellular: NetLoss: PPP connected - detaching PDP and stopping PPP
OVMS# cellular setstate NetLoss
Set modem to state: NetLoss

D (705866) events: Signal(system.modem.netloss)
D (705866) cellular: mux-tx #3: AT+CGATT=0
I (705866) gsm-ppp: Shutting down (hard)...
I (705876) gsm-ppp: StatusCallBack: User Interrupt
I (705876) gsm-ppp: PPP is shutdown
D (705876) events: Signal(system.modem.down)
I (705876) netmanager: Set DNS#1 0.0.0.0
I (705876) netmanager: Set DNS#2 0.0.0.0
I (705876) netmanager: MODEM down (with WIFI client up): staying with WIFI client priority
D (705876) netmanager: send cmd 3 from 0x0
D (705876) netmanager: done cmd 3 from 0x0
D (705876) events: Signal(network.modem.down)
D (705876) events: Signal(network.interface.change)
D (705886) netmanager: MongooseTask: got cmd 3 from 0x0
D (705906) netmanager: CleanupConnections: conn 3f8e89a8 -> iface st1
D (705916) netmanager: CleanupConnections: conn 3f8e9e9c -> iface st1
D (705916) netmanager: CleanupConnections: conn 3f8dd494 -> iface st1
D (705916) netmanager: MongooseTask: done cmd 3 from 0x0
D (715256) cellular: State timeout NetLoss => NetWait
I (715256) cellular: State: Enter NetWait state
D (715256) cellular: mux-tx #3: AT+CGATT=1
I (715256) gsm-nmea: Startup
D (715256) cellular: mux-tx #1: AT+CGPS=0
D (715256) events: Signal(system.modem.netwait)
D (717256) cellular: mux-tx #1: AT+CGPSNMEA=258
D (717276) cellular: mux-tx #1: AT+CGPSINFOCFG=5,258
[...]

PR #1189, opened 2025-09-01
cellular -> NetLoss loop fix

zorgms:
When the modem detects a NetLoss, it usually enters a continuous loop of NetLoss, NetWait and NetStart when a Wi-Fi connection is used. The problem can be simulated with ‘cellular setstate NetLoss’. The fix provides that when a NetLoss loop is detected, the modem then performs a PowerOffON, which reliably solves the problem.

PR #1192, opened 2025-09-04
NetLoss fix update

zorgms:
The fix in #1189 sometimes causes the module to crash. This extended fix works perfectly on my modem.

#1191


dexterbg:
LGTM, in fact speeds up the reconnect from ~23 to ~13 seconds on my modem with an artificial net loss.

I cannot easily test a real net loss here, as the next cell tower is just 150m away, but I don't see why this wouldn't work there as well.

PR #1194, opened 2025-09-05
server.v3 - add configurable connection delay and jitter to server

zorgms:
Introduces new parameters for connection stability wait time and random jitter before connecting to the server. These can be configured via 'server.v3' config, improving robustness against network flapping and reducing simultaneous reconnect storms. Updates logic in Ticker1 to use these new parameters and adds related member variables.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,10 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Server V3: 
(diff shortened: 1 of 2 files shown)


dexterbg:
Missing: new config variables need to be documented and briefly explained in the `changes.txt` (and in the user manual if that has a suitable section yet). Even if they don't get added to the web UI config pages (which is perfectly OK here, shouldn't normally need tweaking), these still are changes users should be able to notice and/or look up.

A minor note: please take care to replicate the exact bracket/indentation style of the module, lines 933…938 seem to be indented by one additional level.

PR #1195, opened 2025-09-07
server V3: fixed job queue overflow

zorgms:
When server V3 requested all metrics, the job ends with queue overflow and drop data.

Split transmission to avoid MQTT/job queue congestion (job queue overflow issue): We send a maximum of MAX_PER_CALL metrics per call. The remaining metrics are sent on subsequent scheduler ticks (ticker1 triggers this again). This prevents large bursts that could overflow internal queues.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v3/src/ovms_server_v3.cpp
@@ -83,7 +83,7 @@ static void OvmsServerV3MongooseCallback(struct mg_connection *nc, int ev, void 
         opts.will_topic = MyOvmsServerV3->m_will_topic.c_str();
         opts.will_message = "no";
         opts.flags |= MG_MQTT_WILL_RETAIN;
-	opts.keep_alive = MyOvmsServerV3->m_updatetime_keepalive;
+	      opts.keep_alive = MyOvmsServerV3->m_updatetime_keepalive;
         mg_set_protocol_mqtt(nc);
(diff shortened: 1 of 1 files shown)


dexterbg:
Keeping a local metrics pointer is a crash recipe, as the metrics list is dynamic and metrics can be added and removed any time, potentially invalidating the pointer.

As a safe alternative you can keep track of the metric index position last transmitted. The websocket metrics transmission does a similar chunking, see `WebSocketHandler::ProcessTxJob()` for reference.

Explaining this now reminds me we should actually introduce a lock for the metrics list to protect list walkthroughs against concurrent metrics additions & removals -- that applies to all metrics loops, but is outside the scope of this change. I'll take a note. Still even with a lock a local metrics pointer is a no go, unless the module keeping the pointer is the one governing the registration/deregistration of that metric.

The chunking also should be applied to `TransmitModifiedMetrics()`, as the set of modified metrics can exceed the chunk size as well. You're also indicating a wrong variable use here:
```
    // Force next modified-metric cycle soon:
    m_lasttx = 0;
```
This is within `TransmitAllMetrics()`, so should set `m_lasttx_sendall` instead. `m_lasttx` would be the trigger for `TransmitModifiedMetrics()`.

zorgms:
All the different update times listed as default in the web UI don't correspond to the times in the code. I'm a bit confused right now. I'll correct this first and rethink the update routines accordingly. I don't think so many different routines are needed; V2 can manage with just three different times. Splitting them up for larger packages should make this work without any issues.

zorgms:
I will open a new PR, it is too confusing to bring the current changes here

PR #1196, opened 2025-09-07
cellular: fixed GPS on/off reactivate

zorgms:
- Fixed incorrect variable names, which prevented the GPS reactivation system from working.
- Fixed incorrect order in setting variables and generating messages.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
+++ b/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
@@ -321,7 +321,7 @@ modem::modem(const char* name, uart_port_t uartnum, int baud, int rxpin, int txp
   m_gps_usermode = GUM_DEFAULT;
   m_gps_parkpause = 0;
   m_gps_stopticker = 0;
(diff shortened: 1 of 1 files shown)


dexterbg:
In line 324 you start the `m_gps_startticker` unconditionally on the initial modem instantiation, I doubt that's correct.

Regarding code style, please split long lines in readable separate lines & keep lines split that way.

zorgms:
If GPS is turned off, the react function is no longer executed after a reboot, because the start ticker is always 0. The query as to whether the GPS is allowed to start is present in Ticker1.

Btw, with 2k resolution and 32", there is plenty of space in width for reading.

dexterbg:
OK… but the value of the `gps.parkreactivate` config is per `ConfigChanged()` already mirrored in `m_gps_reactivate`, so you can use that variable in the ticker for better readability and performance.

Thanks for reformatting. Code styling & formatting generally has two main goals, readability and diffability (or better: mergability). Both need short & semantically adjacent lines. Three-way merging works best with three line synchronized views (base, local, remote), having short lines with sensible breaks is crucial for this.

PR #1197, opened 2025-09-07
smart EQ: removed GPS on/off at parking

zorgms:
- Removed vehicle-specific system for GPS on/off when parking
- If this function was enabled, Cellular GPS parking will be configured and activated


dexterbg:
I assume this shouldn't be merged before PR #1196?

PR #1199, opened 2025-09-11
server V3: fixed job queue overflow

zorgms:
- Fixed job queue overflow
- Limited to 20 readings per call
- Prioritized optional GPS tracking readings
- Optimized Ticker1
- Disabled some interval times/conditions in Ticker1; there are simply too many without use

based on PR is #1195


dexterbg:
LGTM, except for keeping the removed interval variables & code: if the new scheme fully replaces the old, you should remove the now dead code accordingly. Nothing is lost, you can always restore code parts from previous revisions.

But you can do so later on, I think this can already be tested in the current state, so I'll merge this now.

PR #1202, opened 2025-09-18
Server V3 On-Demand Metrics and Configuration

zorgms:
Implemented support for client-initiated requests to fetch specific metrics or config values via new MQTT topics, reducing baseline data usage. Increased MQTT subscription topic count to 4 to accommodate the new request channels. Added minimal pattern matching for metric requests and ensured normal delta sending remains unaffected. Updated chunk sizes for metric transmissions and improved update interval logic.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,16 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- server V3 (additions):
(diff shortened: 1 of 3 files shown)


zorgms:
Sorry, but #1204 and #1202 overlap in some places. If the changes in #1204 are approved, then this should be rejected. Thanks for your understanding.

zorgms:
The revision in #1204 makes more sense than merging this one. The update routine and the web server overflow are still present here, so data is repeatedly lost during "sendall."

PR #1208, opened 2025-09-26
Add GPS auto start on vehicle awake option

zorgms:
Introduces a new configuration option to power on GPS for a set time when the vehicle awakes from parking, reducing time to first GPS fix. Updates the modem logic, configuration handling, and web UI to support the new [modem] enable.gps.awake setting.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -4,6 +4,9 @@ Open Vehicle Monitor System v3 - Change log
 - Cellular: GPS auto pause on parking re-activate when Car awakes. This reduces time to first GPS fix.
   New config:
     [modem] enable.gps.awake   -- GPS powered on for park re-activate time when parked and Car awakes (default: false)
+- Cellular: GPS auto pause on parking reactivate when Car awakes. This reduces time to first GPS fix.
+  New config:    
+    [modem] gps.parkreactawake (bool)   -- GPS is switched on for the gps.parkreactlock (minutes) time when the GPS parking pause is active and the car wakes up (default: no)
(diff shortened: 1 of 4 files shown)


dexterbg:
Basically OK, except for:
- As this only works in combination with an enabled GPS park pause, the config naming should reflect that. I.e. the config should follow the naming scheme already introduced for the park pause variables, and rather be named `gps.parkreactivate.awake` or similar instead of the general `enable.gps.awake`.
- Accordingly the web UI should also place the config below the park pause input, as it's applicability depends on that.
- Setting the stop ticker to `m_gps_parkpause` is not correct, following the park pause logic that should rather be `m_gps_reactlock * 60`, shouldn't it?
- …and the web UI's info on the config variable should be updated accordingly.

Btw, it seems the previously added park pause config variable names in the `changes.txt` need to be updated.

Regards,
Michael

zorgms:
The stop ticker is overwritten by the GPS lock anyway, so this should be fine. When the car is started, both tickers are reset (m_gps_startticker / m_gps_stopticker). Or am I making a mistake?

I'll do the reordering right away.

dexterbg:
The question is, which stop timeout do you want to apply in case the car cannot get a GPS lock (e.g. when parked in some underground garage). I'd rather apply the reactlock timeout, as the parkpause timeout will normally be set much higher.

`enable.gps.awake` is still in the `changes.txt`.

Another issue I didn't see: in the `changes.txt`, please have the config value units made clear from the description, i.e. "in minutes, default 5" for `gps.parkreactlock` and "bool, default no/false" for `gps.parkreactawake`

PR #1209, opened 2025-09-29
Remove m.time.utc from priority GPS metrics

zorgms:
The metric m.time.utc has been removed from the default priority GPS metrics list in both the server and web configuration. Documentation and UI text have been updated accordingly to reflect this change.

I only put it in for testing and forgot to take it out again.


zorgms:
oh no, that was right, sorry

PR #1212, opened 2025-10-03
Enhance Smart EQ: ticker1 budget, ADC recalibration, CAN parsing

zorgms:
- Added ticker1 time budget guard to cap per-tick workload, with new config option. 
- Introduced automatic 12V ADC factor recalculation using BMS 12V voltage, with adjustable offset and web UI controls. 
- Improved CAN poll reply handling with stricter length checks, corrected switch fallthrough, and added new vehicle state and underhood metrics. 
- Removed unused/duplicate poll PIDs and raw metrics, and updated web configuration for new features.


zorgms:
I need to look into this more closely; unfortunately, I haven't noticed it in the last few days.

PR #1219, opened 2025-10-09
Cellular config: add network type selection

zbchristian:
Add the option to select the mobile network mode (e.g. 2G, 4G) to the cellular configuration.
This allows to switch off the automatic selection in case of frequent net losses.

This has been tested for the SIMCOM 7670E and should be checked for the other models.

Two more fixes:
- correct the AT command to read the SIM ICCID:  `AT+CICCID` instead of `AT+ICCID`.
- GPS was switched off when parked, even, when the feature was disabled. Missing check has been added.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
+++ b/vehicle/OVMS.V3/components/ovms_cellular/src/ovms_cellular.cpp
@@ -924,11 +924,14 @@ modem::modem_state1_t modem::State1Ticker1()
         }
       switch (m_state1_ticker)
         {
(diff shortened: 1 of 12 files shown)


dexterbg (changes requested):
Christian, nice catch with the ICCID command, wasn't detected before because the "AT+ICCID" variant interestingly also works on the standard models SIMCOM 5360 & 7600.

I never needed to tweak the network type selection, but apparently some regions still have coverage issues. The feature is almost 100% well implemented, with just one issue to be addressed.

Regards,
Michael

PR #1222, opened 2025-10-13
Refactor smart EQ polling and wakeup logic

zorgms:
Moved poller definitions to a new header file eq_poller.h and updated references. Improved wakeup command logic and CAN bus awake/sleep handling. Enhanced ADC factor calculation with median and trimmed mean. Added new poll replies for DCDC activation and mains phase frequency. Removed obsolete modem network type handling and related web config options.


zorgms:
pollstate handling needed more tests

zorgms:
There's a bigger problem, so I'm withdrawing this PR.

PR #1229, opened 2025-10-25
Refactor config preset and cleanup smart EQ module

zorgms:
Introduces a new 'preset' command to apply and clean up smart EQ configuration defaults, consolidating and removing legacy config logic from the constructor. Removes the 'restart.wakeup' feature and related web UI, updates config handling, and improves code formatting and structure for maintainability.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_commands.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_commands.cpp
@@ -859,8 +859,8 @@ void OvmsVehicleSmartEQ::xsq_preset(int verbosity, OvmsWriter* writer, OvmsComma
 }
 
 OvmsVehicle::vehicle_command_t OvmsVehicleSmartEQ::CommandPreset(int verbosity, OvmsWriter* writer) {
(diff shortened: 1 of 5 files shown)


dexterbg:
ZorG, are you sure you want to actually set fixed config values from the `preset` command?

I'd recommend deleting the instances instead, that way you gain falling back to the actual defaults as coded, so defaults can evolve with the code.

Side note: you don't need to set/delete all instances individually, and it's actually also bad for performance and flash wear. If you know you're going to change potentially many instances, use `OvmsConfigParam::GetMap()` / `::SetMap()` or `OvmsConfig::GetParamMap()` / `::SetParamMap()` instead, so the whole config store update operation becomes atomic.

You can find examples in the web cfg source, `ConfigParamMap` is essentially a standard key/value map.

zorgms:
Thx for your input.

dexterbg:
ZorG, this isn't correct yet: you need to copy the `GetMap()` result to a local map, modify that, then pass the modified map to the setter. `GetMap()` does not return a copy of the internal map, it returns a pointer to the internal map. When passing the internal map to the setter, you will completely erase the map, as the setter does a clear on the internal map first.

dexterbg:
As an alternative, you can use `GetParamMap()`, that call returns a copy of the map.

PR #1231, opened 2025-10-29
Update vehicle_thinkcity.cpp

MylesTwete:
Added Case for TPMS sensor data handling


dexterbg:
Myles, welcome :-)

If you want to inform users about new user level features, add an entry to the `changes.txt` as well.

PR #1234, opened 2025-11-02
Add erase inactive OTA partition feature to firmware page

zorgms:
Introduces a new action and UI button to allow users to erase the inactive OTA partition from the firmware configuration web page. This helps free up flash space and provides confirmation dialogs and error handling for the operation.

ota_erase_1

ota_erase_2

ota_erase_3


dexterbg:
ZorG, this function doesn't make much sense for normal users, it will probably lead to users making mistakes and needing more support, and your explanation & info text indicates a basic misunderstanding regarding the underlying function:

Erasing an OTA partition does not free up any flash space. The flash partition layout is fixed (see `partitions.csv`), an empty OTA partition will not be available for anything else than another firmware image.

The `ota erase` command is meant to remove some temporary/experimental/rescue firmware image the user shall not be able to reactivate. This is a rarely needed developer/support feature, not meant for normal users.

The web UI should remain focused primarly on normal users. A developer/supporter (or power user) knows `ota erase`, so the standard web UI should not make this more accessable.

Regards,
Michael

PR #1235, opened 2025-11-03
Improve Renault Zoe Ph2 integration

CrashOverride2:
- remove unncessary variables
- remove Waterpump Counter and Reset (Ph2 doesnt need it)
- Use TXCallback for reliably sending commands instead of blindly sending message multiple times (especially for lock/unlock commands to reduce wear on solenoids and for PTC activation and deactivation)
- add manual PTC activation (heating while charging or defective compressor)
- add manual lighting activation (headlights for 30s like key fob button)
- add coming home function
- add remote climate by key fob
- add simple pre-climate scheduler (console/webgui)
- more detailed documentation as per user request
- readd unlock trunk command (this time its working)
- add manual dcdc activation and keepalive
- add automatic dcdc activation if battery voltage threshold is reached

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,18 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Renault Zoe Ph2:
+    Remove waterpump lifetime left metric and waterpump reset command
+    Add manual PTC activation (heating while charging or defective compressor)
+    Add lighting command (same like key fob button)
+    Add coming home function 
+    Add remote pre-climate with key fob
+    Add simple pre-climate scheduler
+    Add dcdc manual activation
+    Add automatic dcdc activation if battery threshold reached
+    Add working trunk unlock command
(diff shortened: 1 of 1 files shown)


dexterbg:
Carsten, I seem to have missed this before: it seems your unlock operations do not actually need the PIN to work, the parameter isn't used.

In that case, you really need to use the framework's PIN validation and make the PIN code mandatory. Currently anyone getting into the driver's mobile phone or getting command access in any other way can unlock the car.

The framework supplies the protected PIN password config (Config → Vehicle / config password pin) for cases like these. To validate a PIN entry agains this, use `PinCheck(pin)`.

Sorry for not seeing this in the first place. I assume you'd like to fix this before the merge?

And please also add an entry to the user history `changes.txt` briefly outlining the new user level features.

Regards,
Michael

CrashOverride2:
Oops, of course I'll add the PIN check — I didn’t even notice that it wasn’t working. I tested it today, and yes, it works without it :-D. 

I’ll add the code and test it first to make sure everything runs smoothly and also update the changes.txt file.

CrashOverride2:
It’s now working with the correct PIN only. 😃

PR #1237, opened 2025-11-08
Add new scheduled pre-climate feature for smart EQ

zorgms:
Introduces scheduled pre-climate functionality, ported from Renault Zoe Ph2, including new shell commands, config namespace, and web UI for managing schedules. Adds logic for parsing and triggering pre-climate events, updates configuration handling, and integrates with the vehicle's climate control system. Also includes related UI and config changes to support multiple times per day, per-weekday scheduling, and SoC checks.


dexterbg:
ZorG, please read my suggestion in issue #1239.

I think it would be best not to introduce the specific implementation if the generalization is around the corner.

zorgms:
I think a global implementation faces some hurdles.

This #PR also includes modifications to the existing system, which is also used via the Open Vehicle Android app.

PR #1241, opened 2025-11-09
update smart EQ: climate control duration and data storage options

zorgms:
Update for the old EQ schedule precondition system:
Introduces support for climate control activation with configurable duration and scheduled preconditioning. Adds options to store climate data in flash, updates web configuration UI, and refactors related logic for improved flexibility and reliability.


dexterbg (commented):
It seems the `ParseScheduleTime()` method is no longer needed as well?

zorgms:
No longer needed, it was part of the Zoe Ph2 precondition code.

dexterbg:
Why did you re-include the `OvmsVehicleSmartEQ::WebCfgClimateSchedule()` method? You should now be able to simply use `OvmsVehicle::WebCfgClimateSchedule()` instead, or are there any vehicle specific changes necessary?

The diff says there are none… and it says I missed another missing config rename on the day instances… about to fix that.

zorgms:
Sorry, I was a bit confused yesterday, I've corrected it now.

PR #1245, opened 2025-11-11
Move climate precondition scheduling to VehicleTicker1

zorgms:
Climate precondition scheduling logic was relocated from Ticker60 to VehicleTicker1 for improved timing and control. Log messages were updated for clarity, and redundant code was removed from Ticker60.

I don't see why, but at Ticker60 will not work.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -1315,7 +1315,7 @@ void OvmsVehicle::VehicleTicker1(std::string event, void* data)
       CommandClimateControl(true);
       ESP_LOGI(TAG,"Restarting climate control as per schedule");
       }
-    }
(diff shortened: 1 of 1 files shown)


zorgms:
Okay, the problem was overwriting Ticker60 in EQ; I think it's better if that's implemented in VehicleTicker1.

dexterbg:
I don't like the unnecessary processing introduced with checking the scheduler config once per second. It's OK to move the code outside of `Ticker60()`, as the ticker methods always have been meant to be overridden by a vehicle, so calling the super method wasn't required before.

But doing the checks and activation once per second still isn't necessary, once per minute is sufficient. So please encapsulate the part in an `if ((m_ticker % 60) == 0) { … }` branch.

zorgms:
It's in an `if ((m_ticker % 60) == 0) { … }` section.

Starting line 1237 - 1318. Should this be in its own section?

PR #1250, opened 2025-11-15
smartEQ: Remove climate control features and related code

zorgms:
Eliminated all climate control timer, data, and web configuration code from the smartEQ vehicle module. This includes removal of climate-related metrics, commands, web pages, feature handling, and periodic tasks, simplifying the codebase and configuration options.


dexterbg:
```
…/vehicle_smarteq.cpp: In member function 'virtual void OvmsVehicleSmartEQ::ConfigChanged(OvmsConfigParam*)':
…/vehicle_smarteq.cpp:323:10: warning: variable 'getString' set but not used [-Wunused-but-set-variable]
     auto getString = [&xsqcfg](const char* key, const char* def) -> std::string {
          ^
```

PR #1261, opened 2025-11-16
Add esp32egpio module - control ESPs GPIOS through EGPIO framework

CrashOverride2:
Add a new esp32egpio module to control the ESP32’s native GPIOs using the existing, well-integrated, and easy-to-use EGPIO framework.

This is primarily intended for external OVMS-based modules and serves as an addendum to @zbchristian’s work.

It has been running flawlessly to control an additional heater in a friend’s car, and I thought it might be useful for the project as well.


dexterbg:
LGTM. Only thing I'd recommend is to assert at most one of the two EGPIO modules is enabled, to assist users configuring their own build. I think the `Kconfig` syntax provides mutual exclusion (IIRC), another option could be to change the type of EGPIO subsystem to a selection of none / MAX7317 / native or similar.

CrashOverride2:
The idea of switching the GPIO framework is an even better one. I will definitely take a look at it so that it becomes easier and more logical for the user to configure.

PR #1262, opened 2025-11-17
Poller Handle  VWTP Pings

frogonwheels:
* Expands error code to signed integer in callbacks
* Doesn't throttle once-off polls
* Use ms timer to handle poll time-outs

Questions:
* I have it as 1s, adding a 1/2 s in extension. Not sure this is correct.


dexterbg:
> Questions:
> 
> * I have it as 1s, adding a 1/2 s in extension. Not sure this is correct.

From my tests with my Mii it seems to be sufficient, but that needs to be verified on other vehicles. I *think* it shoul be sufficient, I think the 2 second timeout was originally chosen rather because of the 1 second ticker resolution than to accomodate actual VCUs being so slow to respond.

dexterbg:
Anything left here after merging PRs #1305 & #1306?

frogonwheels:
Yep. One more

frogonwheels:
Handle the last few things - the VWTP Ping responses.

PR #1263, opened 2025-11-19
Update vehicle_thinkcity.cpp

MylesTwete:
TPMS Status (0x345) Case added.  This addition uses Standard Metrics for TPMS Alerts and Warnings to allow TPMS Sensor Status data to be reflected in the TPMS GUI of the vehicle App as well as in the notifications.  These changes were included in a local build, tested on 2 THINK City vehicles and functionality verified.   
Screenshot_20251119_123257_Open Vehicle
New Behavior:  On starting to drive, each tire in the TPMS GUI now shows a check mark if the tire sensor is confirmed working and tire pressure not very low.  Otherwise, that tire label comes up RED and with an Alert triangle.  
Screenshot_20251118_171004_Open Vehicle

After about a minute when first pressures come in, the triangles and check marks go away and are replaced by pressures.  
Screenshot_20251118_171037_Google

The color assignments remain as long as alert conditions persist.  Testing confirmed the RED alert goes away a few minutes after correcting a low pressure condition. 

Remaining issue: Left vs Right tire data always or often appears switched, however this is likely due to the Automatic TPMS Learning incorporated with THINK City cars.  Direct TPMS CAN data collection has shown that either the L/R tire order in the data is flipped or that it could go either way when the TPMS sensors are automatically "learned".  Until we know for sure which it is, no sense in changing anything.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
@@ -239,7 +239,44 @@ void OvmsVehicleThinkCity::IncomingFrameCan1(CAN_frame_t* p_frame)
 		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(MS_V_TPMS_IDX_RL, (float)d[3] * 0.3625, PSI);
 		    }
 		  break;
-
+    case 0x345: // TPMS Status Alerts and Warnings
+      {
+        std::vector<short> tpms_alert(4);
+        uint8_t tpms_warning;
+        
+        for (int j = 0; j < 4; j++)
+        {
+          tpms_warning = 0;
+          
+          if (j == MS_V_TPMS_IDX_FL)  // Front-left tire
+            tpms_warning = (d[0] & 0x0E) >> 1;  // Extract LF warning bits
+          else if (j == MS_V_TPMS_IDX_FR) // Front-right tire
+            tpms_warning = (d[1] & 0xE0) >> 5;  // Extract RF warning bits
+          else if (j == MS_V_TPMS_IDX_RR)  // Rear-right tire
+            tpms_warning = (d[1] & 0x0E) >> 1;  // Extract RR warning bits
+          else if (j ==MS_V_TPMS_IDX_RL)  //Rear-left tire
+            tpms_warning = (d[2] & 0xE0) >> 5;  // Extract LR warning bits
+          
+          if ((tpms_warning == 2) || (tpms_warning == 7)) // Very Low (Red) or missing sensor
+          {
+            tpms_alert[j] = 2; // "ALERT"
+          }
+          else if ((tpms_warning == 5) || (tpms_warning == 0)) // Fast deflation or Unknown State
+          {
+            tpms_alert[j] = 1; // "WARN"
+          }
+          else if (tpms_warning == 6) // Normal Tire Pressure
+          {
+            tpms_alert[j] = 0; // "OK"
+          }
+          else // Undefined States
+          {
+            tpms_alert[j] = 0; // Mark as undefined/unknown
+          }
+        }
+        StandardMetrics.ms_v_tpms_alert->SetValue(tpms_alert); // Use SetElemValues for vector
+      }
+      break;
     case 0x511:
       // Enerdel battery pack?
       // TODO: Enable battery pack cells monitoring


dexterbg:
Myles, you should add a `changes.txt` entry to inform users about the new TPMS support, and you might consider adding a user manual page for the Think -- that's missing up to now, as there was no maintainer. Use some other `docs/index.rst` page as a template.

dexterbg:
Regarding the wheel assignment, if you cannot figure out which sensor is which wheel, have a look at the Smart EQ, which has the same issue. @zorgms (the Smart maintainer) added a user configurable sensor mapping, and also added support for this to the Android App.

Bildschirmfoto vom 2025-11-21 15-55-28

It should be easy to adopt this scheme. Maybe even a generalized version of the wheel mapping could make sense, if this is helpful for more vehicles.

zorgms:
> It should be easy to adopt this scheme. Maybe even a generalized version of the wheel mapping could make sense, if this is helpful for more vehicles.

WiP

zorgms:
Like this?

image

MylesTwete:
Thanks guys.  I'll definitely look at doing that.  THINK's automatic TPMS learning procedure (initiated with a long button press) depends on the front tires pressurized to near 36# and the rear ones pressurized to near 43#.  While the data order on CAN is LF, RF, RR, LR, I've never been convinced that the car determined right from left, only front from back.  And then even Front/Back can be spoofed by pumping the tires up in the opposite way.  I think the car just assigns L/R based on which reported first, but I don't know.  And so, it might change with each learning, each car, etc.  So for sure, what you suggest is along the lines of what I was thinking would be needed, so very useful---thanks.

zorgms:
The same is true for the smartEQ; the sides or front/rear are reversed or both.

#1268

dexterbg:
Myles, in case you missed this: ZorG already implemented the mapping for you in PR #1268. You only need to pull the latest `master` to get the mapping feature.

MylesTwete:
> Myles, in case you missed this: ZorG already implemented the mapping for you in PR #1268. You only need to pull the latest `master` to get the mapping feature.

Awesome...will check it out.

MylesTwete:
Thanks, yes, I just took the latest Edge build out for a spin with one of my cars.  I tested out the Tire Change buttons in both the TPMS tab in the OVMS UI and the TPMS change popup in the Android App.  I first swapped L/R in the OVMS UI and saved.  Testing on the road showed the Front tire orientation was now corrected but the Rear orientation was now wrong.  So in the App, I swapped the RR and LR.  The tire pressure orientation was now correct but the RR tire data completely disappeared on the screen.  Tapping on where the tire data should be, the L/R popup came up and I changed the orientation for that tire.  At first it seemed it was fixed, but then I noticed they were both showing the same tire pressure (actual should show 4psi diff).  So it seems there's a bug in this.  Still, overall, it's great to have this tool---and very timely for me.
Screenshot_20251122_133323_Open Vehicle 
Screenshot_20251122_133238_Open Vehicle
Screenshot_20251122_133338_Open Vehicle
image
image

dexterbg:
Myles, please check/post the `tpms.*` part of your `config list vehicle` output. From your description I'd say both your rear sensors now map to the same wheel.

Also, it seems ZorG got the rear left/right assignment wrong, assuming the standard order left-right, while your sensor interpretation is right-left. I'd say you need to exchange the `m_tpms_index` indices 2 & 3 in both ID cases to fix that.

MylesTwete:
> Myles, please check/post the `tpms.*` part of your `config list vehicle` output. From your description I'd say both your rear sensors now map to the same wheel.

Sure, sometimes, but it's not that simple.  When I first ran this update, all 4 tires' pressures reflected distinct tire pressures.  This was after swapping L/R using this tool in the OVMS UI.  And that swap fixed the orientation issue I had with the front tires.  But it was inappropriate for the rear tires.  And so, after discovering and testing the tire change popup in the Android App, I swapped the Rear L/R tires, which worked partially since now the Left tire showed the low pressure value.  Unfortunately, the Right tire showed the same value.  After trying to swap these back, the Right tire pressure disappeared.  So, there's at least a bug here in the App.  Maybe it's also a bug in the OVMS UI implementation.

> Also, it seems ZorG got the rear left/right assignment wrong, assuming the standard order left-right, while your sensor interpretation is right-left. I'd say you need to exchange the `m_tpms_index` indices 2 & 3 in both ID cases to fix that.

ZorG introduced a L/R error in the THINK code by changing it.  Yes, as you suggest, swapping the indices 2&3 would fix that.  And that will be bringing it back to the mapping I intentionally coded.  I don't know why there was a felt need to code it with m_tpms_index instead of the system constants for FL, FR, RL, RR, which is what I used as I thought it made it clear that the rear tire order was different for THINK.  I wasn't aware he'd be making this change.  Does he have a THINK car?

Finally, it doesn't matter to me much that the code was changed.  I'd say it's highly likely that whoever created the THINK DBC files actually got this wrong and that the order should be the same as generically coded in OVMS TPMS code: FL, FR, RL, RR.  But even then, the Automatic TPMS Learning on the THINK does not have any way of knowing Left from Right unless wheel speed differences while turning are used to determine it.  So again, it really doesn't matter.  What does matter is that this swapping tool not be buggy.

Looks like ZorG has swapped these two indexes back...I really think that changes in data order and interpretation should be left to folks who have the particular vehicle, its DBC file and documentation.

image

zorgms:
> I don't know why there was a felt need to code it with m_tpms_index instead of the system constants for FL, FR, RL, RR

System constants (static index) do not allow arbitrary reordering of values. `m_tpms_index`, on the other hand, allows the assignment of a new index value as needed.

dexterbg:
→ https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1271#issuecomment-3568221613

PR #1265, opened 2025-11-21
Add TPMS sensor mapping to vehicle config

zorgms:
Moved vehicle configuration handler to a new file and added a TPMS tab for sensor mapping in the web interface. Updated OvmsVehicle to load TPMS wheel index mapping from config, and added corresponding member variable. 

Also fixed climate schedule help text example formatting.

image


dexterbg:
ZorG, be aware I changed some of your mapping code to make it adapt to vehicle type specific wheel layouts, so pull before making further changes.

PR #1266, opened 2025-11-21
smartEQ: Refactor TPMS sensor index configuration storage

zorgms:
Moved TPMS sensor index values from 'xsq' to 'vehicle' config parameter and renamed internal variable to m_tpms_index_sq. Updated related logic in commands, web configuration, and config change handling. Bumped version to 2.1.0 and preset version to 4.


dexterbg:
Do you actually still need `m_tpms_index_sq`, couldn't you just use `m_tpms_index` instead now? I don't see a reason to duplicate that locally… maybe I miss sth?

zorgms:
This is needed for the EQ TPMS website, which I would have to adapt, but I'm too lazy to do that right now.

zorgms:
I'll take another look at why I did it that way.

zorgms:
Yesterday my brain was foggy.

PR #1268, opened 2025-11-21
thinkcity: Refactor TPMS index usage in CAN frame handler

zorgms:
Replaced hardcoded TPMS index constants with m_tpms_index array for pressure and warning extraction in IncomingFrameCan1. This improves maintainability and consistency in tire pressure monitoring system data handling.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
@@ -224,19 +224,19 @@ void OvmsVehicleThinkCity::IncomingFrameCan1(CAN_frame_t* p_frame)
         // TODO: Add Warnings and Alerts --- e.g. for sensor not responding (data is 0xFF)
 			  if (d[0]!=0xFF) // Front-left
 		    {
-		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(MS_V_TPMS_IDX_FL, (float)d[0] * 0.3625, PSI);
+		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(m_tpms_index[0], (float)d[0] * 0.3625, PSI);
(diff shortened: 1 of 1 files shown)


MylesTwete:
I just took the latest Edge build out for a spin with one of my THINKs. I tested out the Tire Change buttons in both the TPMS tab in the OVMS UI and the TPMS change popup in the Android App. I first swapped L/R in the OVMS UI and saved. Testing on the road showed the Front tire orientation was now corrected but the Rear orientation was now wrong. So in the App, I swapped the RR and LR. The tire pressure orientation was now correct but the RR tire data completely disappeared on the screen. Tapping on where the tire data should be, the L/R popup came up and I changed the orientation for that tire. At first it seemed it was fixed, but then I noticed they were both showing the same tire pressure (actual should show 4psi diff). So it seems there's a bug in this. Still, overall, it's great to have this tool---and very timely for me.

Screenshot_20251122_133323_Open Vehicle

Screenshot_20251122_133238_Open Vehicle
Screenshot_20251122_133338_Open Vehicle

MylesTwete:
I see that both of the Think City TPMS Cases (0x344 and 0x345) were changed in #1268.  That ended up changing the Rear tire L/R orientations.  The tire order defined by m_tpms_index[] (FL, FR, RL, RR) is not the same as the order in the DBC file for CAN messages for THINK (FL, FR, RR, RL).  The Rear tire order is different based on what DBC files used for THINK CAN messages have long shown.  Having said that, it's been suspected for some time that that the automatic TPMS learning procedure used on THINK does not get the Left/Right orientation correct and maybe just puts them in some arbitrary order---maybe 1st one goes Left...so, it's fine with the way it's written now...we just need to figure out what causes the RR tire pressure to disappear in the display...
image

dexterbg:
→ https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1263#issuecomment-3567580460

PR #1271, opened 2025-11-23
Fix TPMS index mapping for rear tires

zorgms:
Corrects the mapping of TPMS pressure and warning values for rear-left and rear-right tires to ensure accurate metric assignment. This resolves swapped index usage in CAN frame processing.

#1263 
#1268

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_thinkcity/src/vehicle_thinkcity.cpp
@@ -232,11 +232,11 @@ void OvmsVehicleThinkCity::IncomingFrameCan1(CAN_frame_t* p_frame)
 		    }
 		    if (d[2]!=0xFF) // Rear-right
 		    {
-		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(m_tpms_index[2], (float)d[2] * 0.3625, PSI);
+		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(m_tpms_index[3], (float)d[2] * 0.3625, PSI);
 		    }
 		    if (d[3]!=0xFF) // Rear-left
 		    {
-		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(m_tpms_index[3], (float)d[3] * 0.3625, PSI);
+		        StandardMetrics.ms_v_tpms_pressure->SetElemValue(m_tpms_index[2], (float)d[3] * 0.3625, PSI);
 		    }
 		  break;
     case 0x345: // TPMS Status Alerts and Warnings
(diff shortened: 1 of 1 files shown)


MylesTwete:
FWIW, this bug was introduced with the prior edit of the THINK City CPP file where intentional system TPMS constants were replaced with the tpms_index mapping instead and without consideration that the FL, FR, RR, RL data order was intentional.  I thought the line comments in the code would make the intentionality clear.  Regardless, I seriously doubt THINK's automatic TPMS learning procedure gets L/R orientation correct and it may change with each repeat of the procedure---i.e. could be random.  So, having the new TPMS tire selector/associator tool is a great step forward---now if the bugs can be fixed in that I would truly be happy with TPMS.

dexterbg:
Myles, a bug in the mapping is quite unlikely, as that only provides a dynamic addressing of the wheels.

The only issue introduced by ZorG in the previous PR and solved by this PR was switching the rear values L/R. The only change he did on your code was adding the mapping as suggested, i.e. replacing the fixed wheel constants by the dynamic index. ZorG did not change your underlying CAN decoding in any way. You can verify that by checking the PR's file diff view: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1268/files

But I understand you don't want us to help you other than by describing what you need to do, and that's perfectly OK. I second it's better to teach than to take over.

Now, trying to explain your issue's potential source a bit better: the mapping system in the current state does not block or salvage config changes that lead to multiple sensors mapping to the same wheel. If a change command from your App was lost (can happen), your config could now have that situation.

Another possible source is the App's sensor mapping UI, that currently does not show the actual assignment, but always returns to the default when opening the selector. If you then accidentally apply that state, you reset the assignment.

You can see if your config is "broken" by checking the module's web UI mapping page (that's reading/writing the config directly) or by checking the `tpms` variables in `config list vehicle`, those are the actual mapping indices currently used by the module.

When using the default layout, the wheel values are: 0=Front Left, 1=Front Right, 2=Rear Left, 3=Rear Right

No two sensors should have the same wheel index.

With just the front sides exchanged, your mapping should look like this:
```
OVMS# config list vehicle
vehicle (readable writeable)
[…]
  tpms.fl: 1
  tpms.fr: 0
  tpms.rl: 2
  tpms.rr: 3
[…]
```
…and the web UI should accordingly look like this:
Bildschirmfoto vom 2025-11-23 19-05-18

Regards,
Michael

MylesTwete:
Thanks Michael.
It's not that I don't appreciate help on the code---after all, I'm more likely to break things than fix them given my experience.
I just was very surprised to see the THINKCity main module was changed without my knowledge given I'm likely the only person in the world right now using OVMS on a THINK and the only one set up to actually test changes like this.

I haven't dowloaded ZorG's changes made to correct the RR/RL indexing switcheroo introduced as I was good with it being the way it was anyway.  I did just test a hunch though.  Where things stood as of my last post on this, only 3 sensor pressures were visibly displayed, with 1 of the rear ones missing (or perhaps with background color for text?).  So in OVMS UI, TPMS config tab, I set the Rear tire orientation saved it.  Thinking this might fix the problem, I took a test drive just now.  Still showing just 3 sensors even though now the rear one displayed was moved to the Right side.  This persisted.  So I checked the assignments in each case via the popups and they were correct.  On a hunch, I changed the one missing to the opposite tire, then back.  Still no change.  Then with the other tire, I changed it, then changed it back.  After that, both rear tire pressures were now visible again and reflected properly the tire pressures (RL being very low pressure, RR being normal).  
So, there is some bugginess to this switching mechanism, but it looks like it can be cleared up with enough "select, save, test" cycles.  And once it's set, all should be good.

Final snapshot:

Screenshot_20251123_144709_Open Vehicle

Thanks :-)

zorgms:
Note: Use the latest build version of the app from here:
https://github.com/openvehicles/Open-Vehicle-Android/tree/master/latest

And activate the following in the app's settings so that the new sorting is sent to the module.

image

As I can see, the app still needs to be adjusted here; currently, this option is only available for the smart EQ.

MylesTwete:
No luck with THINK using the latest App and module code.  As you suggest, that "sort TPMS value by Firmware" is not available to THINK, so I wasn't able to make any tire switcheroo work with the App.  I also tried setting both configurations for L/R Rear tires in the OVMS UI but the only way the correct tire aligns with the RL is when both RL and RR end up displaying the same value.  I suppose I'll need to look at SmartEQ code at some point and see what's missing for THINK.  Maybe in a few days...

PR #1273, opened 2025-11-24
Add TPMS sensor mapping display command

zorgms:
Introduces the 'mapping' command to show current TPMS sensor mapping, including wheel positions, sensor assignments, and validation checks for duplicates and out-of-range indices. Enhances user visibility and troubleshooting of TPMS configuration.

image

image

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_tpms/src/ovms_tpms.cpp
+++ b/vehicle/OVMS.V3/components/ovms_tpms/src/ovms_tpms.cpp
@@ -593,7 +593,7 @@ void tpms_map_set(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc,
       return;
       }
     // Map: dest_wheel gets the sensor that was previously at src_wheel
-    new_map[dest_idx] = current_map[src_idx];
+    new_map[dest_idx] = src_idx;
     }
   
   


dexterbg:
ZorG,

thinking about this now, adding a user layer around the easily corruptable set of config variables is necessary. By this I mean just showing the mapping doesn't help to prevent mistakes, and telling users to use the config command may be a recipe for desaster in this case, as a user setting a mapping variable to an illegal value can produce a constantly crashing module without the user knowing how to fix this.

Also, changing the mappings one wheel at a time has the inherent issue of producing an inconsistent/invalid state, and the added issue of losing valid TPMS sensor data, as we cannot rearrange the existing metrics values.

So the mapping status command is a step in the right direction, but we should extend this by a) a command to reconfigure the full mapping set in a single step, and b) a defined status output format for the Apps to query the full mapping by a single call.

Configuring the mapping could be done by something like `tpms map set fl=fr fr=fl …` with the command validating the new mapping before actually changing the config. When changing the mapping, the command should rearrange the existing TPMS vector metrics accordingly, so the readings get corrected immediately.

Also, while your implementation tries to accomodate for vehicle specific wheel layouts on codes & names, it fails to do so when reading the config. See my config reader in the vehicle class on how to do this in a generic way.

The status command should also check if the current vehicle actually uses the mapping, and this condition should also have a place in the App output format version, so the App can accomodate to the currently selected vehicle without the user needing to set a global config for this (keep in mind there are users with multiple different vehicles).

Regards,
Michael

dexterbg:
Forgot a side note: reinventing the `tpms status` output in the mapping status command doesn't feel right. May be better to factor out the sensor part from the status command and call that from both commands.

zorgms:
Okay, the TPMS reordering is rather vehicle-specific. With the smartEQ, I record the data in variables and reorder them every 60 seconds, then set the corresponding metrics. Others set the metrics directly with the fixed index. So, all fixed index values ​​would first have to be converted to the new indexes. Then, reordering after changing the index would be relatively simple.

`tpms status` is untouched? You mean the status shouldn't be displayed with TPMS mapping? You mean the status shouldn't be displayed with `tpms mapping`? Currently, the mapping is displayed with status values ​​so that you have a reference to the values.

dexterbg:
A minimal implementation will normally look like the one on the Think, simply setting TPMS values directly when reading the CAN message. Reordering the four vectors solves the inconsistency for any kind of handling well, works immediately (not relying on a ticker to occur) and really isn't that complex.

I mean the `tpms status` sensor output is sufficient for `tpms mapping` as well, so can simply be included there, unless I miss some added info in your version that is relevant for the mapping.

CrashOverride2:
Hello guys, I've already received a few inquiries regarding TPMS sensor assignment on the Zoe Ph2 also. 

I've started experimenting with reading and writing the sensor IDs into the BCM so that they are displayed correctly on the dashboard.

How does it look on the Smart? Is there something similar, or why are the sensors assigned via software in OVMS? Does the Smart not have an individual display telling you which tire needs air, so that it wouldn’t matter here?

zorgms:
The smartEQ does not display any sensor information, only a warning if the pressure is too low.

vehicle/OVMS.V3/components/ovms_tpms/src/ovms_tpms.cpp
```diff
@@ -121,10 +121,18 @@ void tpms_status(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc,
 
   if (StandardMetrics.ms_v_tpms_alert->IsDefined())
     {
-    const char* alertname[] = { "OK", "WARN", "ALERT" };
+    const char* alertname[] = { "OK", "WARN", "ALERT", "UNKNOWN" };  // Added UNKNOWN
```
dexterbg: Adding "UNKNOWN" here is superfluous as you handle that case explicitly below.

dexterbg:
Looks good. I suggest renaming the "show" command to "get", as that a) better relates to "set"/"reset" and b) "show" implies more of a user level command, while "get" is recognizable as not being meant for users.

dexterbg:
On the `get` format: I suggest changing that to an extended/modified version of the V2/MP `Y` TPMS data message format:

https://docs.openvehicles.com/en/latest/protocol_v2/messages.html#car-tpms-message-0x59-y

The current format definition lacks the wheel count and the ability to accomodate to empty/short vectors. The meta info (undefined/stale/valid) may be unnecessary here, but why not keep it, just in case it gets in handy.

Btw, extending the `Y` message to include the mapping as an additional vector could also be an option, possibly eliminating the need to query the current state from the V2/MP App, as `Y` is sent regularly by default.

zorgms:
Let's stick with this example.
First number = how many values ​​per category; repetition is not necessary.
Then follow the wheel layout names, optionally with an index?
Pressure, temperature, status, alarm, each with a leading character?

Like this?
`4,FL,FR,RL,RR,230.5,245.0,238.2,241.8,21.0,22.5,20.8,23.1,100.0,100.0,100.0,100.0,0,0,0,0`
Or so?
`4,FL:0,FR:1,RL:2,RR:3,P0:230.5,P1:245.0,P2:238.2,P3:241.8,T0:21.0,T1:22. 5,T2:20.8,T3:23.1,H0:100.0,H1:100.0,H2:100.0,H3:100.0,A0:0,A1:0,A2:0,A3:0`

dexterbg:
You now leave the `Y` format definition again. The format is defined that way to be extensible and automatically accomodating for different vector types and coverage as needed. Most vehicles cannot fill all four vectors, many sensors can only provide a pressure or a health level.

By using the generic vector format structure, clients can easily parse any number of vectors of arbitrary length simply during a forward scan, and easily skip data they don't need. Additionally tagging the vectors is not necessary, as the vectors included are predefined and guaranteed to be extended only at the end.

The `Y` record is designed that way to provide the means to add future additional sensor data possibly needed. Think about the concept of a "wheel" here in a more generic way of a "force transmitting device", transcending road vehicles. Example: on a man carrying octacopter, each motor/propeller would be a "wheel".

So to stay within the `Y` definition, the example would look like this:
```
4,FL,FR,RL,RR,4,230.5,245.0,238.2,241.8,1,4,21.0,22.5,20.8,23.1,1,4,100.0,100.0,100.0,100.0,1,4,0,0,0,0,1,4,0,1,2,3,1
```

On the Think, only supporting pressures and alerts, this would reduce to:
```
4,FL,FR,RL,RR,4,230.5,245.0,238.2,241.8,0,-1,0,-1,4,0,0,0,0,1,4,0,1,2,3,1
```

On the VW e-Up, only supporting health and alerts and not using the mapping, a similar example would be:
```
4,FL,FR,RL,RR,0,-1,0,-1,4,97.6,99.6,89.0,97.6,0,4,0,0,0,0,0,0,-1
```

By extending the `Y` structure like this, the formatter for the server V2 component can be reused for the `tpms get` command, and the App's parser can be used for both applications as well -- even better yet, if we add the mapping vector to the `Y` message, the App doesn't need to call `tpms get` at all, as the `CarData` `Y` parser will already read & supply the mapping.

zorgms:
Do you mean something like this?
`I (84186) ovms-server-v2: Send MP-0 Y4,FL,FR,RL,RR,4,210.0,220.0,230.0,240.0,1,4,21.0,22.0,23.0,24.0,1,0,-1,4,0,0,0,0,1,4,0,1,2,3,1`

zorgms:
```
I (520656) ovms-server-v2: Incoming Msg: MP-0 C7,config set vehicle tpms.fr 3
D (520926) events: Signal(config.changed)
D (520926) webserver: ConfigChanged: config.changed vehicle
I (520936) vehicle: TPMS mapping changed, rearranging metrics
D (520936) vehicle: TPMS remap: old_pressure.size=4, count=4
D (520936) vehicle:   Wheel[0]: old_sensor_idx=0 -> new_sensor_idx=0
D (520936) vehicle:   Wheel[1]: old_sensor_idx=1 -> new_sensor_idx=3
D (520936) vehicle:   Wheel[2]: old_sensor_idx=2 -> new_sensor_idx=2
D (520936) vehicle:   Wheel[3]: old_sensor_idx=3 -> new_sensor_idx=3
```

`I (572186) ovms-server-v2: Send MP-0 Y4,FL,FR,RL,RR,4,210.0,240.0,230.0,240.0,1,4,21.0,24.0,23.0,24.0,1,0,-1,4,0,0,0,0,1,4,0,3,2,3,1`

dexterbg:
Yes, exactly this.

dexterbg:
The `Y` payload should become the output of the `tpms get` command as well, and the protocol documentation needs to get the update… I think this is ready to be merged then, but I'll have another look at the final state.

zorgms:
[car_TPMS_0x59_Y.md](https://github.com/user-attachments/files/23900207/car_TPMS_0x59_Y.md)

[car_TPMS_0x59_Y.pdf](https://github.com/user-attachments/files/23900208/car_TPMS_0x59_Y.pdf)

PR #1274, opened 2025-11-24
Add TPMS fix for iOS OVMS App

zorgms:
Introduces a new configuration option to enable a TPMS temperature fix for iOS Open Vehicle app integration. When enabled, tire temperature values are kept above 1 to ensure TPMS data is displayed correctly in the iOS app.
Added this option to TPMS mapping Website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ provides data logging to SD card and to a server.
   - [User Support Forum](https://www.openvehicles.com/forum)
   - [Android App](https://play.google.com/store/apps/details?id=com.openvehicles.OVMS&hl=en_US) [APK file](https://github.com/openvehicles/Open-Vehicle-Android/tree/master/release)
   - [iOS App](https://apps.apple.com/us/app/open-vehicles/id490098531)
+  - [Home Assistant integration](https://github.com/enoch85/ovms-home-assistant)
 - _Distributors_
   - _Original Design_
     - [Medlock & Sons (North America)](https://medlockandsons.com/product/ovms-v3/)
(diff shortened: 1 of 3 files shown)


dexterbg:
Adding sensor spoofing to work around a bug (?) in an open source client is really ugly.

@markwj Mark, any chance this could be fixed on the iOS App side?

zorgms:
https://github.com/openvehicles/Open-Vehicle-iOS/pull/55

I submitted a pull request and closed it after three months because there was no response. I don't think it will be easy to fix the iOS app anymore, as it no longer complies with Apple's current guidelines. @Dimitrie78 implemented this fix for the smart EQ, and that made iOS users happy.

dexterbg:
OK… assuming the iOS App is beyond fixable at the moment: the proper way to do this kind of workaround is then to modify not the actual sensor metrics on the module, but to alter the values sent to the V2 server on the fly.

That way the correct & unaltered sensor values are still available for applications or plugins accessing the metrics values directly, as well as for clients using a V3/MQTT connection.

So this is rather a Server V2 config, and thus should be moved there regarding the config namespace as well as the UI.

It should also be made absolutely clear from the UI wording & config naming, that this is a temporary workaround, that will be removed once a new iOS App release is possible.

I.e. the config variable should be moved to `server.v2` and named `workaround.ios_tpms_display` or similar.

The config UI info on the option should also be rephrased:
```
<p>Workaround for legacy iOS App, hiding zero/negative temperatures to keep other TPMS data visible.</p>
<p>With the fix enabled, use command `tpms` to show actual temperatures at/below zero.</p>
```

zorgms:
#1280

PR #1276, opened 2025-11-29
smartEQ: Refactor TPMS indexing for consistent sensor mapping

zorgms:
Updated TPMS value assignments to use m_tpms_index for all sensor data storage and retrieval, ensuring consistent mapping between physical wheel positions and internal arrays. Simplified setTPMSValue logic to remove redundant index mapping and bounds checks, improving code clarity and reliability.

In #1228, incorrect tire indices were merged.
0x673 and PollReply_TPMS_InputCapt are not in the same order.


zorgms:
Please only merge if #1278 is also being merged.

PR #1281, opened 2025-12-06
Refactor webserver config handling into separate files

zorgms:
Moved configuration handlers for autostart, backup, cellular, firmware, locations, logging, notifications, pushover, status, webserver, and wifi into dedicated source files. Cleaned up includes in ovms_webserver.h and web_cfg.cpp, improving modularity and maintainability of the webserver component.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.h
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.h
@@ -33,22 +33,16 @@
 #define __WEBSERVER_H__
 
 #include <forward_list>
-#include <iterator>
 #include <vector>
-#include <memory>
 #include <utility>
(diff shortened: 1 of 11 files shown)


vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.h
```diff
@@ -38,10 +38,17 @@
 #include <memory>
 #include <utility>
 #include <map>
+#include <string.h>
+#include <stdio.h>
+#include <string>
+#include <sstream>
+#include <fstream>
+#include <dirent.h>
```
dexterbg: Includes not needed for the class API definition shall be placed in the source module instead, so modules only needing the class API don't need to pull in unnecessary additional headers.

PR #1284, opened 2025-12-09
Add conditional TPMS panel to status page

zorgms:
TPMS panel is now displayed on the status page only if valid TPMS data is available, determined by the output of 'tpms map get'. This prevents showing the panel when only empty or default data is present.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg_status.cpp
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg_status.cpp
@@ -218,6 +218,22 @@ void OvmsWebServer::HandleStatus(PageEntry_t& p, PageContext_t& c)
       "<li><samp id=\"modem-cmdres\" class=\"samp-inline\"></samp></li>"
     "</ul>");
 
+  output = ExecuteCommand("tpms map get");
+  // Show TPMS panel only if we have actual data (not just "0,-1" empty sections)
+  // Format: wheel_count,names...,pressure_count,data,validity,...
+  // Empty example: 4,FL,FR,RL,RR,0,-1,0,-1,0,-1,0,-1,0,-1 (all data sections are 0,-1)
+  if (!output.empty() &&
+      output.find(",0,-1,0,-1,0,-1,0,-1,") == std::string::npos)
+    {
+    c.print(
+      "</div>"
+      "<div class=\"col-sm-6 col-lg-4\">");
+    c.panel_start("primary", "TPMS");
+    output = ExecuteCommand("tpms status");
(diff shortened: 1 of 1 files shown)


dexterbg:
Adding the TPMS sensor data to the "Status" page like that doesn't fit the purpose of the page. The "Status" page is meant to provide an operational status overview of the module & vehicle, i.e. mostly component status & basic vehicle mode of operation. As part of that, the most important metrics coupled to these status modes can be included in the "Live" panel.

The TPMS status isn't an operational status but rather a metrics view, and the metrics are not coupled to an operational status. A proper placement of a TPMS sensor display thus would be on the "Dashboard" page, and it should be providing a live view there. That needs a different approach, i.e. direct metrics stream updates instead of calling a shell command (see https://docs.openvehicles.com/en/latest/components/ovms_webserver/docs/metrics.html).

Side note: deciding on displaying the TPMS status should not be done on base of analysing a shell command output. As the webserver runs on the module, it can check for TPMS metrics being defined directly by testing `StandardMetrics.ms_v_tpms_alert->IsDefined()` et al, just like the `tpms status` implementation does.

If becoming part of the dashboard, another option would be to extend the `DashboardConfig` API by the vehicle specific TPMS configuration. That opens up more options like reacting to some user configuration in the getter.

If you still want to provide a TPMS display of any kind for the "Status" page for now, I suggest adding that as a user configurable plugin. Extending the "Status" page panels can be done easily by registering for the page's `body.post` plugin hook.

See https://docs.openvehicles.com/en/latest/components/ovms_webserver/docs/hooks.html

This would be the full plugin for a simple `tpms status` monitor:
```
<!--
  Hook plugin for /status:body.post
  - add custom panel "TPMS" after "Vehicle"
-->

<!-- TPMS Status Panel -->
<div class="col-sm-6 col-lg-4" id="panel-tpms-cell" style="display:none">
  <div class="panel panel-primary" id="panel-tpms">
    <div class="panel-heading">TPMS</div>
    <div class="panel-body">
      <samp class="monitor"
        data-events="\.tpms\."
        data-updcmd="tpms status"
        data-updcnt="1">
      </samp>
    </div>
  </div>
</div>

<script>
  $('#main').one('load', function(ev) {
    $('#panel-tpms-cell').insertAfter($('#panel-vehicle').parent()).show();
  });
</script>
```

Alternatively that plugin could also provide live metrics, but that would need to accomodate for the wheel layout and partially available sensor data, so become more complex.

A hook plugin could also be supplied for the "Dashboard" page (see Twizy plugins for examples), but a TPMS live view for that page would make sense for most users, so that would make sense being added to the framework.

zorgms:
It was just a stopgap solution for me; I don't need it and it offers no added value since I don't have a Wi-Fi connection to the vehicle. This simple implementation is the maximum amount of time I'm willing to invest, so it can go straight in the bin.

PR #1289, opened 2025-12-12
Reduce stack usage in Duktape VFS and idtag utility

zorgms:
Raised the stack size for both DuktapeVFSLoad and DuktapeVFSSave tasks from 5*512 to 6*1024 to improve stability and prevent stack overflows during script execution.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_script/srcduk/ovms_duk_vfs.cpp
+++ b/vehicle/OVMS.V3/components/ovms_script/srcduk/ovms_duk_vfs.cpp
@@ -143,7 +143,7 @@ DuktapeVFSLoad::DuktapeVFSLoad(duk_context *ctx, int obj_idx)
   Ref();
   Register(ctx);
   TaskHandle_t task = NULL;
-  if (xTaskCreatePinnedToCore(LoadTask, "DuktapeVFSLoad", 6*1024, this,
+  if (xTaskCreatePinnedToCore(LoadTask, "DuktapeVFSLoad", 5*512, this,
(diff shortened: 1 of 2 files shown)


zorgms:
I'm not sure if the value needs to be increased so extremely; that came out as a result of backtrack analysis.

dexterbg:
Can you please supply all details of the crash? The loader & saver shouldn't need that much stack space, and internal memory is tight on some vehicles.

zorgms:
Stack Overflow Fix Summary
DuktapeVFSLoad / DuktapeVFSSave Tasks
Problem: Crash log showed Task stack was corrupted or overflown, Stack=0 for DuktapeVFSSave task.

Root Cause: The original stack size of 5*512 = 2560 bytes was insufficient for:

VFS file operations (fopen, fread, fwrite, fclose)
String handling and path resolution
SPIFFS/SD card driver calls which have deep call stacks
Fix: Increased from 5*512 (2560) → 6*1024 (6144 bytes)

Minimum recommended: 4*1024 = 4096 bytes should be sufficient, but 6*1024 provides a safety margin for edge cases (long file paths, nested directory structures).

dexterbg:
Please post the crash log you refer to. Where do those recommendations come from, what are they based on?

dexterbg:
Or, better yet, post some test code & data reproducing the crash.

dexterbg:
Checking the current crash report table on my server (https://dexters-web.de/f/firmware/developer/crashreport.csv) actually shows five crashes from stack overflow within `DuktapeVFSSave`, but none in `DuktapeVFSLoad`.

So this should be reproducable given the right data & environment, and maybe then also solvable in another way.

PR #1296, opened 2025-12-19
BmsCheckChangeCellArrangementVoltage update

KommyKT:
To update the layout when modify BmsSetCellArrangementVoltage(x, y);
 Last time I modified the BmsSetCellArrangementVoltage(96, 12); -> BmsSetCellArrangementVoltage(96, 8); but the bms stat command output is not updated. Now either this code update is fixed.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle_bms.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle_bms.cpp
@@ -87,54 +87,65 @@ struct reading_entry_t
  */
 bool OvmsVehicle::BmsCheckChangeCellArrangementVoltage(int readings, int readingspermodule /*=0*/)
   {
-  bool res = false;
-  if (readingspermodule <= 0)
-    {
-    // Passing in 0 will leave the readings per module unchanged.
-    readingspermodule = m_bms_readingspermodule_v;
-    }
-  if (readings != m_bms_readings_v)
-    {
-    res = true;
-    if (m_bms_bitset_cv == 0)
+    bool res = false;
+
+    // Defining valid readingspermodule
+    if (readingspermodule <= 0)
       {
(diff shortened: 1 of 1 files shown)


dexterbg:
Tamás, this is a non-solution to a non-issue, and you cannot have verified this. Did you use "AI" to create this?

If not, please explain/show the issue, i.e. post your annotated `bms stat` output, and what you intended this code change for.

Regards,
Michael

KommyKT:
I use ai.

My problem is the lates and older versions is BMS stat output.
Screenshot 2025-12-20 at 6 59 01
Screenshot 2025-12-20 at 7 00 33
See the bms stat output with the edge and in the code the bms init process  BmsSetCellArrangementVoltage(96, 8);
show 12 reading /model instead of 8 readings/module.
but with the ai code is show the desired output

The Ioniq 28kWh battery moduls not have equivalent battery cells. 
In module:
 1,2,3,10,11,12 has 6 battery cell
 4,5,6,7,8,9 has 10 battery cell 
 
 and each module has 1 temp sensor
 
CFF3A32E-5FA7-4390-A203-6B4454BAB097

dexterbg:
> but with the ai code is show the desired output

*sigh*

No it's not.

I can tell that without even testing the code. How can I be so sure about that? Because the function your "AI" thinks changing would have an effect _isn't used at all by the Ioniq vFL module_.

And even if it was used, it can never cause the BMS status output to match your Ioniq battery cell layout, because that function does not do the cell layout formatting, it only reorders sensor values.

**IOW: your "AI" is too stupid to even understand how code works on the most basic level, let alone produce any suggestion for an improvement that isn't total nonsense.**

Tamás, you are wasting my time by posting unvalidated "AI" bullshit.

There is of course a real solution to your issue, but right now I think you should start using your own brain again, so I'll help you when you figured it out. I've given you some hints in this post.

PR #1302, opened 2025-12-31
prevent multiple profile0 timer starts

sharkcow:


dexterbg:
> It seems like the software timer for profile0 gets started multiple times, which shouldn't happen. I have modified the code to prevent this.

Hm. What do you expect this change to do?

If you think restarting an already running timer would create another concurrent run, that's false. A timer has only a single instance & control state, restarting while running simply resets the timer. Quoting from the docs:
```
  xTimerStart() starts a timer that was previously created using the
  xTimerCreate() API function.  If the timer had already been started and was
  already in the active state, then xTimerStart() has equivalent functionality
  to the xTimerReset() API function.
```

If you mean it gets restarted without need, e.g. after the processing already has taken place, that would need another approach.

sharkcow:
OK thanks, that was wrong.

PR #1310, opened 2026-01-04
Config framework: race condition fix & storage transaction scheme

dexterbg:
Happy new year everyone :)

The OvmsConfig framework has been bad at optimizing multiple successive changes from the beginning, and has been subject to a race condition since async event processing. This change aims at solving both.

A typical write pattern of the OvmsConfig API looks like this:

```
  MyConfig.SetParamValue("myparam", "instance1", "value1");
  MyConfig.SetParamValue("myparam", "instance2", "value2");
  MyConfig.DeleteInstance("myparam", "instance3");
  MyConfig.SetParamValue("myparam", "instance4", "value4");
  …
```

A typical non-copying read pattern looks like this:

```
  OvmsConfigParam* p = MyConfig.CachedParam("myparam");
  …
  for (ConfigParamMap::iterator it=p->m_instances.begin(); it!=p->m_instances.end(); ++it)
    …
```

Up to now, each write here potentially does a) a full rewrite of the config store file associated with "myparam", and b) signals "config.changed" for "myparam" to all listeners. (Potentially because `SetParamValue()` checks for an actual value change.)

So that's a) a lot of unnecessary & slow file I/O, b) unnecessary wear of the flash memory, c) unnecessary repeated calls for all config change event listeners, and worst of all d) has a race condition.

The race condition is the reader already traversing the instances map of "myparam" while the writing process still changes the map. The chance of that happening rises with the number of successive config writes. Even copying the instances map didn't help, because that wasn't protected either.

Crashes from this race condition may show a prohibited memory access from within some config map's RB tree internals, but a map read may also silently return random data.

To solve these issues, I've added a mutex to synchronize access to the param and instances maps, and implemented a transaction scheme on base of the recursive mutex lock to postpone all file writes and signal emissions while a lock has been set, and only perform these once for each config param changed.

The commit normally is done when the lock is released, except when a config backup or restore is running at that time. The commit will then be deferred (in case of a backup) or discarded (in case of a restore).

Each standard `MyConfig` API call to access a single instance needs to acquire the lock now, so some performance can also be gained on multiple successive gets by adding an enclosing lock.

Acquiring the transactional config lock for a function/block typically is simply…

```
  {
  auto lock = MyConfig.Lock();
  … access MyConfig (read/write) …
  }
```

You can also try locking for a given time, and alternatively get a lock pointer instead. The lock is basically an `OvmsRecMutexLock`, so can also be unlocked and relocked within the block.

I've extended all code places doing multiple config changes or relying on accessing the instances map by the config lock. This mostly applies to web UI configuration pages, the pattern there is mostly to simply get a config lock for the whole function, unless the UI handler needs to do a blocking operation in another context that also needs to access the config -- this is currently only the case for changing the vehicle type, as that involves a synchronous vehicle/poller task shutdown.

On the config reader side, as essentially every listener to the "config.changed" and "config.mounted" events needs to lock the config, I've added this in the event framework around the listener callback loop. This reduces the locking overhead and keeps config change listeners simple.

A listener can in theory release the lock to do a blocking operation like the vehicle change by accessing the config store mutex, but I recommend avoiding that. Performing blocking calls to other contexts within an event listener should generally be seen critical, as that can easily lead to other deadlocks or event queue overflows.

Only the netmanager task currently does this; the `PrioritiseAndIndicate()` operation to select the default interface and configure the DNS list is executed synchronously in the LwIP context. I've fixed the config deadlock there by moving the config read into the listener, but still think about reworking the call to be done asynchronously (as a netmanager job).

(Note: the file logger task's config access was non-critical (outside the loop), but I changed that as well to establish the pattern.)

Please inform me if I missed some other config listener doing this.

This touches a lot of places and changes an 8 year old foundational framework component, so I'd love to get some audit and test feedback for this before actually including this in the edge build.

To monitor the config transactions and events, set log level debug for "config" and "events".

Thanks!
Michael

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_config.cpp
+++ b/vehicle/OVMS.V3/main/ovms_config.cpp
@@ -775,6 +775,17 @@ void OvmsConfig::DeregisterParam(std::string name)
     }
   }
 
+
+bool OvmsConfig::IsDefined(std::string param, std::string instance)
+  {
+  if (!m_mounted) return false;
+  auto lock = Lock();
+  OvmsConfigParam *p = CachedParam(param);
+  if (p == NULL) return false;
+  return p->IsDefined(instance);
+  }
+
+
 void OvmsConfig::SetParamValue(std::string param, std::string instance, std::string value)
   {
   if (!m_mounted) return;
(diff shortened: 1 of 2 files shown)


dexterbg:
@zorgms ZorG, did you give this a try on the Smart EQ? As the Smart module makes excessive use of config changes, some also triggered by CAN/OBD reads, I see the Smart EQ as a prime test candidate.

dexterbg:
FYI, I've been running this for over a week now on my actual vehicle module (Mii / VWUP) without any issues.

zorgms:
I haven't noticed anything negative. My modules have been running stably for a while now, apparently because I'm not using Wi-Fi and therefore the network manager doesn't need to adjust anything.

What is new, however, is that the data connection is interrupted when charging starts, but this only lasts for about 20 seconds.

dexterbg:
On the data (server v2?) connection being restarted on charge start: I have no idea how that could be connected to this change. I assume it's due to another change or some strange coincidence. Changing the "network" config can trigger a network interface check, but that would only stop the server connection if there actually is an interface change to be applied. That normally doesn't result from a config change, unless the signal quality thresholds or the DNS are changed.

But the charge events do not normally imply any config change. If you do some config change for the smart on charge start, maybe track down how that is processed? Still that shouldn't be any different by this change.

dexterbg:
Included in edge build version 3.3.005-603-ge52c972e

PR #1312, opened 2026-01-05
VWEUP: T26: fix/improve handling of incomplete/faulty profile0 readings

sharkcow:
check for fixed string "Option" in received profile0, improve retry logic
remove blocking delay in callback function

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_vweup/src/vweup_t26.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_vweup/src/vweup_t26.cpp
@@ -295,7 +295,7 @@ void OvmsVehicleVWeUp::vehicle_vweup_car_on(bool turnOn)
     ResetTripCounters();
     if (ocu_awake && m_sendOcuHeartbeat != NULL) {
       ESP_LOGD(TAG, "T26: stopping OCU heartbeat timer");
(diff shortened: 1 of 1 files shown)


dexterbg:
> remove blocking delay in callback function

What do you mean by this? Was some code change possibly lost?

If you mean the `xTimerStop()` 100 ms timeout, that's not a delay, that's a limit for a delay (and still present two more times in that function).

sharkcow:
Yes, but if a delay occurs it will be blocking, right? And I thought we don't want that in a callback. From my understanding, the other instances are not inside a callback, but I could change those as well. But if you think it's not necessary, I can put the 100ms timeouts in again.

dexterbg:
`ReadProfile0()` is always called by `IncomingFrameCan3()` (i.e. running in the vehicle (poller) task context), so if that potential 100 ms blocking time is critical for the CAN3 processing (is it?), the others should be as well.

The point is, handling the timer stop differently raises questions for anyone trying to understand the code, and you didn't differentiate that deliberately to fix something. So I'd say either be sure a failed stop is handled correctly and set all to 0, or keep all at 100 ms.

sharkcow:
OK I get it. Have no changed all to 0, seems to be more consistent also with the rest of OVMS.

PR #1313, opened 2026-01-08
smartEQ: Refactor config handling and update BMS thresholds

zorgms:
Reworked configuration management to use GetParamMap for transactional updates and improved migration/cleanup of deprecated keys. 
Updated BMS cell voltage and temperature limits and thresholds to new recommended values. 
Changed PRESET_VERSION to a date-based format for better tracking.

#1310

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_commands.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_commands.cpp
@@ -936,55 +936,97 @@ void OvmsVehicleSmartEQ::xsq_preset(int verbosity, OvmsWriter* writer, OvmsComma
 }
 
 OvmsVehicle::vehicle_command_t OvmsVehicleSmartEQ::CommandPreset(int verbosity, OvmsWriter* writer) {
-  //MyConfig.SetParamValueBool("xsq", "cfg.preset.act",  true); // activate preset config
-  MyConfig.SetParamValueInt("xsq", "cfg.preset.ver",  PRESET_VERSION); // preset config version
+  // Use GetParamMap() to get COPYs and update in transactions
+  auto map_xsq = MyConfig.GetParamMap("xsq");
+  auto map_vehicle = MyConfig.GetParamMap("vehicle");
+  auto map_modem = MyConfig.GetParamMap("modem");
+  auto map_net = MyConfig.GetParamMap("net");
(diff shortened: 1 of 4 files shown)


dexterbg:
This is over-optimization, probably because I didn't explain the issue & already implemented solution well enough in PR #1310 -- sorry.

It's not advised to use the map API for every config operation.

With adding the config lock/transaction with PR #1310, using the map API is especially no longer necessary to improve the performance: getting the transactional lock makes all single instance calls afterwards automatically use the transaction. There is little overhead on the `MyConfig` API calls remaining with that.

Use the map API only where using it makes the *code* more simple.

You now actually made the code more complex, because you now need to reimplement type conversions, defaults and non-changing reads locally. If you do mostly assignments or need to loop over instances, using the map is much more readable than calling the single instance API methods. But for example replacing…

```
      MyConfig.SetParamValue("system.adc", "factor12v", adc_factor);
      MyConfig.SetParamValueBool("xsq", "calc.adcfactor", calcADCfactor);
```
…by…
```
      auto map_adc = MyConfig.GetParamMap("system.adc");
      auto map_xsq = MyConfig.GetParamMap("xsq");
      map_adc["factor12v"] = adc_factor;
      map_xsq["calc.adcfactor"] = calcADCfactor ? "yes" : "no";
      MyConfig.SetParamMap("system.adc", map_adc);
      MyConfig.SetParamMap("xsq", map_xsq);
```
…really doesn't improve anything, neither performance nor readability. And it gets even worse in `WebCfgADC()`.

Having said that, you made me think about adding the typed & defaulting getters & setters to the `NameMap` & `OvmsConfigParam` APIs. That would enable removing the remaining overhead especially for config change event listeners, as they already get the `OvmsConfigParam` pointer, and it would make working on the map directly also make more convenient.

zorgms:
I hadn't changed either of them initially and opted for a uniform structure. Clearly, changing them wouldn't make sense for the points mentioned.

Should they revert to their old state?

Or should I wait until the new standard helpers are released?

dexterbg:
I've just added the API extension to PR #1310, please have a look. You can now use the typed & defaulting getters & setters on `OvmsConfigParam` and `ConfigParamMap`.

I'm still not sure you will gain readability from using the map API though, but in any case before further changing this PR, I recommend waiting for PR #1310 to be merged. As I didn't get any objections or issue reports, I think I'll do that today.

PR #1332, opened 2026-02-14
Add SD-card OTA flash support

zorgms:
Add an `ota_flash_sd` command (guarded by CONFIG_OVMS_COMP_SDCARD) to download firmware to the SD card (/sd/ovms3.bin.dl → /sd/ovms3.bin) and then flash it to the OTA partition using esp_ota APIs. The routine validates partition sizes, reports progress/status, renames the flashed file to .done, updates the http.mru config. Register the "ota flash sd" command and add a web UI button and JS handler ("Download to SD & Flash") to trigger the flow.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,27 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- OTA: added optional firmware download to SD card before flashing
+    Downloads firmware image to SD card first, then flashes from local file.
+    Useful on slow/unstable connections (cellular) to avoid flash corruption on transfer errors.
+    Uses .dl extension during download, renames to ovms3.<partition>.done on completion
+    (e.g. ovms3.ota_0.done), preserving history for both OTA partitions.
+    URL resolution: explicit argument > http.mru > server config (same as ota flash http).
+    All flash operations run as background tasks with progress polling via "ota flash status".
+    SD card I/O uses 8KB heap buffers + 4KB stdio buffering for ~16x faster transfers.
+    Web UI: "Download to SD & Flash" button added to firmware page (if SD card available),
+    with live progress bar and automatic page reload after reboot.
+    Shows cellular connection warning when on WiFi-AP.
+  New command:
+    ota flash sd [<url>]                  -- Download firmware to SD card, then flash
+    ota flash status                      -- Show current flash operation progress
+  New config:
+    [ota] sd.buffer (bool)                -- Buffer auto/http-update downloads to SD card (default: no)
+    [ota] min.sd.space (int)              -- SD card free space reserve in MB (default: 50)
+    [ota] auto.reboot (bool)              -- Automatic reboot after successful flash (default: no)
+- New vehicle: Volkswagen e-Golf via J533 connector
+    Initial support added for: Odometer, SoC, Est. Range, Charge Status
+    https://docs.openvehicles.com/en/latest/components/vehicle_vwegolf/docs/index.html
 - Module: added heap integrity check + alert command for usage in e.g. custom event scripts,
     enables adding checks on specific system events and/or with higher frequency.
     Example: perform heap integrity check when the server V2 gets stopped:
--- a/vehicle/OVMS.V3/components/ovms_ota/src/ovms_ota.cpp
+++ b/vehicle/OVMS.V3/components/ovms_ota/src/ovms_ota.cpp
@@ -60,6 +60,7 @@ static const char *TAG = "ota";
 #include "crypt_md5.h"
 #include "ovms_vfs.h"
 
+#define OTA_BUF_SIZE 8192  // buffer size for OTA download to SD card
 OvmsOTA MyOTA __attribute__ ((init_priority (4400)));
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -268,7 +269,7 @@ void ota_flash_http(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int arg
   bool reboot = MyConfig.GetParamValueBool("ota", "auto.reboot", false);
 
 #ifdef CONFIG_OVMS_COMP_SDCARD
-  if (MyConfig.GetParamValueBool("ota", "auto.buffer", false) && path_exists("/sd"))
+  if (MyConfig.GetParamValueBool("ota", "sd.buffer", false) && path_exists("/sd"))
     {
     writer->printf("SD buffer enabled: routing download via SD card from %s...\n", url.c_str());
     MyOTA.LaunchOTAFlash(OTA_FlashCfg_SD, url, reboot);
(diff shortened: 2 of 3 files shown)


dexterbg:
ZorG,

your changes entry indicates an aborted download would corrupt flash memory:

> Useful on slow/unstable connections (cellular) to avoid flash corruption on transfer errors.

That's not the case, it simply invalidates the download partition contents. An invalid partition fails the checksum test and cannot be set as a boot partition, so there is no way a user could get into trouble from this.

Also…

> Shows cellular connection warning when not on WiFi.

…does not match the implementation (and also isn't strictly related to this extension).

I understand you want to enable keeping both OTA images intact by this, which is a sensible goal, especially given we're about to reduce the firmware partitions to two.

My issues with the implementation then are:

1. This needs continuous manual interaction by the user. I.e. the next auto update will again use the standard way.
2. Having an extra command a user would normally expect to be just a special case of `ota flash vfs` introduces a source for irritation and accidents.

I suggest reworking this into an option that can be configured for the OTA auto update process, plus a shell option for `ota flash http` to enable the SD buffering mode.

E.g.:
- new bool config `ota` `auto.buffer`, yes = try buffering the auto update download to SD card if possible (fallback to direct writes if not)
- `ota flash http -b` = same for manual updates

To avoid wasting data volume on failing downloads and filling the SD card, I also recommend adding a test for sufficient capacity on the SD card when determining if buffering can be used (see command `vfs df` on how to query the capacity). Enable leaving some headroom, e.g. for users using the SD card to also record logs & CAN traces. The spare size required could be another config, e.g. `auto.buffer.spare`.

Regards,
Michael

zorgms:
This implementation of `ota flash sd` is an alternative to `ota flash http`, the only difference being that the firmware is first written to the SD card and then flashed. The user experience is the same as with `ota flash http`. It is not intended to replace `ota flash auto`.

I have already removed the cellular network notifications, as they were arriving rather randomly. Pure cellular interaction via shell ends with `ota flash http/sd` after the command is sent.

The available free space on the SD card is checked?

```
 if (expected > target->size)
    {
    writer->printf("Error: Download firmware (%d) is bigger than available partition space (%d)\n",
                   expected,(int)target->size);
    http.Disconnect();
    return;
    }

```

Should `expected` or `target->size` have a certain margin during the check so that some space remains free when writing the file?`

What I can imagine is implementing the renaming after flashing to ovms.done as an option and deleting the file by default.

Regards,
Michael

dexterbg:
My suggestion also isn't to replace the standard process, but integrate with it, as it shouldn't be a separate way but just a mode of operation for HTTP updates. Why shouldn't it apply to the standard OTA update, if a user decides to want this?

The target size check does not check the SD card's available space, it validates the OTA partition capacity. The OTA partition does not need to have spare space, as it's dedicated for containing only the firmware, there are no user files there.

zorgms:
Okay, I would implement the buffer option for `auto/http` as a redirect to `ota_flash_sd`. Is that an acceptable solution?

zorgms:
Soweit sind alle Wünsche drin, ich will noch ein paar Tests machen, da nun alles geändert wurde.
Aktuell sieht es so aus, der Hintergrunds Task ist ein Gamechanger:

```
- OTA: added optional firmware download to SD card before flashing
    Downloads firmware image to SD card first, then flashes from local file.
    Useful on slow/unstable connections (cellular) to avoid flash corruption on transfer errors.
    Uses .dl extension during download, renames to ovms3.<partition>.done on completion
    (e.g. ovms3.ota_0.done), preserving history for both OTA partitions.
    URL resolution: explicit argument > http.mru > server config (same as ota flash http).
    All flash operations run as background tasks with progress polling via "ota flash status".
    SD card I/O uses 8KB heap buffers + 4KB stdio buffering for ~16x faster transfers.
    Web UI: "Download to SD & Flash" button added to firmware page (if SD card available),
    with live progress bar and automatic page reload after reboot.
    Shows cellular connection warning when on WiFi-AP.
  New command:
    ota flash sd [<url>]                  -- Download firmware to SD card, then flash
    ota flash status                      -- Show current flash operation progress
  New config:
    [ota] sd.buffer (bool)                -- Buffer auto/http-update downloads to SD card (default: no)
    [ota] min.sd.space (int)              -- SD card free space reserve in MB (default: 50)
    [ota] auto.reboot (bool)              -- Automatic reboot after successful flash (default: no)
```

dexterbg:
Okay… so you ignored my suggestion to integrate with the standard process but decided to fully replace the standard process instead.

My first issue with this is, that Mark @markwj is currently working on the OTA component to enable repartitioning as announced on the developer list. A minor local extension within just the flash functions would be easy to merge/rebase for him, but a more or less full reorganization of the functions adds a lot of additional effort for him. Mark would need to tell if he is OK with this, but let's first look further.

My second issue, and a major no-go, is that your rework adds an additional mandatory requirement of available internal 8-bit memory to the flash process by removing the option to perform a flash process within an existing context (console / mongoose) -- the background flash task needs at least 6KB for the stack, plus TCB etc.

Flashing is a critical operation that must remain usable as long as possible under tight memory conditions. 8-bit memory requirements vary widely by vehicle implementation and additional customizations, and we need to take non-public applications of the system into account here as well. Removing the option to reuse an existing context for the flash process may render OTA flashing impossible for some users.

Currently users can still perform an OTA update if memory is too tight for the auto update task or the web UI. With this rework, that option won't be available anymore.

Regarding flashing via the web UI, the new way to display the flash progress in the web UI unnecessarily adds to this issue: the status query (`ota flash status`) is executed as a web command stream within its own dynamically created context, adding another required 6KB+ of 8-bit memory here as well. As of now, flashing via web UI already runs as a command stream so has it's own context anyway; adding yet another task doesn't suit any purpose here.

The new status query also doesn't deliver warnings or errors occuring during the flash process to the user. If your goal here would be to provide a nicer graphical feedback, that would better be done by processing the command output via Javascript. You could e.g. extract the progress state and use that to control a percentage gauge, and display any other messages below. But my impression is you needed to introduce this just to get the background task progress to display.

Flashing with SD buffering still is a separate process in your UI & shell design, which I still think is neither necessary nor beneficial for users. And regarding the web UI, you also didn't take the flash process in the setup wizard into account.

Minor side note: naming the config for reboot behaviour on _manual_ flash processes "_auto_.reboot" breaks the context of "auto.*" being associated with the auto update process. Also, the reboot behaviour actually should rather be a command option than a configuration.

Generally try to group related configs using a common name prefix scheme: with "sd.buffer", "min.sd.space" should rather be named "sd.minspace".

Up to here I don't think this can be merged even partial, it's too far off into the wrong direction -- sorry for your effort invested, really appreciate your intent to make an improvement.

And still your original core idea of optionally buffering to the SD card is good. I suggest taking a step back and having another look at my original suggestion of just integrating the SD buffering option into the standard HTTP flash process.

Using larger buffers to speed up I/O is also a good idea, but you (and the newlib function `setvbuf()`) use standard `malloc()` for allocations, i.e. the buffers will also be allocated from precious 8-bit RAM. Allocate buffers using `ExternalRamMalloc()` instead, and pass a preallocated extram buffer to `setvbuf()`.

Regards,
Michael

zorgms:
added SD card mounted check and fixed Firmware file size check.

zorgms:
If Mark is already working on the OTA update, it makes no sense for me to make any changes or adjustments. It should remain as it is so as not to hinder his work.

Regards,
Michael

PR #1333, opened 2026-02-14
Add timeout PrioritiseAndIndicate

zorgms:
To prevent indefinite blocking if the callback is not executed, a 10-second timeout has been added to the TCP/IP callback semaphore in `PrioriseAndIndicate`. An error is logged upon timeout to facilitate troubleshooting. Previously, a missed callback signal could permanently block the calling task and ultimately trigger the task watchdog. This fix eliminated crashes during OTA updates via cellular data.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,10 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Network manager: prevent Task Watchdog crash in PrioritiseAndIndicate()
+    The TCPIP callback semaphore (m_tcpip_callback_done) now uses a 10 second timeout
+    instead of waiting indefinitely. Prevents deadlock when the TCPIP task fails to
+    execute the callback, e.g. during WiFi state transitions.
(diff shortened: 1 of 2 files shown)


dexterbg:
ZorG,

same as on PR #1334, I also see a timeout here as a last resort workaround, and most likely a crash & reboot would be indicated if this happens, as in most cases the network won't work after this operation fails.

> deadlock when the TCPIP task fails to execute the callback, e.g. during WiFi state transitions

I'm quite sceptical about this analysis, pretty sure that cannot happen unless the LwIP task (`tiT`) dies due to some critical LwIP bug, and I have never seen that happening.

Please explain how you triggered this, and post the logs indicating this condition.

Regards,
Michael

zorgms:
After implementing my PR #1334 solution, I only experienced random crashes during "ota flash http/sd" over mobile data. The crashes during Wi-Fi state transitions can only be explained by the module searching for a Wi-Fi network, which is why it crashed. I no longer have any logs.

With this implementation, I no longer have the "ota flash http/sd" crash, but there are still occasional reboots without a crash message. (unknown reboot)

The solution is far from perfect; it only helps a little.

Regards,
Michael

zorgms:
Durch Änderungen beim OTA (#1332 ) ist es auch ohne diesen PR stabil.

PR #1334, opened 2026-02-14
Server V2/V3 MG lock timeout

zorgms:
In server module V2/V3, a 5-second timeout was introduced for all MongooseLock requests. Each call point now checks whether the lock was successful and issues a warning if the timeout occurs. This prevents the OVMS event task from blocking indefinitely while the Mongoose task holds the lock. Previously, this caused the task watchdog to crash during OTA updates over cellular data.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,12 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Server V2/V3: prevent Task Watchdog crashes caused by MongooseLock contention
+    All MongooseLock() calls in server_v2 and server_v3 event handlers now use a 5 second
+    timeout instead of blocking indefinitely. On timeout, the operation is skipped with a
+    warning log and graceful early return, preventing the OVMS Events task from triggering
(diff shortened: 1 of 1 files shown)


dexterbg:
ZorG,

I came to the same conclusion about the nature of the deadlock, and have started developing a fix this week.

I see introducing timeouts more as a last resort workaround than a fix, as this means accepting that even with a fully stable network connection, network operations initiated from event handlers may fail arbitrarily. I'd rather try to eliminate the actual deadlock cause.

My current plan & work in progress is to reduce the mongoose lock coverage & make it more atomic, i.e. unlock mongoose ASAP when back in our code space. That especially applies to the mongoose callback invoking the respective network handler, e.g. executing commands received from the network.

That way the mongoose lock is available when a network handler requires the event lock (e.g. for event registration), so an already running event handler can still perform network operations as needed, and the event lock will become available afterwards for the network handler to finish.

I'd like to try my plan first, and am open for any additional thoughts on this.

I suggest keeping this PR open, in case we need some hybrid solution.

Regards,
Michael

zorgms:
Durch Änderungen beim OTA (#1332 ) ist es auch ohne diesen PR stabil.

Issue #704, opened 2022-03-04
Bolt EV - swcan needed


kallisti5:
I'm not sure of the details to set this up.. but the Bolt EV is need of the swcan?   The ODB2 port goes to "sleep" after a while and doesn't respond to commands.

kallisti5:
https://github.com/mjuhanne/OVMS-SWCAN seems to be what folks are talking about.  Information is limited, but it sounds like swcan needs to be wired to the private GMLAN bus?

I plan on trying to fully remove my Onstar module at some point (and wiring the microphone to restore bluetooth microphone support)  Guessing the OnStar module would have unrestricted GMLAN access?

kallisti5:
I found some documentation of the OnStar connectors online from the Bolt EV...   Looks like GMLAN low and high speed is readily available at the Onstar box...

![X1](https://user-images.githubusercontent.com/100110/156850857-50a3eb06-ccbe-4b19-bb7e-7e0aee86e757.png)
![X2](https://user-images.githubusercontent.com/100110/156850862-1556cec7-b10f-4718-ae7f-9563166f920b.png)

kssmll:
SWCAN(Low speed GMLAN) is on pin 1 of the OBD2 connector.

kallisti5:
> SWCAN(Low speed GMLAN) is on pin 1 of the OBD2 connector.

I think in the Bolt EV you can't access the whole set of GMLAN commands however. There is a gateway between the OBD port and the ECU to prevent access to door lock control , etc.

kssmll:
First of all, you need swcan for car's wake up. And GM cars does not use gateways. What  set of GMLAN commands you talking about?

kallisti5:
> First of all, you need swcan for car's wake up. And GM cars do not use gateways. What set of GMLAN commands you talking about?

https://www.chevybolt.org/threads/can-bus-commands-to-remote-start-the-bolt.33637/#post-544071

Msbog:
Hello, I want to confirm that sending commands via SWcan located behind the mirror of the camera connector
( black wire : GND and green wire : SWCAN)  works and can start air conditioning, heating, unlocking, locking, etc. Tried!

kallisti5:
> Hello, I want to confirm that sending commands via SWcan located behind the mirror of the camera connector ( black wire : GND and green wire : SWCAN) works and can start air conditioning, heating, unlocking, locking, etc. Tried!

[...]

PR #1335, opened 2026-02-20
Add optional SWCAN module initialization by vehicle

kssmll:
SWCAN must be initialized manually from the car module. During initialization, CAN3 will be deinitialized and CAN4 used for SWCAN. Configuration "use_swcan_adapter" added to voltampera and boltev.
The module added to default build configuration.

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/can.cpp
+++ b/vehicle/OVMS.V3/components/can/src/can.cpp
@@ -1283,12 +1283,12 @@ canbus::canbus(const char* name)
 
   using std::placeholders::_1;
   using std::placeholders::_2;
-  MyEvents.RegisterEvent(TAG, "ticker.10", std::bind(&canbus::BusTicker10, this, _1, _2));
+  MyEvents.RegisterEvent(m_name, "ticker.10", std::bind(&canbus::BusTicker10, this, _1, _2));
   }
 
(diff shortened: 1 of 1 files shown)


dexterbg:
Alexander,

this looks all good to me, except for one thing: the event deregistration for `TAG` in `canbus::~canbus()`.

This is wrong because the tag "can" is currently shared by all buses, and deregistering it will remove all event listeners for that tag, so disable `BusTicker10()` for all other buses:

```
OVMS# event list 
…
ticker.10:  housekeeping, can, can, can
```

To fix this, I suggest changing the event registration to use the bus name (`m_name`) instead of `TAG`, and accordingly use `m_name` for the deregistration as well. That way each bus will deregister only its own listener on deletion.

Regards,
Michael

PS: I saw you also set the LED control option to false by default, am I right to assume that solves the issue reported by @jamesb2147 in https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/issues/704#issuecomment-3882650962 ?

kssmll:
Michael, thanks for the review. I’ve added a fix for event registration/deregistration. As for the LEDs, yes, this should resolve that issue. Moreover, as I understand it, those LEDs were used for debugging and are not needed in the release version.

PR #1336, opened 2026-02-21
Add optinal SD-buffered OTA update

zorgms:
OTA updates can optionally be saved to the SD card first and then flash from file

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_ota/src/ovms_ota.cpp
+++ b/vehicle/OVMS.V3/components/ovms_ota/src/ovms_ota.cpp
@@ -62,9 +62,10 @@ static const char *TAG = "ota";
 #include "ovms_malloc.h"
 
 #define SOFAR 300000                    // update progress callback interval in bytes
+#define TASK_MEM_SIZE 3584              // size for auto OTA flash task (in bytes) - set to 3584 after testing with 2048 to 4096. old value 6144
 #define SD_WRITE_BUF_SIZE 2048          // buffer size for SD write operations in internal RAM - setvbuf
 #define EXT_RAM_BUF_SIZE 8192           // buffer size for OTA operations in external RAM - ExternalRamMalloc
-#define OTA_MINIMAL_SIZE 2*1024*1024    // Minimum expected size of a valid firmware file (>2MB)
+#define OTA_MINIMAL_SIZE 2097152        // Minimum expected size of a valid firmware file (>2MB)
 
 OvmsOTA MyOTA __attribute__ ((init_priority (4400)));
 
@@ -1287,7 +1288,7 @@ void OvmsOTA::LaunchAutoFlash(ota_flashcfg_t cfg /*=OTA_FlashCfg_Default*/)
     }
 
   xTaskCreatePinnedToCore(OTAFlashTask, "OVMS AutoFlash",
-    6144, (void*)cfg, 5, &m_autotask, CORE(1));
+    TASK_MEM_SIZE, (void*)cfg, 5, &m_autotask, CORE(1));
   }
 
 bool OvmsOTA::AutoFlash(bool force)


dexterbg:
Mostly perfect, and mergable without interfering with Mark's work. I've got some minor issues/recommendations, but will fix/optimize them myself after merging.

I think `ota_flash_vfs()` and `AutoFlashSD()` should probably also make use of the `FlashPartitionFromFile()` utility, but that can be done later on. Until then, your task stack size reduction is too optimistic, as these flash methods need more stack. I'll adjust that as well.

dexterbg:
Zorg, I need to revert this. This needs more testing, I've found serious & not yet explainable issues with this: first of all, downloading to SD takes ~ factor 13 longer (~455 seconds) than flashing directly with the previous build (~35 seconds).

Second, and more critical, downloading to SD has now frequent premature aborts:
```
Downloading to SD... 66% (2712284 bytes so far)
Error: Download size (2748184) does not match expected (4080704)
…
Downloading to SD... 51% (2103008 bytes so far)
Error: Download size (2114908) does not match expected (4080704)
```

My SD card has no known issues, I currently have no idea what's the cause for this.

I also see flashing directly is slower than before, taking ~50 seconds now (~35 before). So the buffering seems to be fundamentally not working as expected. I thought the OvmsHttpClient using a smaller buffer & by default streaming the polled data is the bottleneck, but changing that didn't help.

In conclusion, this isn't ready to be given even to "edge" beta testers yet (let alone users). Maybe I'll pull this into a testing branch for further investigation.

markwj:
I'm really wondering what is the use case of this? I am aware of the issue of downloads occasionally failing (particularly over cellular), but that could be fixed by having a resumable download (the HTTP transfer can be restart at an offset, and we can keep a track of the last good offset in 4KB chunks that we successfully wrote to flash).

web server -> SD card -> flash seems overly complex (given that we have an OTA flash partition free and ready to do whatever we want with).

dexterbg:
According to Zorgs initial PR #1332 on this, he wants to keep both OTA partitions intact (bootable) when a download aborts.

An added benefit could be having the option to keep some firmware images on the SD card for later reflashing without download. Both are rather special cases not occurring with regular users (most of those won't even have an SD card installed).

Another solution could be to add just a general file download utility, so users can download firmware images and then simply use `ota flash vfs`.

A restart option for aborted downloads would help keeping data volume low for mobile-only users, so may make sense independent of this.

zorgms:
@dexterbg 
I've also noticed that downloading to the SD card is extremely slow. The more write cycles, the slower it gets, hence the 2kb buffer for batch writing. A buffer larger than 2kb would be somewhat better, but there was no speed increase with anything over 4kb.

I'm only familiar with download interruptions from mobile data, so in my previous PR, I ran all update options, including automatic updates, in a background task, which eliminated most download problems.

Could the slower OTA update be due to ExternalRamMalloc? That's the only change in the download-to-flash process.

@markwj 
The ability to resume the download would be fantastic.

dexterbg:
The `OvmsHttpClient` internal buffer has always been in SPIRAM, so no. One of my uncommitted enhancements for this is to also allow `OvmsHttpClient` with a larger buffer, and to switch the reader into a buffering mode, but none of that helps. I've checked the `esp_ota` & flash writes to see if they need to reflash partial sectors, but from my first look they take partially filled sectors into account and postpone writing them until more data has been received. So… no idea currently.

dexterbg:
OK, the actual flash/file write was done from the heap after copying from the HTTP buffer… I'll test that.

markwj:
> According to Zorgs initial PR #1332 on this, he wants to keep both OTA partitions intact (bootable) when a download aborts.

That requirement seems unrealistic. Re-writing these flash partitions can never be an atomic operation. There will always be a delay between erasing the flash and writing the new contents. For example, even just re-writing the partition table (one single 4KB block) has the risk of corruption.

Also unnecessary. If we are re-writing the new firmware there, who cares that the old contents are maintained up until the time the re-write completes?

> An added benefit could be having the option to keep some firmware images on the SD card for later reflashing without download. Both are rather special cases not occurring with regular users (most of those won't even have an SD card installed).

OK. This I understand.

> Another solution could be to add just a general file download utility, so users can download firmware images and then simply use `ota flash vfs`.

This is the better solution, imho. Something like a CURL or WGET for downloading via http to SD Card.

zorgms:
I've now conducted numerous OTA HTTP download tests using mobile data, both direct flashing and buffering via SD card. Both methods were rarely successful and often resulted in a crash.

I've moved the OTA HTTP download to a background task, similar to OTA AUTO. This allowed the OTA HTTP download to run almost flawlessly, and even if it was interrupted, there was no crash. Therefore, I would prefer this solution and keep the SD card option available.

https://github.com/zorgms/OVMS-Firmware-EDGE/tree/ota_http2task

dexterbg:
I'd like to replicate your test. How did you run the `ota flash http` command, i.e. in which environment/shell? Which other components & connections were active, what was the Wifi mode?

dexterbg:
When testing the partition upgrade, I also ran into the `ota flash http` issue. This happened on the USB console:

```
OVMS# ota flash http 
Current running partition is: ota_0
Target partition is: ota_1
Download firmware from ovms.dexters-web.de/firmware/ota/v3.3/edge/ovms3.bin to ota_1
Expected file size is 4080704
OTA Flash HTTP: Preparing flash partition...
Downloading... 4% (200408 bytes so far)
Downloading... 9% (400424 bytes so far)
Downloading... 14% (600440 bytes so far)
Downloading... 19% (800556 bytes so far)
Download failed (at 879948 bytes)
Error: Download file size (879948) does not match expected (4080704)
```

I probably never had that issue before because I always do manual updates via the web UI, so the command is running in a task.

Quite strange this happens on the USB console, I expected this to be a Mongoose context issue. Flashing from vfs worked normally from the console.

zorgms:
I almost exclusively use the shell via your website to initiate module updates over mobile data. `ota flash auto force` (rarely fails) or, for test versions, `ota flash http <url>` (rarely works).

When using HTTP, I saw a UART overflow in the USB console, but I assume this is just a reaction. The code runs without problems in the task, so I don't suspect a code error.

zorgms:
Normal `ota flash http`:

image

Background Task `ota flash http`:

image

dexterbg:
I wrote:
> When testing the partition upgrade, I also ran into the ota flash http issue.

I was wrong: that wasn't this issue, it was caused by my ISP having had a high packet loss rate.

Zorg wrote:
> I almost exclusively use the shell via your website to initiate module updates over mobile data.

That's using the V2 command function, just like the App, and that is the actual culprit here. USB console & the module's builtin web server don't have the issue (that's how I do manual updates).

This isn't an issue of the OTA framework or of a mobile data connection, it happens with any command running for too long now. You can reproduce this with any custom script running for more than 60 seconds, or even a simple "sleep 70".

When you run commands via the V2 or V3 server command channel -- or via SSH --, they are executed synchronously by the Mongoose receive callback, i.e. within Mongoose (NetMan) context. With Mongoose now being protected against concurrency, that means all other Mongoose clients need to wait for the command to finish. As server V2 & V3 need to transmit data from ticker events, they get stuck there, in turn blocking all event processing and thus triggering the watchdog after 60 seconds.

Long running commands "worked" before introducing the Mongoose lock, but at the risk of corrupting Mongoose internals. I designed the web command handler to use individual tasks to avoid blocking, so that doesn't have this issue, but at the cost of some 8-bit memory overhead. V2, V3 & consoles OTOH are optimized for as little 8-bit memory footprint as possible, and we should try to keep that.

Blocking Mongoose for more than a couple of seconds is the real issue here, that needs to be avoided.

Using a task for command execution is a possible solution, but if we do that, we need to do it as a general solution, not just to solve manual ota updates. That could be a standard "run" command, that allows to selectively run any command in a new background task. When finished, the command output would be sent as a text notification (possibly with other delivery options).

That would allow to run (start) any kind of critical command from any kind of context, i.e. also from an event script, if available memory permits. A fallback for tight memory situations could be to extend the async console by a background command queue, but I'm not yet sure about that -- it would mean the USB console could get blocked, and that should be avoided.

Another option could be to rework V2, V3 & SSH into not needing synchronous Mongoose access for events. That would probably need quite some changes, and make some things much more complex, notably the connection handling.

The more I think about it, the more I like the idea of a "run" command. I'll probably implement that, unless a better idea is found.

dexterbg:
> The more I think about it, the more I like the idea of a "run" command. I'll probably implement that, unless a better idea is found.

This has now been implemented by commit f2e27b5a659bf3a4c06637619bd9d7a54f978eda -- please test.

PR #1339, opened 2026-02-25
OTA: flash http to task like flash auto

zorgms:
To make `ota flash http` more reliable over mobile networks, I've moved it into a task, similar to `ota flash auto`.

Feedback via the web UI flashing process is now generated through status polling.

#1336


zorgms:
obsolet by run command

PR #1342, opened 2026-03-04
chevrolet_c6_corvette updates

leres:
add TPMS support
add missing RPM case to IncomingPollReply()


dexterbg:
Craig, adding new user level functionality justifies adding a note to the `changes.txt` file.

PR #1343, opened 2026-03-08
e-Golf: Report PRNDL state in standard metrics

ewfuentes:
Start listening to the powertrain CAN bus to grab the PRNDL status and report via standard metrics.

I initially tried listen only mode on CAN2, but I found that after 5-10 seconds, I would stop receiving messages. If I restart the bus with `can can2 start listen 500000`, I would start getting messages again, only for it to drop out after a few seconds again. Strangely, if I use the active mode, I get messages consistently. Perhaps the CAN Controller is getting into an error state that isn't recovered from correctly in the listen state?

I'm not entirely sure what the expected semantics of the `drivemode` metric is, but I'm using it to distinguish between drive and b-mode. I'm happy to do something different.


dexterbg:
Erick, remember: when adding new user level features, add a note to the `changes.txt` file.

PR #1346, opened 2026-03-09
e-Golf: Capture VIN from powertrain CAN

ewfuentes:
As the title states, capture the VIN from the powertrain CAN bus and set the `vehicle.vin` metric. Since the VIN is transmitted as multiframe data, only set the VIN once all three parts have been received.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
@@ -104,8 +104,7 @@ void OvmsVehicleVWeGolf::IncomingFrameCan2(CAN_frame_t* p_frame) {
             }
             break;
         }
-        case 0x6B4:
-        {
+        case 0x6B4: {
             // This message contains the VIN in 3 parts, with the first byte identifying the frame.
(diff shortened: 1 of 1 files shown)


dexterbg:
Setting the VIN once would be sufficient, but you actually keep processing the VIN forever. Deliberately? Doesn't matter much if the VIN has a low frequency.

ewfuentes:
This is the simplest thing that does the job. I'm not familiar enough what happens downstream to know if there are any bad consequences of setting the value repeatedly. 

Currently, I use the first three bits of `m_vin_parts_received` to track with of the parts have been received. I could use the fourth bit to track if the value has been set. So instead of clearing `m_vin_parts_received` after setting the VIN, I would `m_vin_parts_received |= 0x08`. Then we would never enter that if block again. If that sounds reasonable, I can make that change and add a comment explaining what's happening.

dexterbg:
It should be sufficient to add an `if (m_vin_parts_received != 0x07) { … }` check around the whole `0x6B4` frame processing, or alternatively `if (m_vin_parts_received == 0x07) break;` at the beginning of the case block.

ewfuentes:
I made the change you requested. Feel free to merge whenever you're ready.

PR #1347, opened 2026-03-09
Fix remote climate control for Nissan Leaf AZE0_2 (i.e. 2016-2017)

### LINKED ISSUE #301: 2016 Nissan LEAF Remote Climate Control
The [LEAF OVMS docs currently](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_nissanleaf/docs/index.rst#2016-2017-models) state that remote climate won't work on 2016 Nissan LEAF's.

However, testing myself (thanks to @emrysr) I've observed that remote climate trigged by OVMS does work if the [Nissan TCU is unplugged](https://photos.app.goo.gl/MuvpCaXQUjbCdoox6), it's in the same location as earlier LEAF models. However, when the 2016 TCU was unplugged Bluetooth calling and voice commands stopped working, presumably since the internal microphone is now routed thought the TCU for some reason on the 2016 model. 

It would be great to be able to find a work around for this. Has anyone else experienced this? Shall I update the docs?

ta264:
This is my fix for the remote climate control on 2016-2017 AZE0-2 Nissan Leaf (with the later 3G TCU and NissanConnect rather than CarWings)

The wakeup command wasn't correct - capturing the CAN writes from my TCU I found the correct wakeup sequence and applying this makes it work on my 2017 30kWh leaf.

Before merging we should probably ask some ZE0 and AZE0 users to test that their cars still work - I have made a change there to only send the wakeup frames once before repeating the command.  This is how my TCU is operating.

Please let me know if you have any comments or I should change the code style at all -  this is my first time working on OVMS.

TODO: update documentation.  In my testing the TCU needs to be unplugged for remote climate to work reliably.  I believe this will affect the microphone and I hope de-pinning the TCU CAN wires will fix this, though I need to verify myself.

Fixes #301

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_nissanleaf/docs/index.rst
+++ b/vehicle/OVMS.V3/components/vehicle_nissanleaf/docs/index.rst
@@ -137,7 +137,9 @@ Note: If you prefer not to unplug the Nissan TCU, all OVMS functions appart from
 2016-2017 LEAF models (ZE0-2)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**Remote climate control will only work when plugged in and actively charging on LEAF 2016-2017 models.** This is because in 2016 Nissan moved the TCU from the EV CAN bus to the CAR CAN bus.
+To use OVMS to activate remote climate the Nissan TCU (Telematics Control Unit) module must be disconnected from the CAR-CAN bus.  If you remove the TCU completely the hands-free bluetooth microphone won't work.  To retain the hands-free microphone, follow the instructions `here <https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_nissanleaf/docs/index.rst#disabling-the-tcu-for-remote-climate-control>`_.
+
+You can re-use the built in GPS and GSM antennas for OVMS using two Fakra Z male to SMA male plug adapters.  Use these adapters to connect the OVMS GPS to the blue Fakra connector on the car and connect the OVMS GSM to the pink Fakra connector on the car.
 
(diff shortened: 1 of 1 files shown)


dexterbg:
Stylewise, the Leaf module already is a m…ixture. Maybe applying a code formatter to get back to some common ground would make sense, once you're done updating/fixing the functions.

Regarding functionality, I cannot help, don't have access to a Leaf. It will be easier to get feedback if we add this to the `edge` build, so beta testers don't need to build their own firmware. Your choice.

ta264:
Thanks very much for the feedback. I think running a code formatter over the leaf file would be a good idea... Is there a preferred standard?

I'll tidy up the documentation and then it can be merged if you're happy.

dexterbg:
No prefs, use what you like, but you should coordinate this with the other Leaf contributors of the recent past, especially @developerfromjokela , @zhongfu and @wayne-love , so you can agree on a style.

Regarding the tool, have a look at astyle (https://astyle.sourceforge.net/astyle.html), we used that on the e-Up module, rc file here: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/vehicle_vweup/src/.astylerc

ta264:
Thanks, I think this is ready to go now.  I've verified that the hands-free microphone doesn't work with the TCU disconnected but is fine if you de-pin the CAR-CAN wires from the big TCU connector.

I've got a few more changes I'd like to make (charge port unlock command and it looks like there might be a bug in the auto charge config).  I think these might be better in a separate PR and then once I'm done with changes I'll sort out the formatting if that's OK with you.

PR #1351, opened 2026-03-15
smartEQ: OBDII PID query can be switched on/off.

zorgms:
OBDII polling affects the TomTom radio; for example, navigation data has to be loaded and radio stations tuned every time the car is started. We were able to pinpoint the problem to a specific PID address. Therefore, I modified the polling process and divided it into individual PIDs, which can be individually disabled in the web UI Features.


dexterbg:
Looks good to me.

zorgms:
I will let the community test the changes until Friday, as I have neither the time nor a TomTom myself.

zorgms:
I got a bit lost; I'll approach the topic again.

PR #1380, opened 2026-04-14
Increase the maximum number of filter arguments for can log commands

leres:
Take advantage of #1373.

(I do not understand why "min" is 0 for can log monitor and 1 for all the others. I guess I would expect them all to use min=0.)


dexterbg:
min is 0 for the monitor because that command has no mandatory argument, the others do.

PR #1385, opened 2026-04-28
cellular: fix SIM7600 stuck-NetWait on LTE-only carriers + stability hardening

samr037:
Addresses two issues observed on SIM7600 modems used on LTE-only carriers
(e.g. Free Mobile FR with CNMP=39 modem mode):

1. SIM7600 firmware bug: with `AT+CEREG=1` (basic URC mode), the query
   `+CEREG?` returns stat=4 (Unknown) even when the modem IS registered
   on LTE. Combined with LTE-only network mode (no 2G/3G fallback), CREG
   and CGREG also report "Searching", so all three registration indicators
   stay below the `Registered` threshold and the OVMS state machine stays
   stuck in `NetWait` indefinitely.

   Fix: switch SIM7600 init to `AT+CEREG=2` (verbose mode with location
   info). The modem then correctly reports stat=1 (RegisteredHome).

2. CEREG response parser: previously read the character at position +1
   from the first comma. With CEREG=2, query responses become
   `+CEREG: <n>,<stat>,<lac>,...` and URCs become `+CEREG: <stat>,<lac>,...`
   — same comma position, different semantics. The parser now distinguishes
   them by checking the second field length (single digit = stat, multi-char
   hex = LAC).

3. Defensive: added `volatile` to `GsmPPPOS::m_connected` and `m_shutdown`
   (written by LwIP `tiT` task on core 1, read by modem task on core 0 —
   Xtensa can reorder cross-core stores without volatile or atomics).

4. Defensive: increased modem task stack 5120 -> 7168 bytes (observed 75%
   utilization in production) and LwIP `tiT` task stack 3072 -> 4096 bytes
   (callbacks invoke `MyEvents.SignalEvent` which allocates from PSRAM).

Tested on OVMS HW 3.3 with SIM7600 (firmware LE20B04SIM7600G) on Free
Mobile (FR) LTE — modem stays connected through WiFi/cellular handoffs,
no NetWait stalls.


dexterbg:
Sam, welcome :-)

Please also introduce yourself on the developer list, we're mostly harmless ;-)

**## 1-3**

The modem bug workaround is probably ok, but needs to be widely tested first, as another 7600 firmware revision might just happen to have the very same bug on verbose mode, so the fix could make things worse. IOW this needs to wait for the 3.3.006 main release now, to be then included in edge for 3.3.007.

I'd also like to collect some more data on this if possible. The issue hasn't been reported before, although many modules are running the 7600 with the `LE20B04SIM7600G` firmware revision. The 4G only config was introduced specifically for Germany by the smart community, and is assumed to be in wide use there. So maybe the bug rather gets triggered by the provider or the cellular towers in your region.

If you have some other user reports, please add them here. Also, can you suggest a way for users on how to test for this bug, which steps to do & what to look for? Thanks!

**## 4**

The stack size changes are not ok.

Stack is precious internal 8 bit RAM, so all raises need to be considered harmful. A utilization of 75% is totally fine, in fact may hint at a potential savings opportunity, if the level holds with growing uptime.

I'm willing to accept a raise for the modem task, as there are a couple (few!) of stack overflows in "OVMS Cellular" in the crash reports. But that should be done in small steps, so maybe going to 5632 first.

There are no stack overflow reports for "tiT" in the reports, so there is no need to change that stack.

A comment on your note:

> (callbacks invoke MyEvents.SignalEvent which allocates from PSRAM)

The stack allocation of `MyEvents.SignalEvent()` is for the call (params + return address + canary bytes), and the `event_queue_t msg` local object. The latter recently grew by 4 bytes to now 20 bytes, but as the method isn't normally part of a recursion that shouldn't be the issue. If you have done a call stack analysis and found a stack critical execution branch, please share.

Regards,
Michael

leres:
On 5/5/26 02:12, Michael Balzer wrote:
> I'd also like to collect some more data on this if possible. The issue 
> hasn't been reported before, although many modules are running the 7600 
> with the |LE20B04SIM7600G| firmware revision.

All three of my "hot" modules have LE20B03SIM7600M21-A firmware. These 
came from fasttech.com.

I have a cold spare module. It's cell modem is one I hot air reworked to 
upgrade with a 7600 I got from ebay. It has LE11B02SIM7500A firmware 
(the label definitely says 7600 so maybe it's not really a 7600?)

I also have a spare modem board that I upgraded, probably the same 
firmware. (The first 4 of the S/N match and the first 9 of imei match.)

If it would be helpful I could test changes with this module. (Maybe my 
modems are for the "wrong" region to be useful?)

		Craig

OVMS# ota
Hardware:          OVMS WIFI BLE BT cores=2 rev=ESP32/3; MODEM SIM7600
Firmware:          3.3.005-857-g650f2640/ota_1/main (build idf 
v3.3.4-854-g9063c8662 May  4 2026 09:28:55, product v3.3)
Partition type:    v3-30 (factory, ota1, ota2, 1MB store)
Partition table:   0x8000
Running partition: ota_1
Boot partition:    ota_1
Factory image:     3.2.018
OTA_O image:       3.3.001-280-gea30eeb7-dirty
OTA_1 image:       3.3.005-857-g650f2640

PR #1387, opened 2026-04-28
ovms_netmanager: working network ping with per-interface routing + dns command

samr037:
The existing 'network ping' command was gated behind CONFIG_OVMS_DEV_NETMANAGER_PING
and called esp_ping_new_session() which is not provided by this ESP-IDF fork — it was
a non-linkable stub that would have failed to build if the dev flag had been enabled.

This commit:
- Reimplements network_ping on top of LwIP raw sockets (no ESP-IDF version coupling)
- Adds optional flags:
    -I <iface>  source interface (e.g. st1, pp2) for multi-interface routing tests
    -c <count>  number of pings (1-100, default 4)
    -s <size>   payload bytes (0-1400, default 32)
- Adds new 'network dns <hostname>' command (A/AAAA lookup via getaddrinfo)
- Removes the CONFIG_OVMS_DEV_NETMANAGER_PING ifdef gate (now production-ready)

The -I flag uses bind-to-source-IP (LwIP in this ESP-IDF fork lacks SO_BINDTODEVICE),
which lets users probe specific interfaces — useful for debugging cellular path
issues when WiFi is also up and owns the default route.

Tested on HW 3.3 / SIM7600:
- ping 8.8.8.8                  : 4/4 via WiFi  (~10ms, ttl 116)
- ping 8.8.8.8 -I pp2           : 4/4 via LTE   (~70-180ms, ttl 114, src 10.111.x.x)
- ping 8.8.8.8 -I st1           : 4/4 via WiFi  (~10ms, ttl 116, src 192.168.x.x)
- ping 192.168.50.254 -I pp2    : 100% loss (LAN unreachable from cellular — correct)
- ping doesnotexist.invalid     : clean DNS error
- dns google.com                : returns A record
- PPP Connects counter unchanged across all probes (no session disruption)

### DIFF
--- a/vehicle/OVMS.V3/main/ovms_netmanager.cpp
+++ b/vehicle/OVMS.V3/main/ovms_netmanager.cpp
@@ -43,6 +43,12 @@ static const char *TAG = "netmanager";
 #include <sys/socket.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <sys/time.h>
+#include <lwip/icmp.h>
+#include <lwip/inet_chksum.h>
+#include <lwip/ip.h>
 #include "metrics_standard.h"
 #include "ovms_peripherals.h"
 #include "ovms_netmanager.h"
@@ -50,9 +56,6 @@ static const char *TAG = "netmanager";
 #include "ovms_config.h"
 #include "ovms_module.h"
 #include "ovms_boot.h"
-#ifdef CONFIG_OVMS_DEV_NETMANAGER_PING
-#include "ping/ping_sock.h"
-#endif // CONFIG_OVMS_DEV_NETMANAGER_PING
 
 #ifndef CONFIG_OVMS_NETMAN_TASK_PRIORITY
 #define CONFIG_OVMS_NETMAN_TASK_PRIORITY 5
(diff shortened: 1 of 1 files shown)


dexterbg:
Sam,

the current `ping` implementation was introduced with PR #873, meant to be part of the upgrade to a newer IDF version, which didn't succeed since. I doubt we'll be able to update any time soon, so reimplementing `ping` is OK.

Adding the DNS lookup is also a good idea, but I'd rather have the command named as `host` or `nslookup` for similarity with existing systems.

Generally, when adding new user level features, a brief description needs to be added to the `changes.txt` file. See the file for the standard scheme to introduce new commands.

As we're tight on remaining flash space, this probably won't be merged until 3.3.006 has been finalized.

More things to address:

- As you remove all uses of build config `OVMS_DEV_NETMANAGER_PING`, also remove the entry from `Kconfig`.
- Running through the internal `netif_list` from outside LwIP introduces a race condition with a potential use-after-free. Use `OvmsNetManager::GetNetifList()` to get a snapshot of the list from outside LwIP.
- Our standard command syntax scheme doesn't use a space between an option argument and the value, and we normally place options in front of mandatory arguments, even if the actual positioning is free. So the usage template should be `[-I<iface>] [-c<count>] [-s<size>] <host>`.

A suggestion on the interface selection: we never have more than one interface of a kind, and needing to look up the currently assigned interface number first is a PITA especially for scripts. I suggest accepting just the interface name without number, e.g. `pp`, to select the first matching interface of that kind.

The DNS lookup could also do a `PTR` lookup if the argument is an IP address, to come even closer to the `host`/`nslookup` functionality, but that can also be added later on.

Regards,
Michael

Issue #1394, opened 2026-05-07
Can't stop OVMS timed climate control session - Nissan Leaf


DBMandrake:
Hi All,

I'm currently running edge Firmware: 3.3.005-869-ga0afe28f0/ota_0/edge (build idf v3.3.4-854-g9063c8662c May 5 2026 15:38:09, product v3.3-5) and using it with OVMS Connect on a 2016 30kWh Nissan Leaf.

One of the features I love is the timed climate control that was recently added, and I have configured an after work schedule using OVMS Connect.

This works perfectly (including custom runtimes that exceed the car's normal runtime - for example setting it to 30 minutes when on battery, which would normally stop after 15 minutes) except for one small problem I've just noticed - it can't be stopped before the time is up!

If the car is plugged in when the timed schedule runs and I press the climate stop button in the app it is ignored, (or at least the charge point remains active) and if I press the charge stop action the charge and climate control stops, then both start again a few seconds later!

If I unplug the car it stops, then starts again from battery a few seconds later. If I get in the car and turn it on as if I'm about to drive away the car turns on normally, however if I turn the car off again the timed climate control starts again!

If it's running timed climate on battery the stop climate control action does stop it, but it starts again a few seconds later.

If I disable the climate scheduler in the app after the schedule is running then stop climate control remotely it still stops and starts up again.

So it seems nothing short of rebooting the OVMS module can cancel the timed climate control action before the set time period is up. 

It would seem logical that a stop climate control command OR turning on the car should cancel any timed climate control session that is active ? (And perhaps disabling the climate schedule should stop any currently running session as well ?)

dexterbg:
Simon, Zorgs fix (PR #1396) has been merged and is available in edge build `3.3.005-919-gc408fc42c`, please test.

DBMandrake:
Hi - Thanks for the heads up!

I've just tested 3.3.005-922-g1a3378f1b, so a slightly later build than the one you asked for but I can confirm issuing an explicit stop climate control command from a client app cancels the restart timer now.

However, starting the car still doesn't cancel it, so preconditioning comes back on if you turn the car on and off again within the schedule run length. (Including driving the car in between) I'm not sure that's desirable - turning on the car should be a signal that you're done with preconditioning.

I also tested what happens if the car is already turned on when the scheduled time occurs - it reports `vehicle: Climate control already active, skipping` and as you would expect the climate control does not come on if the car is turned off soon after.

[...]

PR #1396, opened 2026-05-08
Nissan Leaf: stop OVMS timed climate control

zorgms:
Added variables to CommandClimateControl to terminate time-controlled climate control when the user wishes to stop it manually.

#1394

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
+++ b/vehicle/OVMS.V3/components/vehicle/vehicle.cpp
@@ -1433,7 +1433,13 @@ OvmsVehicle::vehicle_command_t OvmsVehicle::CommandCooldown(bool cooldownon)
   }
 
 OvmsVehicle::vehicle_command_t OvmsVehicle::CommandClimateControl(bool enable)
-  {
+  { 
+  if (!enable)
+    {
+    // stops the scheduled climate restart
+    m_climate_restart = false;
+    m_climate_restart_ticker = 0;
+    }
 #ifdef CONFIG_OVMS_SC_JAVASCRIPT_DUKTAPE
   if (MyDuktape.DukTapeAvailable())
     {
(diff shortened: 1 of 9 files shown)


dexterbg:
Doesn't this need to be added to all `CommandClimateControl()` vehicle overrrides?
And if so, an API call like  `StopClimateRestart()` at the vehicle class level would make sense.

zorgms:
Yes, strictly speaking, those two variables should be set to `false` in every `CommandClimateControl()` call; however, the need for this has never arisen until now. `StopClimateRestart()` certainly makes sense, but I don't have a globally applicable solution for it that wouldn't require adding it to every single `override`.

dexterbg:
I meant to combine that call with adding it to every override, just to remove the implementation details from the specific vehicle level.

As the scheduler is now a vehicle independent framework feature, the issue basically applies to all vehicles. We could argue only vehicles actually also enabling the web UI (`HandleCfgPreconditionSchedule()`) would use the feature, but the Nissan Leaf example shows that's irrelevant -- the feature is obviously also used if only available via shell (probably because it's also available via the Android App), which tells it's a good feature ;-).

zorgms:
Wouldn't it make more sense to integrate the WebUI globally?
Do you have any ideas for a simple, centralized implementation to ensure that the variables are also set when `CommandClimateControl(false)` is called?

dexterbg:
The issue with enabling the web UI generally is, not all vehicles support climate control. Some need special hardware or connections to support it, for example the UpMiiGos needs the T26 connection, and the Bolts need the external SWCAN module.

zorgms:
I have reviewed the 11 overrides, and I believe that the setting of variables requires a vehicle-specific solution. For instance, in the case of the EQ, I would introduce a distinction such that the `CommandClimateControl()` stop command generates a different response depending on whether `Scheduled Precond` is active or not. Some vehicles can stop the climate control immediately; the EQ and Zoe cannot. Therefore, I do not have a concept for an API; instead, I would propose a code extension for the existing vehicle implementations.

dexterbg:
Sounds reasonable. Ultimately the respective maintainers should take care of this, but the feature isn't in `main` yet, so probably is still unknown to most users.

dexterbg:
The Zoe change is incorrect, as you remove the `OvmsVehicle::CommandClimateControl()` fallback, disabling any potentially registered Javascript plugin implementations for the command. Zoe ph2 doesn't have that fallback, so is OK, as are the others.

dexterbg:
Not yet correct, you're emitting the "Stop CommandClimateControl not possible, Zoe doesnt support it" notification before actually having the final `NotImplemented` result. If a plugin has registered for the command, that may well cover stopping the CC.

PR #1398, opened 2026-05-09
Kconfig: MAXT90 depends on OVMS_VEHICLE_OBDII

### LINKED ISSUE #1397: vehicle_maxus_t90 fails to build: missing OBDII dependency in Kconfig
## Summary

The Maxus T90 EV vehicle component fails to compile when `CONFIG_OVMS_VEHICLE_MAXT90=y` and `CONFIG_OVMS_VEHICLE_OBDII` is unset, with:

```
components/vehicle_maxus_t90/src/vehicle_maxt90.h:4:27: fatal error: vehicle_obdii.h: No such file or directory
```

## Background

The vehicle was added in #1242. The `select` was missing in that PR; the bug went unnoticed because the most common starting template (`sdkconfig.default.hw31`) sets `OBDII=y` independently.

## Root cause

`OvmsVehicleMaxt90` subclasses `OvmsVehicleOBDII` and `vehicle_maxt90.h` includes `vehicle_obdii.h`. However, the `vehicle_obdii` component's include directory and sources are gated by `CONFIG_OVMS_VEHICLE_OBDII` in `components/vehicle_obdii/component.mk`:

```make
ifdef CONFIG_OVMS_VEHICLE_OBDII
COMPONENT_ADD_INCLUDEDIRS:=src
COMPONENT_SRCDIRS:=src
...
endif
```

The MAXT90 Kconfig entry in `main/Kconfig` does not `select OVMS_VEHICLE_OBDII`, so enabling MAXT90 doesn't pull OBDII in. MAXT90 is currently the only vehicle in-tree that subclasses `OvmsVehicleOBDII`.

## Repro

From a clean checkout, with a `sdkconfig` that has `CONFIG_OVMS_VEHICLE_OBDII` unset:

```
cd vehicle/OVMS.V3
make defconfig         # MAXT90 gets enabled by default
make -j$(nproc) all    # fails in vehicle_maxt90.cpp
```

## Fix

One line in `vehicle/OVMS.V3/main/Kconfig`:

```diff
 config OVMS_VEHICLE_MAXT90
     bool "Maxus T90 EV (MT90)"
     default y
+    select OVMS_VEHICLE_OBDII
     help
       Enable to include support for Maxus T90 EV.
```

PR to follow.

kezarjg:
`OvmsVehicleMaxt90` subclasses `OvmsVehicleOBDII` and includes `vehicle_obdii.h`, but the MAXT90 Kconfig didn't declare the dependency. Building with `CONFIG_OVMS_VEHICLE_MAXT90=y` and OBDII unset fails with `fatal error: vehicle_obdii.h: No such file or directory`.

`depends on OVMS_VEHICLE_OBDII` makes MAXT90 only selectable when OBDII is enabled, matching the convention used by the other vehicle Kconfig entries.

Fixes #1397.

## Test plan
- [x] sdkconfig with MAXT90=y and OBDII unset: MAXT90 is dropped on `make oldconfig`; build succeeds.
- [x] Default `sdkconfig.default.hw31`: build still succeeds.
- [x] In `make menuconfig`: MAXT90 only listed when OBDII is enabled.

### DIFF
--- a/vehicle/OVMS.V3/main/Kconfig
+++ b/vehicle/OVMS.V3/main/Kconfig
@@ -370,6 +370,7 @@ config OVMS_VEHICLE_NONE
 config OVMS_VEHICLE_MAXT90
     bool "Maxus T90 EV (MT90)"
     default y
+    depends on OVMS_VEHICLE_OBDII
     help
       Enable to include support for Maxus T90 EV.
 


dexterbg:
This is all correct and yet wrong, because your AI missed that we only use dependencies, not forced selections.

Would you also please write issues and PR descriptions yourself? The AI generated texts are bloated AF, wasting the time of everyone needing to read it.

kezarjg:
Fair point on both. Force-pushed with `depends on` instead of `select`, and reworded the description. Apologies for the noise.

PR #1406, opened 2026-05-23
Allow setting BMS values with DBC / Duktape Polling

frogonwheels:
Support setting BMS configuration in DukTape
Allow setting BMS values with DBC and the Duktape Poller Scripting (that uses the same engine)
Allow specifying array elements as DBC targets

### DIFF
--- a/docs/source/userguide/scripting.rst
+++ b/docs/source/userguide/scripting.rst
@@ -1012,6 +1012,24 @@ The OvmsVehicle object is the most comprehensive, and exposes several methods to
 - ``result = OvmsVehicle.ObdRequest(arguments)``
     Perform OBD/UDS request (synchronous)
 
+- ``OvmsVehicle.BMS.VoltageSetCellArrangement( readings, readingspermodule)``
+    Set the Voltage Cell Arrangement
+
+- ``OvmsVehicle.BMS.VoltageSetCellDefaultThresholds(warn, alert [, max_grad] [, max_std_dev])``
+- ``OvmsVehicle.BMS.VoltageSetCellLimits(float min, float max)``
+- ``OvmsVehicle.BMS.VoltageSetCell(int index, float value)``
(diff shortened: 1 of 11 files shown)


dexterbg:
Michael, nicely done as usual.

I've found a minor naming inconsistence, `VoltagesRestartCell` should be named `VoltageRestartCells` to stay in the general scheme and to match the respective temperature call.

A more fundamental issue is channeling DBC BMS calls through the event system. That would have a significant effect on the system load and performance, saturate the events queue even for small cell counts and thus lead to potentially important events getting dropped, which in turn will trigger a reboot.

The event framework isn't designed for high frequency internal message passing between components, but for high level system state changes. Each event needs a scan for attached scripts, each event needs to be processed by the Duktape engine.

What's wrong with executing the BMS API calls inline from the DBC decoder?

Regards,
Michael

frogonwheels:
Thanks.

I'll fix the naming.

On the event thing:
My goal was to avoid bringing the vehicle (which is where the BMS accumulating and averaging resides) into DBC.  I was thinking that BMS is not measured continuously, but I guess you're getting it in bursts of a few hundred events which might be problematic.

Maybe I just need to set a call-back which can be done in the vehicle constructor/destructor?  That way I can use the same struct for communicating, but just leave it on the stack... and since vehicle.h already includes dbc.h it can reside in dbc.

An alternative would be to factor the BMS statistics functionality out into another singleton that  runs independently of the vehicle class!

Thoughts?

//.ichael

dexterbg:
I still don't see the issue with calling the BMS API directly.

> My goal was to avoid bringing the vehicle (which is where the BMS accumulating and averaging resides) into DBC.

Why? The BMS API is defined at the `OvmsVehicle` class level, so is generalized. If the vehicle's DBC file defines BMS processing, the vehicle obviously wants the data to be passed to the BMS API, doesn't it?

Issue #1405, opened 2026-05-19
Smart EQ contactor switching cycles glitch


glynhudson:
I've just been made aware of a glitch on the Smart EQ which can cause the contactor cycle counter to reset to 0. This bricks the HV. Apparently, the glitch can happen if the 12V runs low, or if unknown devices are detected on the ODB2 port. There has been a recorded case of this happening to an OVMS user. Although it's unknown if OVMS was the cause of this issue, but I wonder out of caution should OVMS use should be discouraged on these vehicles? 

## Documented examples: 

- https://www.smart-emotion.de/forum/thread/4407-bms-glitch-for-contactor-switching-cycles/
- https://youtu.be/9ln-2q_ExEQ?si=Bd0lsy_UMCF1rrz4

## Potential Error codes:

P1D2771, P1D277, P1D2785, P1D2871, P1D2877, P1D2885, P0DE71C

## Fix 

It's possible to reset the counter on the BMS, but unfortunately the BMS needs to be removed from the battery pack to do this. The video above documents this process. 

These companies are offering a service to reset the BMS once it's removed

- https://www.smart-emotion.de/shop/product/67-smart-453-bms-reset/
- https://evclinic.eu/product/smart-451-ed-bms-a7899010100-repair/

dexterbg:
Pulling in @zorgms @Dimitrie78 

I remember some smart OVMS user contacting me about an unusal click sound he had with the OVMS attached. I suggested to first try disabling CAN write access, and if that doesn't help unplug the module. Unfortunately haven't heard again from him.

I see you already read the contactor cycles and provide these in metric `xsq.bms.contactor.cycles`. Would it help to monitor the cycle count and raise an alert if the counter changes by a higher than expected rate, or drops below a critical threshold?

Or is the glitch of the kind where the remaining life time just instantly drops to zero? (That would mean you basically cannot attach _any_ device to the OBD port without risking this to happen…?)

Also found this in the thread:
https://www.smart-emotion.de/forum/thread/4407-bms-glitch-for-contactor-switching-cycles/?postID=52027#post52027
> The counter-zero thing was only ever observed with a permanently installed faulty OBD plug triggering them [faults] constantly.

Is that a dependable source & observation?

dexterbg:
Btw, if you add a data notification for the cycle counts, I can provide an anonymized data collection over all smart EQ registered on my server. That would enable you to analyse the counter state distribution and look for outliers.

[...]

PR #1408, opened 2026-05-24
Smart EQ - prevent polling on start.

Dimitrie78:
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/issues/1405#issuecomment-4528368211


dexterbg:
@zorgms please check.

zorgms:
Looks like https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1409/changes/5d30277d3f2a9191c25f22cc2612ff614b6fb0f1, but `can_init=false` makes no sense it will set in code to false and never set true.

Btw: #1409

dexterbg:
Has this been obsoleted by PR #1409 ?

PR #1409, opened 2026-05-24
smartEQ: HV cycles alert, pre cond add IsCANwrite check...

zorgms:


dexterbg:
Please change the data record back to the values being the actual readings, not the interpretation. An interpretation may always turn out to be incomplete or wrong later on, readings still are valid.

Also, add the odometer as a new field at the end to the data record. Data records need to be extended at the end to be compatible with previous versions, if a different structure is needed, changing the version number is not sufficient. V1 is already published now, as I needed to publish a build to compensate a can1 error from another PR.

Generally, as the issue is discussed in terms of a countdown to zero everywhere (including here), we shouldn't confuse users by introducing the reverse interpretation now. In the alert, using the "cycles passed" interpretation will especially confuse, as the number would now rise while the alert says it stepped down. Adding the "cycles passed" interpretation at the new index to the metric array is OK, but also redundant.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
```diff
@@ -322,6 +322,4 @@ void OvmsVehicleSmartEQ::smartCAN2Metrics()
   if(can_trip_energy < 3000.0f) // prevent unrealistic values based on faulty readings
     mt_reset_energy->SetValue(can_trip_energy);
   mt_reset_speed->SetValue(can_avg_speed);
-
-  can_awake = false; // dirty fix to prevent active polling when the CAN bus is already asleep.
```
Dimitrie78, line 326: This served to prevent the CAN bus from remaining active when no CAN frame (350 C0 00 00 00 00 00 00) is received.
While your check for the frame (350 C1 00 00 00 00 00 00) does indeed activate "can_awake," if the corresponding deactivation message is missing, the poller will continue running for 24 hours.
In case you find a better way to prevent this behavior:
If there is no activity on the bus, it should stop automatically after a certain period—regardless of whether the stop frame arrives or not.

zorgms: Ok, ist nachvollziehbar.
Mein Vorschlag ist, dies mit einem ticker abzufangen, damit kann man die weiteren variablen auch auf false setzen.

```
if (can_350_ticker > 0)
    {
    if (--can_350_ticker == 0) 
      {
      ESP_LOGI(TAG, "CAN 0x350 timeout reached");
      can_awake = false;
      can_env_on = false;
      can_battery_on = false;
      mt_bcm_vehicle_state->SetValue("CAN 0x350 timeout reached");
      }
    }
```

dexterbg:
I think something is wrong with cycle reading or processing. Possibly the false zero reading case I saw in the data triggers false alerts now. Example:

```
rec_time,v,cycles_max,cycles_prev,cycles_now,cycles_diff,odometer
2026-05-25 01:08:21,1,200000,0,152247,-152247,0
2026-05-25 09:10:30,1,200000,152247,0,152247,30714
2026-05-25 09:10:46,1,200000,0,152247,-152247,30714

2026-05-25 09:10:30 push notification 'AHV Contactor Cycle: more than expected cycles consumed:|
|  Max Cycles:              200000
|  remaining Cycles:        0
|  Cycles consumed:         200000
|  Cycles diff (last/now):  152247
|Please check the vehicle and contact support!|
|Warning: The counted consumedcycles are above 50000 which is more than expected for a healthy
 HV contactor - please check the vehicle and contact support if this happens repeatedly!|'
```

The vehicle got an initial count correctly at 01:08, then a zero count at 09:10:30 triggering the alert, but then the counter recovered to the correct reading just seconds later.

Only the "more than expected cycles consumed" alert was triggered, not the "stepped down" alert.

Edit: there are more vehicles showing this pattern.

zorgms:
Ich hatte heute morgen auch so eine Schreckensmeldung, das mein Zähler auf null ist.
Der smart wird offenbar immer noch vom alten COM Modul in Intervallen geweckt, ca. alle 3-5h. 
Vielleicht noch um remaining Cycles > 0 erweitern?

zorgms:
#1411

PR #1414, opened 2026-05-26
SmartEQ: improve CAN, polling and contactor handling

zorgms:
Multiple SmartEQ fixes and tweaks: use IsAwakeEQ() for awake metric, make HV contactor cycle processing robust (avoid false positives by using current/previous 'now' values, update logging and alerts), increase command delays (200ms -> 500ms) and reduce wakeup retries, adjust charge start/stop logic (set m_poll_on_charge, conditionally reset trip counters and gate CAN write when m_can_active), refine smartCANmode to properly stop/start bus and clear poll lists, fix Ticker60 ADC recalculation block, and update several poll intervals for various PIDs. Also includes minor formatting/logging improvements.


zorgms:
for notice @Dimitrie78

PR #1416, opened 2026-05-28
smartEQ: Extend BMS contactor log fields

zorgms:
Include additional telemetry in the XSQ-BMS-ContactorLog notification: bms_mileage, can_soc, can_soh, USM_12V, contactor_state, bms_prod_data, bat_serial and evc_traceability. The NotifyStringf format string and argument list were updated to send these values (hold time remains 30 days).

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
@@ -205,16 +205,7 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
     case 0x658:
       {
       REQ_DLC(6);
-      uint32_t bat_serial = CAN_UINT32(0);
-      
-      // Store battery serial number (only if not already set or changed)
-      static uint32_t last_bat_serial = 0;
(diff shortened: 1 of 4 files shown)


dexterbg:
- For historical records like this using the record number as a version, it's good practice to increase the version number on field additions, I suggest bumping it to 2.
- As per `PollReply_BMS_ProductionData()` the `mt_bms_prod_data` string contains a comma by default (CSV meta character), so needs to be either encoded (e.g. using `mp_encode()`) or deliberately interpreted as two values here, in which case it needs to be extended by falling back to "," if empty to keep the CSV record structure intact.
- Same applies to `mt_evc_traceability`, containing two commas by default if set. Encoding is generally necessary for any data record value that may contain CSV or MP meta characters (comma / LF).
- `mt_bat_serial` contains a hexadecimal value without hex base indicator, for easiest processing of the history download with standard spreadsheet software it's recommended to tag hex values with an "0x" prefix or "h" suffix.

dexterbg:
- …and when bumping the version number, it's also recommended to split the documenting header comment, i.e. put "V2:" on a new line followed by the added fields on the next line.

dexterbg:
The throttling will not work correctly that way. I'll explain or -- as you're on vacation -- directly fix this tomorrow.

PR #1420, opened 2026-05-30
Add VIN reading to HIonvFL

KommyKT:


dexterbg:
Please make sure `m_vin` is zero-terminated, e.g. by making it 18 bytes long for your 17 bytes payload.

dexterbg:
Also, there's hardly the need to read the VIN every 60/120 seconds.

KommyKT:
Fixed in latest commit:

* VIN buffer increased to 18 bytes and explicitly null terminated
* VIN polling interval changed to 3600 seconds

PR #1421, opened 2026-05-31
smartEQ: Refine CAN write/polling logic and web UI text

zorgms:
Adjust smartEQ CAN write/polling behavior and related UI copy. Key changes:

- Remove automatic smartOBDpolling(true) invocations from several command paths to avoid switching the CAN bus active unnecessarily.
- Increase retry attempts for climate control/wakeup sequences (loops -> 15) and reduce inter-send delay from 500ms to 200ms; reorder send/check in loops so state is checked before/after writes to improve responsiveness.
- Update smartAwake/smartSleep guards to consider m_can_active and m_enable_write_sleep correctly to prevent unwanted polling toggles.
- Restrict certain polling vectors: only poll full TPMS when vehicle is on (IsOnEQ()), and only add charge-related polls when actually charging (IsChargingEQ()).
- Remove redundant polling activation in CommandCanVector.
- Update web configuration labels and helper text: clarify two CAN write options, refine descriptions for sleep behavior and 12V trickle charging.


zorgms:
#1414

dexterbg:
Just in time ;-)

PR #1423, opened 2026-06-03
smartEQ: Add 12V trickle-charge count and alert

zorgms:
Track 12V trickle-charge activations and alert on frequent activations. Added a new metric xsq.12v.trickle.count and a deque to store activation timestamps; increment the count when trickle charging is activated and prune old entries in a new Ticker3600. If trickle charging is activated 3 times within 24 hours an alert is sent and trickle charging is disabled until reboot. Included the new count in the BMS contactor Notify payload (V3 format). Also adjusted Is12VchargeEQ logic and refactored Ticker60 ADC recalibration checks to use the new helper, plus minor ticker comment/formatting fixes.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,7 +1,12 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
-
+- smart EQ:
+    As soon as the 12V trickle charge begins, a timestamp is generated and stored in `m_12v_trickle_charge_times`.
+    The counted value is set in `xsq.12v.trickle.count`.
+    Within `Ticker3600`, a routine removes all timestamps older than 24 hours and updates `xsq.12v.trickle.count` by -1.
+    New metric:
+      xsq.12v.trickle.count             -- Count of 12V trickle-charge cycles in 24h
 
 2026-05-17 MB   3.3.006  OTA release
 - VW e-Up: extended battery health / aging info (with web UI integration): total battery age,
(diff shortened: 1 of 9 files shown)


dexterbg:
Review results:
- You modify the `config.changed` handling to use the previous RAM state as a default. That leads to an inconsistent state when `xsq` parameters are deleted (as e.g. `xsq default` does), where the operational config of the component does no longer match the one configured. A missing/deleted config needs to imply using the default value.
- `Is12VchargeEQ()` is now fully dependent on polling, both `mt_evc_dcdc` and `ms_v_env_charging12v` only get updated from a poll reply handler. That means if polling gets disabled (e.g. automatically) within a 12V trickle charge phase, `Is12VchargeEQ()` will remain `true` indefinitely. If that's unintended behaviour: sticky states like that can be avoided by clearing/resetting critical polling dependent state variables when polling gets disabled.
- The new metric `xsq.12v.trickle.count` should be added to the user guide and the `changes.txt`.

dexterbg:
Btw, add this to the `3.3.006` section of `changes.txt`, as we're still in preparation for the release.

dexterbg:
Another note on the `Is12VchargeEQ()` macro: it's really hard to tell if the logic in `Ticker10()`…
```
  // check 12V charging state for powermgmt system
  if (Is12VchargeEQ() != StdMetrics.ms_v_env_charging12v->AsBool(false))
    {
    StdMetrics.ms_v_env_charging12v->SetValue(Is12VchargeEQ());
…
```
…is meant to work like it does. The comment and code suggest this block is *meant* to copy the Smart state to the metric, but as `ms_v_env_charging12v` itself is part of the `Is12VchargeEQ()` logic, it only does that in one case. I suggest rephrasing this, or at least extending the comment to explain the logic.

PR #1424, opened 2026-06-04
console: fix vfs tail follow-mode use-after-free on session disconnect

kezarjg:
`vfs tail <file>` is a documented VFS command (user guide; the CAN-logging guide also shows using `tail` to follow a growing log). With no line count it enters follow mode by default — "[tail: in follow mode, press Ctrl-C to abort]" — running its Service() loop on its own task, writing to the launching session's OvmsWriter via a raw pointer. The only stop path is the Ctrl-C Terminator callback.

If the SSH/Telnet session closes or drops by any means other than Ctrl-C, the console (OvmsWriter) is destroyed while the task keeps running; the next writer->write() dereferences the freed object → LoadProhibited.

Decoded backtrace (two separate occurrences):

    LoadProhibited, EXCCAUSE 0x1c, EXCVADDR 0x10, A8=0, task "tail"
    VfsTailCommand::Service()  main/ovms_vfs.cpp:756  -> writer->write(buf, len);
    TaskBase::Task(void*)      main/task_base.cpp:161

EXCVADDR 0x10 / A8=0 = virtual call through a freed (zeroed) vtable.

Fix: OvmsWriter tracks its single active follow-mode task; TerminateCommandTask() sets OCS_StopRequested and waits for the task to exit and release the writer before the transport is freed. Called from ~ConsoleSSH/~ConsoleTelnet, with a backstop in ~OvmsWriter(). Cooperative stop+join (no vTaskDelete) so the task is never force-killed mid-write. Generic to any OvmsCommandTask.

Teardown briefly blocks the close handler until the task exits (≤ one ~250 ms poll); a concurrent Ctrl-C during teardown isn't possible on the disconnect path.

Tested on hw3.3-5 (Solterra): built, flashed, repeatedly dropped an SSH session mid-vfs tail → no crash; normal Ctrl-C exit and one-shot vfs tail -N unaffected.


dexterbg:
The fix applies at the wrong level / from the wrong direction.

The potential necessity of a clean termination for running commands from a session disconnect applies not only to running command tasks, but to generally all interactive commands started within a shell. There is no requirement for these to use an `OvmsCommandTask` instance, the common property of interactive commands is to register an insert callback.

Most commands registering an insert callback do not need any cleanup though, as they don't allocate any resources (e.g. simple `y/n` listeners like `module factory reset`). Other interactive commands besides `vfs tail` needing a cleanup would for example be `enable` and `vfs edit`.

So what's needed here is either an extension of the insert callback handler scheme for passing a termination request (e.g. by introducing a specific session termination character code), or a dedicated registry for a command agnostic termination handler similar to the insert callback registry.

Side note: don't add fixes to `changes.txt`, except if they need users to change something on their side.

Regards,
Michael

kezarjg:
Reworked along your second suggestion — a dedicated termination-handler registry on `OvmsWriter`, parallel to the insert-callback registry, rather than the `OvmsCommandTask`-specific path.

`RegisterTerminationCallback(cb, ctx)` / `DeregisterTerminationCallback(cb)` mirror the insert-callback registry; an interactive command registers a termination handler alongside its insert callback. Consoles call 'RunTerminationCallback()` during teardown — before freeing the transport they own — so the handler can release resources / stop tasks while the writer is still valid (`~ConsoleSSH`/`~ConsoleTelnet` before freeing their transport, `~OvmsConsole` before freeing the shared queue so the Bluetooth console is covered too, and a backstop in `~OvmsWriter`).

Handlers wired up:
- `OvmsCommandTask` — requests its follow-mode task to stop and joins it (cooperative, no `vTaskDelete`); generic to any `OvmsCommandTask`, not just `vfs tail`.
- `vfs edit` and `enable` — free their `editor_state` / `PasswordContext`, fixing the disconnect leaks you pointed out (and they deregister on normal completion to avoid a double-free).
- The y/n insert callbacks with no allocated state (factory reset, OTA prompts) register nothing, as you noted.

Also dropped the `changes.txt` entry.
  
Compiles clean; I'll test on hardware (hw3.3-5 Solterra) this weekend — repeating the mid-`vfs tail` SSH-disconnect case plus normal Ctrl-C / one-shot `tail -N`, and disconnects during `vfs edit` / `enable` — and report back.

kezarjg:
I've pushed a revision (it's the current PR head). Short version: while validating the rework on-vehicle I found the previous approach deadlocks on SSH, so I reworked the teardown to fix it. Wanted to walk you through it plainly and get your read on a couple of design calls, since it's in the area you flagged.

Your review points are all addressed. The cleanup is now a generic termination-handler registry, not `OvmsCommandTask`-specific: any interactive command that takes over the session registers a handler. `enable` and `vfs edit` register one (they free their prompt/editor state); the simple y/n listeners like `module factory reset` allocate nothing and register nothing. I took the dedicated command-agnostic registry option of the two you suggested, and dropped the `changes.txt` entry.

**The deadlock (new; surfaced during testing)** 
The follow-mode cleanup stopped the command task and then waited for it to exit before freeing the transport. On SSH that's creates a deadlock: `~ConsoleSSH` runs on the Mongoose task, which holds the Mongoose lock for the whole `mg_mgr_poll()`, while the `vfs tail` task's `write()` → `SendCallback` needs that same lock. So the teardown waits forever for the task, and the task waits forever for the lock. The SSH server wedges and only a power cycle recovered it. Dropping a session mid-`vfs tail` reproduced it every time once the log was actively growing (a static/idle file didn't).

**The fix; "signal and reap" instead of "stop and wait"** 
When an SSH connection closes with a follow-mode task still bound, I now:
- mark the console *closing*, so the task's in-flight `write()` bails out immediately without taking the lock or touching the (about-to-be-freed) connection;
- ask the task to stop, but don't wait for it; and
- defer freeing the console. A reap pass on the listener's poll cleans it up once the task has actually exited.

`SendCallback` is also re-keyed off the `ConsoleSSH*` rather than the `mg_connection`, so it can never dereference a connection Mongoose has already freed. I kept this **SSH-local** on purpose: Telnet/USB/Bluetooth keep the existing join, since their send paths don't take the Mongoose lock and don't deadlock.

**Validated on-vehicle** 
(hw3.3‑5 Solterra): abrupt SSH disconnects during an active `vfs tail` now survive every time (it wedged 100% before); free heap is flat over 15 disconnect cycles (the reap frees the deferred consoles; no session-slot leak); and Ctrl‑C abort, one-shot `vfs tail -N`, and disconnects during `enable` / `vfs edit` / `module factory reset` are all unaffected.

**Two calls I'd value your view on:**
1. **SSH-local vs. general.** I scoped the defer-and-reap to the SSH console because it's the only transport that actually deadlocks; the termination-handler registry itself stays generic, as you suggested. I'm happy to generalize the deferral across transports if you'd rather it be uniform.
2. **Reap placement.** I run it from the listener's `MG_EV_POLL`. A NetManager ticker or a shared console reaper would work equally well — open to your preference.

(Longer term, routing SSH output through a bounded queue the way `ConsoleAsync` does would mean a follow task never touches the Mongoose lock at all. This change is compatible with that direction but doesn't require it.)

dexterbg:
Sorry for the delay, the Smart EQ issue took longer than expected. We're now clear to move forward.

Your analysis & fix look good to me.

On SSH-local vs. general: Bluetooth and USB of course don't need this scheme, as they don't use Mongoose. But I don't see why this wouldn't apply to the Telnet console as well, as that also deletes instances from the `MG_EV_CLOSE` event, so holding the lock concurring in the same way as the SSH console. If you actually tested this and didn't get the deadlock, maybe that's due to Telnet being much faster, so the race condition is less likely to occur.

As we still support building with Telnet enabled, and other Mongoose based consoles might be needed in the future, we still should factor out the reap queue scheme support into a common base class and apply this to the Telnet implementation as well. No need to move the reaper into Netmanager though, should be sufficient to provide an internal call to be included in the `MG_EV_POLL` handler.

kezarjg:
I couldn't get Telnet to compile, and it looks like an existing bug. 

`console_telnet` hasn't built since the `MongooseClient` refactor -- `console_telnet.h` uses `MongooseClient` as a base without including its header, and `NetManInit` calls a `console` global that no longer exists (`console_ssh` was updated at the time, `console_telnet` was missed). Masked because the default presets leave `CONFIG_OVMS_COMP_TELNET` unset, so CI never compiles it. Two-line fix split out into #1470.

kezarjg:
@dexterbg 

Testing this on hardware, it looks like Telnet may not actually need the defer-and-reap -- though I'd value your read on this, since you know the console layer better than I do.

My understanding is that the reap exists to avoid SSH's teardown deadlock: SSH's write path takes the Mongoose lock, so joining the follow-mode `vfs tail` task from `MG_EV_CLOSE` (which already holds that lock) deadlocks. As far as I can tell Telnet's send path takes no Mongoose lock -- `telnet_send_text` -> `TelnetHandler` -> `mg_send`, and the only `MongooseLock()` in the file is the server bind -- so the join seems to just complete: it stops the follow task while the connection is still valid, then frees. If that's right, Telnet shouldn't hit either the deadlock or the use-after-free.

That seemed to bear out in on-vehicle A/B testing (telnet disconnect during `vfs tail`): the join survived 75 aggressive disconnect cycles with flat heap, while defer-and-reap crashed -- `mg_send` on a freed connection, then heap corruption under churn. But I may be missing context here, so I'd rather check than assume: does keeping the join for Telnet (with the reap SSH-only) match your understanding, or is there a reason to apply the reap uniformly that I'm not seeing?

dexterbg:
Telnet not acquiring the Mongoose lock for sends is another bug of the `MongooseClient` rework. When adding the Mongoose locking, I probably mistook the Telnet handler as being called from the Mongoose handler, and no reviewer caught this as well.

I admit I didn't test building with Telnet. Telnet has been excluded from the standard builds for quite a while, and won't be added again -- we just still include the module in case someone needs it for some special application. I'll merge your build fix and fix the missing send lock.

kezarjg:
Do you still want me to factor out the reap queue scheme support into a common base class and just not implement on Telnet for now?

dexterbg:
OK, lock is in place now, and I've merged your changes locally so far to test them. Already the first disconnect of a running telnet session with `vfs tail` lead to the deadlock, so I'd say telnet needs the deferred reaping as well. But please run your tests to confirm.

dexterbg:
The implementation is nice & clean, and my local tests with forced disconnects show the intended shutdown behaviour.

kezarjg:
## Validation

On-vehicle A-B-A on hw3.3-5 (Solterra), both transports. **A** = current upstream master; **B** = A + this PR. Test images were clean upstream + only this change, built with telnet enabled (`CONFIG_OVMS_COMP_TELNET=y`) and `CONFIG_MBEDTLS_HARDWARE_SHA` disabled -- the latter because upstream otherwise `abort()`s in `esp_sha_read_digest_state` during TLS handshakes (task `OVMS NetMan`, via mongoose -> mbedtls), which is unrelated to this change but fires often enough to confound the results. Both build settings were applied identically to the two arms, so the only delta between them is this change.

| Phase | Firmware | Fault cycles | Result |
|---|---|---|---|
| A1 | upstream master | 6 (ssh 3, telnet 3) | **6/6 crashed** |
| B | + this PR | 132 | **0 crashed** |
| A2 | upstream master | 4 (ssh 3, telnet 1) | **4/4 crashed** |

Every baseline crash was task `tail` (7x `LoadProhibited` at `EXCVADDR 0x10`, 3x `InstrFetchProhibited` at PC 0), decoding to `VfsTailCommand::Service()` / `ovms_vfs.cpp:756`. Controls running the same connect/disconnect churn with no follow task bound stayed clean 6/6, confirming the follow task is what makes teardown dangerous.

B's 132 cycles: fast churn (25 per transport) and slow 4-second holds (10 per transport) dropping the session mid-write; the follow-task-self-exit race (8 per transport, abort delay swept 30-240 ms across the 250 ms poll window); connect/disconnect controls; Ctrl-C abort (stopped and session still usable, 10/10); one-shot `vfs tail -N` (6/6); and `enable` / `vfs edit` disconnects. Zero crashes, zero resets, task count 17->17 throughout -- no console or task leak.

Note that `CONFIG_OVMS_COMP_TELNET` is unset in the standard preset, so CI never compile-checks `console_telnet.cpp`; the telnet side here was built and run with it enabled.

Builds on the telnet build fix (#1470) and the `TelnetHandler` send-lock fix, both already in master.

PR #1428, opened 2026-06-06
webserver: firmware setup page - non-blocking version check + layout fixes

kezarjg:
Two related improvements to the firmware setup page (`/cfg/firmware`), in two commits.

### 1. Load the page instantly; check the server version in the background
Previously the page blocked on a synchronous `MyOTA.GetStatus(info)` that makes a blocking HTTP request to the OTA update server for the available version + changelog; adding time to page load when the module is online and the server is slow.

Now the page renders with `GetStatus(info, false)` (local-only, instant) and the server version is fetched asynchronously via a small JSON endpoint on the same handler: `GET /cfg/firmware?action=updatecheck` returns `{version, update, changelog}`; page JS fills in the "…available" field, the update-available indicator, and the changelog once it returns. Reuses `GetStatus` + `json_encode`; no shell-output parsing.

### 2. Firmware setup page layout alignment (cosmetic)
- Wrap the "Firmware version" / "…available" status rows in a `form-horizontal` so their labels right-align and vertically align with their values, matching the tabbed sections (Bootstrap only styles `.control-label` that way inside `.form-horizontal`; these rows were rendered outside the form).
- The "…available" row's value contains buttons (taller than the version text), so that row's label gets extra top padding to line up with the version number.
- Left-align the "Auto flash" numbered list with the column's other text.

### Testing
Built clean (upstream `master` + these two commits only) and flashed to a hw3.1 module:
- `?action=updatecheck` returns valid JSON with `Content-Type: application/json` and `Cache-Control: no-cache`; gated by the page's existing cookie/apikey auth (unauthenticated and wrong-key requests are denied; valid works); it is GET-only (a POST with the query param falls through to normal form handling) and returns before any form side effects.
- JSON escaping verified for quotes, backslashes, control chars, and UTF-8.
- Failure modes (offline / server unreachable / non-200 / empty body) → empty version → "no response from update server"; AJAX error or timeout → "update check failed".
- Layout alignment verified visually in-browser.

Note: invalid UTF-8 in a server response passes through `json_encode` raw (the browser substitutes U+FFFD and still parses); pre-existing `json_encode` behavior, and the OTA server content is trusted/ASCII in practice.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg_firmware.cpp
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/web_cfg_firmware.cpp
@@ -30,6 +30,7 @@
 
 #ifdef CONFIG_OVMS_COMP_OTA
 #include "ovms_ota.h"
+#include "ovms_utils.h"
 
 /**
  * HandleCfgFirmware: OTA firmware update & boot setup (URL /cfg/firmware)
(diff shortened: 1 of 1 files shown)


zorgms:
I’m not evaluating the code itself, but the firmware page loads noticeably faster with the change.

PR #1430, opened 2026-06-11
feat(vwegolf): remote climate control over KCAN BAP

slothish:
This adds remote cabin pre-conditioning for the VW e-Golf, on top of the decode baseline merged in #1369/#1376.

The car's clima ECU is commanded over BAP on the comfort CAN (KCAN). To talk on that bus at all, the module has to join the OSEK-NM network ring first — so this also brings the OCU wake infrastructure:

- CommandWakeup + OCU keepalive (0x5A7) and NM-alive (0x1B000067). Joins the ring at ~5 Hz, bounded by ACK-grace and a hard session cap so we never talk alone on a sleeping ring (that storms the TEC to bus-off). Stands down when the bus goes idle or the OEM OCU is active — it owns 0x5A7.
- CommandClimateControl: 3-frame BAP burst (start params, temperature, start/stop trigger) to clima node 0x25. Wakes a sleeping bus first and defers the burst to Ticker1 after a settle window, so the command dispatch task isn't blocked. Re-targets instead of racing if a new command arrives while a burst is pending.
- hvac metric driven by ClimaRunning evidence with a hold window, and a suppress window after stop so spin-down frames don't flip it back on.

Also in here:

- fix: CAN2 frames were being forwarded into IncomingFrameCan3. That was based on a mislabeled capture and would falsely hold the KCAN ring alive off FCAN traffic. Removed — each bus decodes its own frames.
- test: CRTD replay now defaults to the committed synthetic fixture instead of a gitignored local capture, so the test gate is deterministic. VWEGOLF_CRTD=<path> still replays a real capture.

Tested on my own car (2018 e-Golf): multiple start/stop cycles over the app, BAP ACKs confirmed both ways, including cold start from a fully sleeping bus (KCAN quiet >250 s — module wakes the ring, clima starts). Native test suite covers the state machine, 86/86.

Like the earlier e-Golf PRs this is heavily vibecoded — built with Claude against compiler feedback, the native test suite, and CAN captures from the car. I review and validate everything on the car before it goes in a PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
+++ b/vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
@@ -1,7 +1,7 @@
 # Climate Control via BAP Battery Control Channel — RE Notes
 
 References:
-- https://github.com/karlsen-technologies/smartkar-cano-new — `docs/canbus-reverse-engineering/BAP_PROTOCOL.md` and `BAP_BATTERY_CONTROL.md` (thomasakarlsen; supersedes e-golf-comfort-can for BAP)
+- https://github.com/karlsen-technologies/smartkar-cano-new/tree/2b191b8e9066d494125b3c0338787a347ec8d205/docs/canbus-reverse-engineering — `BAP_PROTOCOL.md` and `BAP_BATTERY_CONTROL.md` (thomasakarlsen; supersedes e-golf-comfort-can for BAP). Pinned to the commit we validated against (2026-07-09).
 - MIB2 firmware RE by SCjona (PR #1430 review thread) — message/field names below marked *(MIB2 FW)*
 - Local BAP framing notes: `vw-bap-protocol.md` (being aligned to standard terminology)
 
--- a/vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
+++ b/vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
@@ -1,95 +1,113 @@
-# Climate Control via BAP — RE Notes
+# Climate Control via BAP Battery Control Channel — RE Notes
 
-Reference: https://github.com/thomasakarlsen/e-golf-comfort-can
+References:
+- https://github.com/karlsen-technologies/smartkar-cano-new — `docs/canbus-reverse-engineering/BAP_PROTOCOL.md` and `BAP_BATTERY_CONTROL.md` (thomasakarlsen; supersedes e-golf-comfort-can for BAP)
+- MIB2 firmware RE by SCjona (PR #1430 review thread) — message/field names below marked *(MIB2 FW)*
+- Local BAP framing notes: `vw-bap-protocol.md` (being aligned to standard terminology)
 
-## CAN IDs
+Terminology (standard BAP; earlier revisions of this doc used ad-hoc terms):
+
+| This doc (old) | Standard BAP | Meaning |
+|---|---|---|
+| "node" 0x25 | **LSG ID** 0x25 | Logical control unit: Battery Control |
+| "port" | **Function ID** | Function within the LSG |
+| "command"/"status push" opcode 2 | **OpCode** 0x02 SetGet | Set value, request confirmation |
+| "rolling counter" | **Transaction-ID** (array header) | Echoed in FSG response |
+
+## Channel
+
+Battery Control channel, LSG 0x25. OVMS acts as an **ASG** (client); the Battery Control Unit is the **FSG** (server).
 
 | Direction | CAN ID (29-bit) | Description |
 |---|---|---|
-| OVMS → Clima ECU | `0x17332501` | Commands |
-| Clima ECU → bus | `0x17332501` | Keepalive polls, 5s status bursts |
-| Clima ECU → bus | `0x17332510` | Command ACKs, full state broadcasts |
+| ASG → FSG | `0x17332501` | Commands (also carries FSG keepalive polls / 5 s status bursts observed on-car) |
+| FSG → ASG | `0x17332510` | Status responses, ACKs, broadcasts |
 
-BAP node address of clima ECU: **0x25**
+## Function IDs (LSG 0x25)
 
-BAP frame encoding (single/multi-start/multi-cont), opcodes, validity rules → `vw-bap-protocol.md`.
+Names per smartkar BAP_BATTERY_CONTROL.md; "obs." column = what we observed on-car (Captures 2/7 etc.).
 
----
+| Func | Name | Dir | On-car observations |
+|---|---|---|---|
+| 0x01 | GetAllProperties | → FSG | empty SetGet poll seen |
+| 0x02 | BAP-Config | → FSG | empty SetGet poll seen |
+| 0x10 | PlugState | ← FSG | — |
+| 0x11 | ChargeState | ← FSG | — |
+| 0x12 | ClimateState | ← FSG | 7-byte status ~1 s; payload[0] `05`=active `00`=idle |
+| 0x13 | (status) | ← FSG | `04 04`=active `04 00`=idle |
+| 0x14–0x16 | timer/schedule slots 0–2 | ↔ | 8-byte records; Ack confirms writes |
+| 0x18 | ClimateOperationModeInstallation | → FSG | start/stop trigger, 2-byte payload |
+| 0x19 | ProfilesArray | ↔ | Battery Control Profiles (see below) |
+| 0x1A | PowerProvidersArray | ← FSG | slot Ack / status observed |
 
-## Port Map (node 0x25)
+**Authoritative stop detection:** Function 0x12 (ClimateState) payload[0] `05`→`00` (works for both OVMS-triggered and schedule-triggered stops). Do NOT rely on `0x5EA` remote_mode — stale in remote mode.
 
-| Port | Dir | Description |
-|---|---|---|
-| 0x01 | → ECU | Status poll (empty SetGet, 2-byte) |
-| 0x02 | → ECU | Status poll (empty SetGet, 2-byte) |
-| 0x12 | ← ECU | Status broadcast (7 bytes, ~1 s) · payload[0]: `05`=active `00`=idle |
-| 0x13 | ← ECU | Status · `04 04`=active `04 00`=idle |
-| 0x14 | ← ECU | Schedule slot 0 (8 bytes) |
-| 0x15 | ← ECU | Schedule slot 1 (8 bytes) |
-| 0x16 | ← / ↔ | Schedule slot 2 (8 bytes); Ack confirms writes |
-| 0x18 | → ECU | Start/stop trigger (2-byte payload) |
-| 0x19 | → ECU | Clima parameters (8-byte payload) |
-| 0x1A | ← ECU | Schedule slot Ack / status |
+## Battery Control Profiles (Function 0x19)
 
-**Authoritative stop detection:** port 0x12 payload[0] `05`→`00` (works for both OVMS-triggered and schedule-triggered stops). Do NOT rely on `0x5EA` remote_mode — stale in remote mode.
+Charging/climate settings live in 4 stored **profiles** (= "Charge Locations" in infotainment): profile 0 = hidden global/immediate profile used by "start now" operations; 1–3 = departure timers. Commands do not carry one-shot parameters — they **write profile fields**, then trigger via Function 0x18.
 
----
+Profiles are BAP arrays; the **RecordAddr** in the array header selects the record format:
+- RecordAddr 0: full profile, 20+ bytes — includes **temperature** at byte 12, encoding `raw = °C × 10 − 100` (22.0 °C → `0x78`), unit byte, lead/holding times, name…
+- RecordAddr 6: compact profile, 4 bytes — `operation, operation2, maxCurrent, targetChargeLevel`. **No temperature field.**
+
+operation flags (byte 0): bit0 charge · bit1 climate · bit2 climateWithoutExternalSupply · bit3 autoDefrost · bits4–7 seat heaters. `0x06` = climate + climate-on-battery.
+
+Full field tables: smartkar `BAP_BATTERY_CONTROL.md`.
 
 ## Climate Start/Stop — 3-Frame Sequence
 
-Must send all 3 frames in order on `0x17332501`.
+All on `0x17332501`. Step 1 is a SetGet on ProfilesArray (0x19) writing profile 0 in **compact (RecordAddr 6)** format; step 2 triggers it.
 
-### Frame 1 — Parameters (multi-frame, port 0x19)
-```
-DLC=8: 80 08 29 59  [counter]  06 00 01
-```
-`29 59` = opcode=2, node=0x25, port=0x19. `[counter]` = rolling byte, increment each call.
+### Frames 1+2 — profile 0 partial update (long BAP message, group 0)
 
-### Frame 2 — Continuation
 ```
-DLC=5: C0  06 00 [temp] 00
-```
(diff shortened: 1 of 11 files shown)


vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
```diff
@@ -0,0 +1,146 @@
...
+```
+DLC=4: 29 58  00 01     (START)
+DLC=4: 29 58  00 00     (STOP)
+```
+
+---
+
+## Port 0x19 Payload Encoding
+
+`[counter] 06 00 01 06 00 [temp] 00`
```
SCjona: how did you find this temp byte parameter? in thomasakarlsen's capture it's `0x20` which would decode to `13.2°C` according to this, which is out of range for the car.

slothish: Good catch — worth explaining. The formula comes from a full-range setpoint sweep on my car: turning the dash temp knob end-to-end while logging KCAN. The J533 BAP status broadcast `0x17330110` (port 0x1B driver zone, 0x21 passenger) tracks the knob in raw steps of 5 per 0.5 °C detent: `0x3C`=16.0 °C, `0x50`=18, `0x64`=20, `0x78`=22, `0xC3`=29.5, `0xFF`=HI — an exact fit to `raw = (°C − 10) × 10`. That capture is referenced in `vwegolf.dbc` (Captures 2/7).

On the `0x20` in thomasakarlsen's trace: his README describes that start command as using "the settings that are stored in the car", and `0x20` (=13.2 °C) sits below the car's 16.0 °C setpoint floor — so I believe it's a "use stored setting" sentinel rather than an explicit temperature.

Caveat I'll add to the doc: the encoding is confirmed on the status broadcast; I haven't yet conclusively verified the ECU honors an *explicit* temp in the command byte (my verify captures only log RX, so my own TX frame isn't in them). I'll validate that on the car and note it in `clima-control-bap.md`.

thomasakarlsen: I do not know if you guys are following our discussions in the ovms discord. But in there I revealed my implementation of things for my own hardware: https://github.com/karlsen-technologies/smartkar-cano-new

In this repo there exists further documentation on the canbus and relevant messages that are relevant to implementing climate controls (and more) which I have yet to migrate to the e-golf-comfort-can repo.

The documentation in this file uses completely different terminology than whats normally used for BAP communication, making this a bit hard to read but I digress.

Quite simply, no this is not a setting called "use settings that are stored in the car". You always use the setting that are stored in the car.

The context you are lacking here is that you are not controlling a single setting "array". You are making modifications to different [battery control profiles](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_BATTERY_CONTROL.md#battery-control-profiles) that are stored in the car. 

When reading/writing to these profiles you have different [record addresses](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_PROTOCOL.md#record-address) (the different record addresses give you more/less properties to work with for the same object, its a system that lets you do "partial" pull/updates to a profile)

Most likely what you are seeing here with 0x20 being 13.2 °C and thus outside the cars range is most likely because you are reading a version of the profile using a record address that does not have the data in the same places (or not even include it) as you would see with the longer record addresses.

thomasakarlsen: It is important to understand that BAP messages over canbus do not act or function like "normal" canbus messages do. Normally CAN is all about broadcasts, but BAP over CAN is a protocol giving you a request-response protocol. 

In short, BAP exposes function calling and property get/set between control units. And with the property comes different types, like string, int, array, etc. so in order to fully implement features using controls over BAP you kinda need to fully implement BAP and its property manipulation layer. Especially if dealing with arrays.

SCjona: yes, the temperature encoding is correct, but according to MIB2 (infotainment) firmware this is not a field in this message. 

this is what MIB2 firmware tells us (reverse engineered with Claude so don't take this as must be correct):

message name: SetBatteryControlProfileListRA6
decode example:
```
22 06 00 01 06 00 20 00:
[0] 22  unknown, but likely not a counter (MIB2 firmware can't produce this value when writing the message)
[1] 06  revision tag = 6  → RA6 record format (for reference when editing the global profile settings in infotainment this byte is 0x00 and the message contains more fields)
[2] 00  array-header field (*(hdr+8))
[3] 01  array-header field (*(hdr+0xc), element count = 1)
[4] 06  profileOperation  = climate(bit1) | climateWithoutExternalSupply(bit2)
[5] 00  profileOperation2 = none
[6] 20  maxCurrent = 0x20 = 32   ← "temp byte" is maxCurrent = 32 A (1 A/LSB, pass-through)
[7] 00  targetChargeLevel = 0 // aka upper charge level target, no-op
```

so in my understanding this frame is required to make sure climate on battery is enabled, not to set a temperature. the car uses the temperature stored in the global profile (this is configured in infotainment where you set the lower charging level). profiles are charging locations in the UI.

Do you have actual OCU climate activation captures (apart from thomasakarlsen's) that support your RE claims? I don't have a VW subscription.

also for the port 0x18 trigger message firmware says:
```
29 58: propID 0x18 → func 0x958, opcode 2 (SET)
00: profileId = 0 (global profile)
01: control flag either 0x00 or 0x01 (boolean)
```

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
```diff
@@ -827,25 +921,177 @@ void OvmsVehicleVWeGolf::SendOcuHeartbeat() {
...
+    // value would otherwise be cast to uint8_t and produce a nonsense temperature byte
+    // sent to the climate ECU. Keep this in int until after clamping.
+    int climate_temp = MyConfig.GetParamValueInt("xvg", "cc-temp", 21);
+    if (climate_temp < VWEGOLF_CLIMA_TEMP_MIN_C) climate_temp = VWEGOLF_CLIMA_TEMP_MIN_C;
+    if (climate_temp > VWEGOLF_CLIMA_TEMP_MAX_C) climate_temp = VWEGOLF_CLIMA_TEMP_MAX_C;
+    uint8_t temp_raw = (uint8_t)((climate_temp - 10) * 10);
+    data[0] = 0xC0;      // multi-frame continuation, channel 0
+    data[1] = 0x06;      // unknown
+    data[2] = 0x00;      // unknown
+    data[3] = temp_raw;  // target temperature: (celsius - 10) * 10
```
thomasakarlsen: This is not the target temperature byte.

You are building a [battery control profile using record address/type 6](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_BATTERY_CONTROL.md#profile-data-structures) which is the short version. 

This version only has these fields:
Byte | Field | Description
-- | -- | --
0 | operation | Operation flags
1 | operation2 | Additional operation flags
2 | maxCurrent | Maximum charging current
3 | targetChargeLevel | Target SOC %

So what you are actually doing here is setting the maximum charging current.

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
```diff
@@ -827,25 +921,177 @@ void OvmsVehicleVWeGolf::SendOcuHeartbeat() {
...
+    uint8_t data[8];
+
+    // Frame 1: BAP multi-frame start → port 0x19 (clima parameters), 8-byte payload.
+    // 0x80 = multi-frame start, channel 0.  0x08 = total payload length.
+    // 0x29 0x59 = BAP header: opcode=2 (Status push), node=0x25, port=0x19.
+    data[0] = 0x80;
+    data[1] = 0x08;           // 8-byte payload follows across this frame + continuation
+    data[2] = 0x29;           // BAP opcode=2, node=0x25 [5:2]
+    data[3] = 0x59;           // BAP node=0x25 [1:0], port=0x19
+    data[4] = m_bap_counter;  // rolling counter; ACK will echo this | 0x80
```
thomasakarlsen: This is not a rolling counter.

In this specific BAP message you are modifying an [array](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_PROTOCOL.md#bap-arrays). This is the array's header, telling the module which array position to start modifying from and what record address type we are using.

See this [example ](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_BATTERY_CONTROL.md#configure-profile-operation-mode)and breakdown of the exact BAP message you are attempting to send.

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
```diff
@@ -827,25 +921,177 @@ void OvmsVehicleVWeGolf::SendOcuHeartbeat() {
...
+    data[5] = 0x06;           // duration: 0x06 = 15 min (confirmed from capture: port 51 BCD
```
thomasakarlsen: Not the duration.

You are building a [battery control profile using record address/type 6](https://github.com/karlsen-technologies/smartkar-cano-new/blob/main/docs/canbus-reverse-engineering/BAP_BATTERY_CONTROL.md#profile-data-structures) which does not include duration.

thomasakarlsen (changes requested):
BAP is not a broadcast protocol where you fire fixed frames. It is a request/response protocol exposing function calls and typed property get/set (int, string, array) between control units.

Remote climate on the e-Golf is not a "command", instead it is a property manipulation on a battery control profile object and then telling the car to run that control profile.

While climatisation will start based on what is being done in this PR, commands being sent to the car will include values that are not within scope and can give undefined behavior.

Without a proper BAP layer OVMS will struggle to keep up with the car's state and properly communicate with it. Controlling the car via BAP is a multi-step process requiring you to send and receive messages to know that the car is ready for the next step, and to check if your commands were successfully received and stored/acted on.

Implications of there being no BAP layer:

  - The temperature feature will not work in this form. The compact RA6 profile does not contain a temperature field.
  - No response handling. There is no BAP header decode, no multi-frame reassembly of replies, and no handling of the Error opcode (0x07). A negative response from the control unit is invisible and failures will happen silently.
  - No state sync. In the BAP protocol the control unit status replies are the source of truth. It pushes state and can reject, clamp, or override what you asked for. Because
  those replies are not decoded, OVMS' state does not reflect what the car actually did. For example: you set 20 °C, the ECU clamps it to its valid range or ignores it and keeps the
  stored setpoint. OVMS still believes and displays 20. The car ends a session on its own (schedule elapsed, battery too low, a fault, another controller), reports it over BAP. OVMS misses it. Any state that lives only in the BAP response stream is invisible.
  - Persistent-state risk. Because this is a profile write, the misencoded maxCurrent is not just a command, it is changing battery control profile 0. This is the value that handles the global current limit of the car. Even though the charger module might clamp this and ensure everything is within limits, it is still not good to write this to the car.

vehicle/OVMS.V3/components/vehicle_vwegolf/docs/clima-control-bap.md
slothish: You're both right — thank you, this untangled it. I've reworked the branch accordingly (commit d3d1da79f, will push after re-verifying on the car):

- The byte I called "temperature" is the RA6 compact record's **maxCurrent**. My firmware was sending the configured °C as amps (21 °C → 110 A). Now a constant 0x20, matching thomasakarlsen's trace, and I've removed the OVMS-side temperature setting entirely since the car takes the setpoint from the stored global profile.
- What I called a "rolling counter" is the array-header ASG-ID/Transaction-ID nibble — which also explains the echo I saw in the Status response.
- Both BAP docs in the branch are rewritten in standard terminology (FSG/ASG/LSG, OpCode, Function ID, RecordAddr) referencing smartkar-cano-new instead of my ad-hoc port/node terms.

The `(°C−10)×10` encoding itself is real but belongs to the 0x17330110 setpoint status broadcast (dash-knob sweep capture) — I wrongly assumed the command reused it.

@SCjona: no, I don't have OCU activation captures — no VW subscription either. My captures are the dash-knob sweep and RX-only verifies of my own module's commands (which were derived from thomasakarlsen's trace). So for the command direction, his RE + your MIB2 firmware decode are strictly better evidence than anything I have; the doc now marks which claims are on-car observations vs. sourced from your work.

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
slothish: Confirmed — it's maxCurrent. Fixed in d3d1da79f (pushing after on-car re-verify): constant 0x20 (32 A), and the OVMS-side temperature setting is removed since it had no wire effect. Regression test now pins the full RA6 record bytes.

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
slothish: Right — array header, not a counter. Comment rewritten per your BAP_PROTOCOL.md array section (fix in d3d1da79f, pushing after on-car re-verify). One on-car note that may interest you: the FSG's Status response echoes this byte with the high bit set, which is how my ACK matching worked by accident — consistent with it carrying the Transaction-ID.

vehicle/OVMS.V3/components/vehicle_vwegolf/src/vehicle_vwegolf.cpp
slothish: Confirmed — RecordAddr 6, not a duration. The "15-minute" correlation I saw came from the car's own stored holding time, not this byte. Fixed in d3d1da79f, pushing after on-car re-verify.

slothish:
Fix pushed as a48073e (the d3d1da79f referenced in the review replies was my local dev-branch commit; a48073e is the same change rebased onto this PR branch, minus the web-plugin parts that aren't in this PR).

On-car verification today: three start/stop cycles over the app, every command ACKed on 0x17332510 (function 0x18 Status echoes the commanded state), blower running on starts. Cabin followed the MIB setpoint — including tracking a change to LO — confirming the stored-profile model: OVMS no longer carries any temperature setting.

slothish:
Thanks for taking the time to break this down, your smartkar docs cleared up in an hour what I'd been guessing at for weeks. Reading profiles with a RecordAddr-0 GET to actually expose the temp setting seams like the natural next step, but that's for another PR.

zac-dydek:
I tested the earlier version from June 11th and it actually messed some settings up in the car.  I was no longer able to adjust the profile for max charge current, temperature setpoint, and allow climate control on battery mode.  I was also no longer able to charge the car!

I did a "factory reset" of the settings for that module and am now able to charge again.  I haven't tested this new version yet.

SCjona:
@zac-dydek that initial version of the PR wrote something like max charge current 120A (for configured target temperature 22°C) into the profile, out of bounds for the user interface and charging controller so they likely errored out.

slothish:
@zac-dydek sorry about that, and thanks for reporting back. Glad the factory reset got charging working again. The bad profile write is fixed in a48073e — OVMS doesn't touch the temperature/current fields anymore.

dexterbg:
Merging now, but I'm pretty sure this needs an entry to the `changes.txt` as well.

slothish:
Added the changelog entry in #1473.

PR #1431, opened 2026-06-15
Energica: expose BMS low/high cell-group numbers (frame 0x203)

Boogmeister:
The Energica BMS broadcasts the lowest- and highest-voltage cell-group numbers in frame 0x203, but the lines that surfaced them were left commented out as // TODO: Test this, so the group indices never reached the metrics layer.
This activates that reporting via two new integer metrics:

xnr.v.b.c.id.min — cell group number with the lowest cell voltage
xnr.v.b.c.id.max — cell group number with the highest cell voltage

These are scalar group indices, distinct from the existing xnr.v.b.c.* min/max/avg voltage metrics (which report the values). I went with custom xnr.* scalars rather than the per-cell ms_v_bat_cell_* array the original TODO referenced, because the BMS only broadcasts the min/max group numbers on 0x203 — there's no full per-cell array available to populate. Exposing the index makes it possible to locate a persistently low or straggling cell group.
Tested live on a 2023 EsseEsse9+ (81S2P pack); the published indices track the BMS-reported min/max cell groups across charging and balancing.

This was completely done using Claude but tested on my own bike.

I have never done anything like this before so I apologies for anything I have done incorrectly.


dexterbg:
When adding or changing things that affect the user level (like adding new metrics), you should add an entry to the `changes.txt` file. You also should keep the user guide up to date regarding vehicle specific info.

dexterbg:
Merging this now so it can be tested, please don't forget the user documentation.

PR #1433, opened 2026-06-16
smartEQ: Add 1h contactor changes count and alert

zorgms:
Track HV contactor activations within the last hour and alert on excessive switching. Added a new element to xsq.bms.contactor.cycles ([4]=1h_count), stored activation timestamps in m_contactor_act_times, pruned by Ticker3600, and incremented on PollReply_BMS_HVContactorCycles. Data log format (XSQ-BMS-ContactorLog) updated to V4 to include the 1h count. Added configurable alert threshold xsq.bms.contactor.1h.limit (default 10) and send a notification when the limit is exceeded. Also updated CLI/status outputs to show the last 1h count, adjusted metric pre-allocation/comments, updated docs and changes.txt, and fixed mt_obd_mt_day_* metric units to Days.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -8,10 +8,10 @@ Open Vehicle Monitor System v3 - Change log
     New metric:
       xsq.12v.trickle.count                  -- Count of 12V trickle-charge cycles in 24h
     updated metric:
-      xsq.bms.contactor.cycles (vector int)  -- [0]=max, [1]=now, [2]=consumed, [3]=diff, [4]=1h_count
+      xsq.bms.contactor.cycles (vector int)  -- [0]=max, [1]=now, [2]=consumed, [3]=diff, [4]=1h_counted_changes
     New config:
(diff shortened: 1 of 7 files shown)


dexterbg:
You're not actually tracking activation counts, but number of detected counter changes in the polling granularity raster.

That *may* be a metric in itself, but it's not the total number of activations registered by the BMS within the last hour, as within the polling raster multiple activations can take place.

A note on the history log header: please keep in mind these are meant to be used with spreadsheets, so field names should be concise, and should not contain spaces. As you saw I already changed "12V_trickle_charge_count within 24h" to "tcc_12v_24h" for the export. Similarly, instead of "contactor_activation_count within 1h" you should name the new field "cycles_diff_1h".

Further notes:
- Please also update the `XSQ-BMS-ContactorLog` record version to 4 as documented.
- Please change the alert construction to use `NotifyStringf()` instead of `snprintf()` to a fixed size stack buffer.

dexterbg:
The changes in `eq_notify.cpp` don't look correct: you're printing `GetElemValue(2)` (consumed) for "changes within 1h" now.

PR #1436, opened 2026-06-21
smartEQ: Add 12V undervolt history & cooldown handling

zorgms:
Record and persist recent 12V undervoltage events and improve wake/cooldown behavior. Introduces new metrics xsq.12v.undervolt.history and xsq.12v.undervolt.history.vec (up to 10 entries) and implements smart12VHistory() to restore persisted entries after reboot and update string/vector metrics. Adjusts ticker logic to immediately enter poll-off on undervoltage, restart cooldowns, and only alert once to avoid log flooding. Adds cooldown polling when a "technical wake up" is detected, tweaks poll state reporting, reduces ADC history buffer size to 10, registers a new preset command key (tpms.value.warning), and bumps PRESET_VERSION.


zorgms:
image

dexterbg:
Note that you've reached 93 of the 100 available persistence slots. If new framework persistence is added in the standard metrics, your persistence requests may fail.

dexterbg:
Also note that the 4MB remaining space is now down to 6784 bytes.

PR #1439, opened 2026-06-23
Fiat eDoblo: poll VIN only when vehicle is on

strangesource:
## Problem

On the tested e-Berlingo / Stellantis EMP2 vehicle, the VIN is available through the standard OBD-II Mode 09 PID 02 functional request on CAN1 (`7DF 0902`), while the vehicle is fully ready/on (`v.e.on = yes`).

The FTDO module previously polled VIN using a physical ECU request (`7e0 -> 7e8`) and also attempted VIN polling in non-ready poll states. That did not match the observed vehicle behavior.

## Root cause / code path

`vehicle_ftdo_polls` defines the FTDO poll list and poll-state timing. `IncomingVINPoll()` assembles the multi-frame Mode 09 PID 02 response into `v.vin`.

## Fix

- Change FTDO VIN polling to standard functional OBD-II:
  - TX `0x7df`
  - RX module `0` so the poller accepts the standard `0x7e8`-`0x7ef` functional response range
  - `VEHICLE_POLL_TYPE_OBDIIVEHICLE`, PID `0x02`
- Poll VIN only in the ON poll state:
  - `{ 0, 10, 0, 0 }`
- When the ON-only frame timeout marks `v.e.on = false`, return the poll state to Charging so ON-only polls stop after the vehicle is no longer ready/on.
- Skip the initial Mode 09 PID 02 record/count byte when assembling `v.vin`.

## Validation performed

Runtime validation on an OVMS module configured as FTDO:
- With vehicle off:
  - `v.e.on no`
  - `v.vin` unset
- With vehicle ready/on:
  - `v.e.on yes`
  - `v.vin` populated from the functional request

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,7 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Fiat eDoblo: poll VIN via standard functional OBD-II request when vehicle is on.
 - smart EQ:
     As soon as the 12V trickle charge begins, a timestamp is generated and stored in `m_12v_trickle_charge_times`.
     The counted value is set in `xsq.12v.trickle.count`.
--- a/vehicle/OVMS.V3/components/vehicle_fiatedoblo/src/vehicle_fiatedoblo.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_fiatedoblo/src/vehicle_fiatedoblo.cpp
@@ -44,7 +44,7 @@ static const OvmsPoller::poll_pid_t vehicle_ftdo_polls[]
     // Pollstate 2 - car is charging
     // Pollstate 3 - ping : car is off, not charging and something triggers a wake
     // OBD2 broadcast                                    Off   ON CHRG PING (cycle time in sec)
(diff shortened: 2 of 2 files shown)


dexterbg:
Review result:
- Polling the VIN every 10 seconds seems unnecessary, what's the reason?
- Why do you switch into `PollState_Charging()` from the 60 second ticker, what is the reason to include that change for the VIN processing?

strangesource:
Thanks for the review.

The 10s interval was mainly to populate VIN quickly during the short ON/ready window, because this vehicle only answered `7DF 0902` while `v.e.on = yes`. But agree that polling every 10s while ON is excessive. I can reduce the interval.

The `Ticker60()` change was intended to leave the ON poll state when the `0x305` ON frame times out, so the ON-nly VIN poll stops once the vehicle is no longer ready/on. But that’s somewhat unrelated to VIN decoding, so I can drop that.

Let me know your preference and I will update the PR accordingly.

dexterbg:
The interval doesn't control how quickly the poll is sent after the poll state change, but how often. All polls without shifts (newly introduced feature) are sent as fast as possible right after the state change. So you can set the poll cycle of such requests to 9999, which is effectively about once per hour, as the poller has a 3600 seconds overall run cycle. If you need to delay the VIN requests, use a shift.

If the ticker change is unrelated, remove it from this PR. Always try to keep PRs focused on a single feature as far as possible, so they can be easily reverted individually if necessary.

PR #1441, opened 2026-06-23
Fiat eDoblo: add BMS pack temperature polling

strangesource:
## Problem

The FTDO profile already polls BMS min/max cell voltage, HV voltage/current, SOH, and available energy, but standard BMS pack temperature metrics and average cell voltage remain empty.

## Reproduction / observation

On the current FTDO vehicle profile, the following standard metrics are present but unset:

```text
v.b.p.temp.avg
v.b.p.temp.max
v.b.p.temp.min
v.b.p.voltage.avg
```

A read-only UDS scan against the e-Berlingo BMS while `v.e.on = yes` showed the BMS supports these identifiers via `0x6b4 -> 0x694`, UDS service `0x22`:

```text
D817 highest_temperature len=1 data=40  -> 24°C
D87D lowest_temperature  len=1 data=3f  -> 23°C
D877 average_temperature len=1 data=3f  -> 23°C
D43D avg_cell_voltage    len=2 data=0f56 -> 3.926V
```

CAN status remained clean after polling: no RX/TX errors, no overflows, no failed TX frames.

## Root cause / code path

FTDO BMS diagnostic polling is defined in `vehicle_ftdo_polls` and decoded in `OvmsVehicleFiatEDoblo::IncomingBatteryPoll()`. The relevant BMS DIDs were simply not included/decoded yet.

## Fix

Add BMS polling and decoding for:

- `0xd43d` -> `v.b.p.voltage.avg`
- `0xd87d` -> `v.b.p.temp.min`
- `0xd817` -> `v.b.p.temp.max`
- `0xd877` -> `v.b.p.temp.avg`

Also update FTDO docs and change history.

## Validation performed

- Runtime read-only UDS validation on Citroën e-Berlingo / FTDO-compatible vehicle with `v.e.on = yes`.
- Confirmed target metrics were empty before implementation.
- Built firmware from clean `origin/master` branch:

```sh
cd vehicle/OVMS.V3
source ~/.espressif/ovms-idf-v3.3-env.sh
make BATCH_BUILD=1 defconfig SDKCONFIG_DEFAULTS=support/sdkconfig.default.hw31
make BATCH_BUILD=1 -j$(sysctl -n hw.ncpu)
```

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,7 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Fiat e-Doblo: add BMS polling for average cell voltage and min/max/average battery pack temperatures.
 - smart EQ:
     As soon as the 12V trickle charge begins, a timestamp is generated and stored in `m_12v_trickle_charge_times`.
     The counted value is set in `xsq.12v.trickle.count`.
--- a/vehicle/OVMS.V3/components/vehicle_fiatedoblo/docs/index.rst
+++ b/vehicle/OVMS.V3/components/vehicle_fiatedoblo/docs/index.rst
@@ -21,8 +21,8 @@ SOC Display                 Yes (incl. battery capacity)
 Range Display               tba
 GPS Location                Yes (from GPS module)
 Speed Display               tba
-Temperature Display         tba
-BMS v+t Display             tba
+Temperature Display         Partial
+BMS v+t Display             Partial
 TPMS Display                tba
 Charge Status Display       tba
 Charge Interruption Alerts  tba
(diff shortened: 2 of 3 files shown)


dexterbg:
Are you aware cell/module min/max/avg metrics are meant to be set indirectly via the BMS API?

You should only set them directly if you cannot get cell details, but according to the poll list comment you can:
> 0xd440 are single cell details, we don't care about them for now

If you intend to implement that decoding, that would be the better option to fill these metrics.

strangesource:
Thanks, I wasn’t aware.

The `0xd440` comment was pre-existing and not validated by me. I only validated the aggregate DIDs (`d43d`, `d817`, `d87d`, `d877`) because they were low traffic and directly matched the currently empty metrics.

But if direct assignment is only the fallback path, I agree this PR should not expand that pattern, feel free to close it. 
It was mainly my attempt to lern my way a bit more around the code and find some easy extensions to get into the code base.

dexterbg:
I suggest having a look at the single cell decoding first. If that turns out to be incompletely documented or hard to implement, we still can merge this as an initial support. You then just need to be aware you need to remove these direct sets as soon as the single cell decoding is added.

PR #1442, opened 2026-06-23
Fiat eDoblo: detect charge pilot and charging

strangesource:
## Problem

The FTDO profile currently derives `v.c.charging` from increasing SoC. Runtime captures show this can leave charging true after charging has stopped and can be triggered by SoC/status changes unrelated to active charging. `v.c.pilot` is not filled.

## Reproduction / observation

Read-only passive CAN captures were taken on a Citroën e-Berlingo / FTDO-compatible vehicle for these states:

- cable unplugged, charge port open
- cable plugged, wallbox available, not charging
- active AC charging
- wallbox stopped while cable remained connected
- cable unplugged after wallbox stop

Observed stable patterns:

```text
State                                  0x4A2             0x5C3
unplugged / no pilot                   mostly 00 80      00 00 03 20 ...
plugged, EVSE pilot available          mostly 00 00      00 00 03 20 ...
active charging                        00 00             d4/d5 ... dynamic
wallbox stopped, cable still connected 00 80             00 00 03 25 ...
```

This shows `0x4A2` byte 1 indicates EVSE pilot availability, not simply the physical cable latch, and `0x5C3` byte 0 distinguishes active AC charging from stopped/idle states.

## Root cause / code path

`OvmsVehicleFiatEDoblo::IncomingFrameCan1()` used the SoC frame `0x3a8` as a charging heuristic and did not populate `v.c.pilot`. The relevant passive CAN frames are already handled in the same function (`0x4a2`, `0x5c3`), so this is the correct place to decode these metrics.

## Fix

- Set `v.c.pilot` from `0x4a2` byte 1:
  - pilot present when high status bits are clear (`(byte1 & 0xc0) == 0`)
  - no pilot for observed `0x80` and transient `0x40`
- Set `v.c.charging` from `0x5c3`:
  - true for observed active charging byte 0 `0xd4` / `0xd5` while pilot is present
  - false for observed stopped/idle byte 0 `0x00`
- Remove the SoC-increase heuristic for `v.c.charging`.
- No `v.c.state` / `v.c.substate` mapping is added in this PR.

## Validation performed

- Passive runtime captures across the states above, saved to SD card and analyzed locally.
- Built firmware from clean `origin/master` branch:

```sh
cd vehicle/OVMS.V3
source ~/.espressif/ovms-idf-v3.3-env.sh
make BATCH_BUILD=1 -j$(sysctl -n hw.ncpu)
```

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -1,6 +1,7 @@
 Open Vehicle Monitor System v3 - Change log
 
 ????-??-?? ???  ???????  OTA release
+- Fiat e-Doblo: add CAN based charge pilot and active charging detection.
 - smart EQ:
     As soon as the 12V trickle charge begins, a timestamp is generated and stored in `m_12v_trickle_charge_times`.
     The counted value is set in `xsq.12v.trickle.count`.
--- a/vehicle/OVMS.V3/components/vehicle_fiatedoblo/docs/index.rst
+++ b/vehicle/OVMS.V3/components/vehicle_fiatedoblo/docs/index.rst
@@ -24,11 +24,11 @@ Speed Display               tba
 Temperature Display         tba
 BMS v+t Display             tba
 TPMS Display                tba
-Charge Status Display       tba
+Charge Status Display       Partial
 Charge Interruption Alerts  tba
 Charge Control              tba
 Cabin Pre-heat/cool Control tba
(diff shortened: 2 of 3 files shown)


dexterbg:
Review results:
- You're removing setting the `ms_v_door_chargeport` state by this change, which will render the standard charge state notification useless, as that depends on the charge port state.

PR #1443, opened 2026-06-24
smartEQ: updated polling Voltage, sanity check pressure/temp

zorgms:


vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_poller.h
```diff
@@ -49,7 +49,7 @@ static const OvmsPoller::poll_pid_t obdii_79b_polls[] =
 static const OvmsPoller::poll_pid_t obdii_79b_cell_vrt_polls[] = 
 {
   { 0x79B, 0x7BB, VEHICLE_POLL_TYPE_OBDIIGROUP, 0x41, { 0,0,302,102 }, 0, ISOTP_STD },   // Battery Voltages Part 1 (Cells 1-48) 
-  { 0x79B, 0x7BB, VEHICLE_POLL_TYPE_OBDIIGROUP, 0x42, { 0,0,303,103 }, 0, ISOTP_STD },   // Battery Voltages Part 2 (Cells 49-96)  
+  { 0x79B, 0x7BB, VEHICLE_POLL_TYPE_OBDIIGROUP, 0x42, { 0,0,303,102 }, 0, ISOTP_STD },   // Battery Voltages Part 2 (Cells 49-96)
```
dexterbg: Shouldn't the cycles be identical to part 1 in all states? The driving state actually is even more sensitive to a difference, as loads change quickly.

zorgms: Yes, they should; it's just too warm—sorry.

PR #1444, opened 2026-06-29
smartEQ : Filter invalid telemetry spikes

zorgms:
Ignore implausible CAN voltage and speed readings before updating cached values, and stop the BMS poller from overwriting battery voltage with the polled pack value. Also add a cooldown-based reset for the 12V undervoltage alert so the event is logged once per incident and emits clearer trigger/clear log messages.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
@@ -118,8 +118,10 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
       if (raw_temp != 0x7F) 
         {
         can_bat_temp = _temp;
-        }      
-      can_bat_voltage = (float)((CAN_UINT(3) >> 5) & 0x3FF) / 2.0f;
+        }
+      float _volt = (float)((CAN_UINT(3) >> 5) & 0x3FF) / 2.0f;
+      can_bat_voltage = _volt < 450.0f ? _volt : 0.0f; // ignore invalid voltage reading > 450V
+      can_bat_voltage = 
       can_charge_climit = (c >> 20) & 0x3Fu;        
       break;
       }
@@ -130,8 +132,9 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
     case 0x5D7: // Speed, ODO
       {
       REQ_DLC(6);
-      // Apply scaling 
-      can_speed = (float)CAN_UINT(0) / 100.0f;
+      // Apply scaling
+      float _speed = (float)CAN_UINT(0) / 100.0f;
+      can_speed = _speed < 200.0f ? _speed : 0.0f; // ignore invalid speed reading > 200km/h  
       can_odometer = (float)(CAN_UINT32(2)>>4) / 100.0f;
       can_odometer_trip = (float)(CAN_UINT(4)>>4) / 100.0f;
       break;
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
@@ -113,15 +113,12 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
       {
       REQ_DLC(5);
       uint8_t raw_temp = (c >> 13) & 0x7Fu;
-      float _temp = (float)raw_temp - 40.0f;      
-      // Ignore invalid sensor reading (0x7F = 127 → 87°C after offset)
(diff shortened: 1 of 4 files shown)


vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
```diff
@@ -118,8 +118,10 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
       if (raw_temp != 0x7F) 
         {
         can_bat_temp = _temp;
-        }      
-      can_bat_voltage = (float)((CAN_UINT(3) >> 5) & 0x3FF) / 2.0f;
+        }
+      float _volt = (float)((CAN_UINT(3) >> 5) & 0x3FF) / 2.0f;
+      can_bat_voltage = _volt < 450.0f ? _volt : 0.0f; // ignore invalid voltage reading > 450V
+      can_bat_voltage = 
       can_charge_climit = (c >> 20) & 0x3Fu;
```
dexterbg: This cannot be correct.

Also, an invalid voltage isn't ignored, it's changed to 0V -- should probably better just keep the previous reading?

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
```diff
@@ -130,8 +132,9 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
     case 0x5D7: // Speed, ODO
       {
       REQ_DLC(6);
-      // Apply scaling 
-      can_speed = (float)CAN_UINT(0) / 100.0f;
+      // Apply scaling
+      float _speed = (float)CAN_UINT(0) / 100.0f;
+      can_speed = _speed < 200.0f ? _speed : 0.0f; // ignore invalid speed reading > 200km/h
```
dexterbg: Same here (keep previous value)?

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_ticker.cpp
```diff
@@ -58,13 +58,16 @@ void OvmsVehicleSmartEQ::Ticker1(uint32_t ticker)
...
-      m_12v_alerted = true;  
-      smart12VHistory();     // log undervoltage event in history metric and send data log
+      m_12v_alerted = true;
+      m_12v_alerted_ticker = 120; // alert reset cooldown for 120 seconds
+      ESP_LOGW(TAG, "12V undervoltage alert triggered:  %.2fV Module, %.2fV BMS, %.2fV USM (reference: %.2fV)", volt, bms12v, usm12v, m_ref12V);
+      smart12VHistory();         // log undervoltage event in history metric and send data log
       }
     }
-  else if (m_12v_alerted && (volt > 6.0f  && m_ref12V-volt <= m_alert12V))
+  else if (m_12v_alerted && m_12v_alerted_ticker > 0 && --m_12v_alerted_ticker == 0 && (volt > 6.0f && m_ref12V-volt <= m_alert12V))
```
dexterbg: This condition will never reset `m_12v_alerted` when the voltage isn't restored at the time of the check.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_ticker.cpp
```diff
@@ -58,13 +58,16 @@ void OvmsVehicleSmartEQ::Ticker1(uint32_t ticker)
       }
     else if (!m_12v_alerted) // alert only once when undervoltage is detected, to prevent log flooding
```
dexterbg: Side note: is this correct being an `else if`? I.e. a 12V undervoltage detected during non-off poll state or cooldown shall not be logged and not trigger an alert?

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
zorgms: changed to: Update only if the value is plausible

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
zorgms: changed to: Update only if the value is plausible

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_ticker.cpp
zorgms: logic adjusted

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_ticker.cpp
zorgms: The alarm can now be set immediately.

PR #1445, opened 2026-06-30
smartEQ: Fix 12V voltage readings and polling logic

zorgms:
Corrects 12V voltage calculations in the ticker:
- Changes USM voltage source from index 3 to 4 (CAN voltage)
- Renames variable from usm12v to voltcan for clarity
- Adds polling cooldown check to Ticker60 ADC check


zorgms:
Two EQs, two different results for the factor calculation, because the read-out voltages show varying deviations from the 12V battery. The VINs differ by only about 300 vehicles.

PR #1449, opened 2026-07-05
CT5: Set the acceptance filter to allow 0x063 (v.on) and 0x5E4 (v.b.soc)

leres:
The high frame rate of this vehicle (> 1600/sec) causes watchdog resets when the engine is running. Setting the acceptance filter means we don't see the VIN frames but at least we're not crashing over and over.
 
While here remove disabled/unused polling related code (given GM Global B architecture's security and authentication it is unlikely this vehicle will ever support polling).


leres:
The branch I was working with wasn't pristine, let me kill this PR and try again.

PR #1453, opened 2026-07-07
vwegolf: set v.e.on and v.e.awake

SCjona:
Problem: `ms_v_env_on` was never set, so the app always showed the car as parked.

Fix: Set `ms_v_env_on` based off READY (drive-able) status and `ms_v_env_awake` based on ignition status.

Also fix park time being limited to 36.5h

Test scenario => `ms_v_env_on`:
- Driving => true
- Ignition on, not drive-able => false
- Parked => false
- Charging (AC) => false
- Climate precondition on battery => false

Test scenario => `ms_v_env_awake`:
- Driving => true
- Ignition on, not drive-able => true
- Parked => false
- Charging (AC) => false
- Climate precondition on battery => false

Reference sources: https://github.com/commaai/opendbc/blob/master/opendbc/dbc/vw_mqb.dbc
Validated against CAN captures.

### DIFF
--- a/vehicle/OVMS.V3/changes.txt
+++ b/vehicle/OVMS.V3/changes.txt
@@ -8,6 +8,8 @@ Open Vehicle Monitor System v3 - Change log
     means we don't see the VIN frames but at least we're not crashing
     over and over.
 - Fiat eDoblo: added passive lock status from CAN 0x612.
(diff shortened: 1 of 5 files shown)


dexterbg:
Please resolve the conflicts.

PR #1458, opened 2026-07-12
Move BMS Monitor to a separate module from Vehicle + Add Duktape

frogonwheels:


frogonwheels:
This is still draft as there is an argument that the first commit should be separated out.

There are a couple of changes here that need flagging.
* Moving Battery Stats to a separate module is kind of big but does allow us to access it through the dbc engine without involving the current Vehicle!
* The  web-module enable/disable based on whether there are readings does make sense, especially with dbc engine stuff.  I have added RegisterPage() and DeregisterPage() methods onto MyBms that forces the pages to be at the current location in the registration.

dexterbg:
Michael, sorry for the long delay.

Following up on https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1406#issuecomment-5023764000 :

TL;DR: I don't see any advantage in factoring out the BMS part from the vehicle into a new singleton.

I still don't see the BMS subsystem as being special in any way compared to the other vehicle subsystems like the 12V or TPMS monitoring, or even the vehicle charge state monitoring. For all of these subsystems it eventually makes sense to provide a DBC path. Following your proposal, all of these are then to be seen as separate entities and will need to be factored out into singletons. That doesn't help in terms of clarity of the overall system structure, and complicates unloading a vehicle.

But other than the poller, none of these simple subsystems are operationally independent units, they are essentially just standardized helpers/utilities for the actual vehicle implementation.

Having the `OvmsVehicle` class provide these utilities and their public interface is clear, neat & clean. The only inconvenience of accessing these via the vehicle factory's `ActiveVehicle()` API is that you need to verify it's not null -- and that check can be easily encapsulated in some helper.

Do I miss something?

frogonwheels:
So whether or not these are sufficient or not is another matter, but the advantages/reasons are - 
 - The BMS monitor doesn't need the Vehicle implementation, it is a one-way dependency.  It sits between whatever is providing the BMS data and the Metrics.  It could even be a part of the Metrics subsystem.
 - This makes the Duktape implementation cleaner as it can sit there by itself. Again it might make more sense as a subsystem of the Metrics module than it does as a subsystem of the Vehicle Duktape module.
 - When I add the bms support to the poller / dbc it means not adding a dependency on the vehicle module for what is effectively a stand-alone operation for doing some stats to output to the metrics subsystem.

The thought about it being part of the Metrics susbsystem has only just occurred to me, and I would be happy to move it in there rather than create a new Singleton. It could be MyMetrics.BMS. Thoughts?

frogonwheels:
Also - 

- Allowing the Web-module to auto-enable means that anything that sets up BMS dimensions will automatically enable the web-module.
- The shell `bms` command is already at the top level, which effectively acknowledges where it conceptually sits (which is outside of vehicle).  It seems unnecessary to have a dependency on vehicle when it is not necessary.
- This might open up, to a degree, some alternative ideas on how we implement pure script vehicles.

dexterbg:
The BMS subsystem is coupled tightly to the vehicle, in terms of the API, the config and metrics namespace. The only exception is the command namespace, but that's passed by for convenience on other vehicle commands like charge & lock control as well, and doesn't imply a conceptual decoupling.

The integration in the vehicle class also provides a clean way to supply vehicle specific customized overrides. The current API definition allows these for the alert notification and the status command/formatters, override options for other methods can easily be added as needed.

> not adding a dependency on the vehicle module for what is effectively a stand-alone operation

It's not a stand-alone operation, it depends on the vehicle. Up to now mostly in terms of layout/formatting, but as batteries evolve, vehicles may need to customize even the value processing. There are already vehicles like the Bolloré Bluecar, combining two different battery systems (chemical + supercap based). Detaching the BMS subsystem from the vehicle implies needing to fix the provided semantics for all vehicles, which in turn means needing to support all sorts of battery variations in the future. I'd rather give vehicle implementations the option to customize the system as needed -- if some technology evolves into common ground, that can then be refactored into the base class.

Besides that: running a duktape vehicle still implies having a vehicle instance -- be it an extension of the current DBC vehicle type or some dedicated duktape vehicle type.

frogonwheels:
I got half way with another defence but the multiple battery systems
problem really was the clincher.

I'd like to keep the factoring (but place it into components/vehilce), but
keep the virtual functions to override display. Probably should bring in
the webmodule display to a function off of vehicle for a similar reason so
it can be overidden too (so we could have a separate graph per battery
group)..

I think if there are separate batteries with different chemistries then
they would be recorded separately in a new OvmsBMSMOnitor, adding separate
voltage/temperature vectors, and more importantly separate warning
thresholds etc and max/min recordings.

This would need the OvmsBmsMonitor to pass in a list of metrics it would
use to publish to, events to signal to, and parameters to read from.

The other way would be to record everything into a single list of metrics
but partition the lists into multiple sections for each battery group..
somehow we would need to partition:
   readings_per_module_v
  readings_per_module_t
  start_idx_v, end_idx_v
  start_idx_t, end_idx_t
  alerts_v
  alerts_t
  limit_tmax
limit_tmin
 limit_vmax
 limit_vmin

and possibly some other stuff.  It would probably still require having some
metrics recorded seperately.

How do you envisage this happening? I could do a simple factor to start
with and leave it to the next person? But I'd also like to have multiple
battery groups factored into the DukTape module and to the interface for
updating voltage/temperatures and module configuration from the DBC module.

//.ichael

dexterbg:
Moving the web UI into the vehicle module would be OK. As that needs to be explicitly registered by the vehicle anyway, being able to override the function in the vehicle class isn't actually necessary or beneficial, but probably is a cleaner source structuring.

Any changes about the internal structure and especially the model should be avoided for now. That should be reconsidered once we actually have that case with all technical details we can currently only speculate on. It may well be going in that direction of simply having a pack index in addition to the cell & module indexes, but it may also need a totally different battery concept. If someone gets a redox flow battery up to the point of being usable for vehicles, we may need to have a model of fluids, ion concentrations and functional components like pumps instead of cells, temperatures and voltages.

PR #1459, opened 2026-07-12
fix(can): auto-recover wedged mcp2515 RX path via hardware-confirmed stall watchdog

slothish:
Caught this live on my e-Golf: can3 (mcp2515) RX froze mid-drive — Rx pkt and interrupt counters stuck, CANINTF "RX buffer 0 full" latched, Rx err/ovrflw all zero. The existing inactivity watchdog never fires in this state since it only runs while the vehicle is on, so the bus needed a manual `can can3 stop`/`start` to come back. Happened twice in one day under normal KCAN load (~1.3k fps).

This adds a hardware-confirmed RX-stall watchdog to BusTicker10:

- fires only when the bus is started (Active/Listen), `packets_rx` has been frozen for `CAN_RXSTALL_THRESHOLD` (20s), AND the driver confirms unserviced RX pending via a new virtual `CheckRxStalled()` — mcp2515 implements it as a live CANINTF read (RX0IF|RX1IF). Base class returns false so esp32can is untouched.
- pure RX silence never triggers it — a sleeping bus (car off, DC charging) has no CANINTF RX bits set, so no false resets. The SPI readback only happens in the slow path after 20s of frozen counter, nothing added to the hot path.
- recovery is `Reset()`, counted in a new `rxstall_resets` status field ("Wedge Rsts" in `can status`), one ESP_LOGE per event.

On-car validation: with this firmware the watchdog caught and auto-recovered 2 wedges in one drive session (Wedge Rsts: 2), which also pinned the root cause — RX pending in CANINTF with zero ISR entries, i.e. the INT edge got lost. I'm submitting the root-cause fix (level-triggered INT) as a separate PR on top of this one; this watchdog stays useful as the regression detector for it.

Heads up per my usual disclosure: developed with AI assistance, validated against the real car (counters and logs above).


slothish:
closing — accidentally based on my integration branch, will refile clean off master

PR #1460, opened 2026-07-12
fix(mcp2515): convert INT to level trigger, close RX-wedge root cause

slothish:
Root-cause fix for the RX wedge that the stall watchdog PR (#1459) detects and recovers from. Builds on that branch.

The MCP2515 INT pin is level-asserted — it stays low as long as any unmasked CANINTF bit is set, it doesn't pulse. The driver registers it as `GPIO_INTR_NEGEDGE`, so the ESP32 only reacts to the high→low transition. If a single service gets lost (concretely: `xQueueSendFromISR` onto a momentarily full rx queue, return value was never checked), RX0IF stays set, INT stays low forever, and no new edge can ever fire: RX is permanently dead with frozen counters. That's exactly the signature I hit twice on-car, and the watchdog confirmed it — CANINTF had RX pending while the ISR count stood still.

Fix, using the standard level-trigger handshake:

- INT GPIO switched to `GPIO_INTR_LOW_LEVEL`
- ISR masks its own pin before returning (direct `GPIO.pin[n].int_ena = 0` write — the `gpio_intr_disable()` call isn't IRAM-safe) and queues the rx task as before, now with the send result checked (failure counted in a new `isr_queue_overrun` status field)
- the rx task's drain loop re-arms the pin via `gpio_intr_enable()` on every drained exit — level semantics mean a frame arriving between the final CANINTF read and re-enable just re-fires immediately, no lost-event window
- `Start()` unconditionally re-arms as a safety net, so any Reset cycle restores a known-good state
- INT pins are per-instance on this board (checked ovms_peripherals.cpp), no shared-line complications; swcan inherits the same pattern

On-car validation: 7M+ frames on can3 at ~1.3k fps with `Wedge Rsts: 0` and `Isr Ovrflw: 0` — the same drive profile that wedged twice on edge-triggered firmware the day before.

Same disclosure as the watchdog PR: AI-assisted development, validated on the real car.


slothish:
same as #1459 — refiling clean off master

PR #1461, opened 2026-07-12
fix(can): auto-recover wedged mcp2515 RX path via hardware-confirmed stall watchdog

slothish:
Caught this live on my e-Golf: can3 (mcp2515) RX froze mid-drive — Rx pkt and interrupt counters stuck, CANINTF "RX buffer 0 full" latched, Rx err/ovrflw all zero. The existing inactivity watchdog never fires in this state since it only runs while the vehicle is on, so the bus needed a manual `can can3 stop`/`start` to come back. Happened twice in one day under normal KCAN load (~1.3k fps).

This adds a hardware-confirmed RX-stall watchdog to BusTicker10:

- fires only when the bus is started (Active/Listen), `packets_rx` has been frozen for `CAN_RXSTALL_THRESHOLD` (20s), AND the driver confirms unserviced RX pending via a new virtual `CheckRxStalled()` — mcp2515 implements it as a live CANINTF read (RX0IF|RX1IF). Base class returns false so esp32can is untouched.
- pure RX silence never triggers it — a sleeping bus (car off, DC charging) has no CANINTF RX bits set, so no false resets. The SPI readback only happens in the slow path after 20s of frozen counter, nothing added to the hot path.
- recovery is `Reset()`, counted in a new `rxstall_resets` status field ("Wedge Rsts" in `can status`), one ESP_LOGE per event.

On-car validation: with this firmware the watchdog caught and auto-recovered 2 wedges in one drive session (Wedge Rsts: 2), which also pinned the root cause — RX pending in CANINTF with zero ISR entries, i.e. the INT edge got lost. I'm submitting the root-cause fix (level-triggered INT) as a separate PR on top of this one; this watchdog stays useful as the regression detector for it.

Heads up per my usual disclosure: developed with AI assistance, validated against the real car (counters and logs above).

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/can.cpp
+++ b/vehicle/OVMS.V3/components/can/src/can.cpp
@@ -388,6 +388,7 @@ void can_status(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, c
     writer->printf("Wdg Timer: %20" PRId32 " sec(s)\n",monotonictime-sbus->m_watchdog_timer);
     }
   writer->printf("Err Resets:%20d\n",sbus->m_status.error_resets);
+  writer->printf("Wedge Rsts:%20d\n",sbus->m_status.rxstall_resets);
   }
 
 void can_explain_flags(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, const char* const* argv)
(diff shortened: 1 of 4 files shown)


dexterbg:
Review results:
- Is the RX stall reset count actually worth a separate status counter (instead of using `watchdog_resets` for this as well)? From the user perspective, I'd say no. A user primarily needs to know a watchdog condition occured and the bus needed to be reset = frames were lost. If I want to know details about what specific type of watchdog condition was detected, I can fetch them from the log. If you still think this condition needs to be counted separately, I won't object, but you then need to address the next two issues:
- **If** it is worth a separate counter, it should also be worth being logged. It currently isn't, but `canbus::StatusChanged()` still signals a change on the field, which leads to CAN loggers emitting identical status messages. Logging the field would need to be added to `canbus::LogStatus()` and `canformat_crtd::get()`.
- **If** it is worth a separate counter, the status command output shouldn't output it as "Wedge Rsts", as "Wedge" is too close to "Wdg", potentially confusing users. Something along "Rx stall resets" would be ok.

slothish:
Agreed — folded it into `watchdog_resets`. Dropped the separate field, its `can_status` line and the checksum term; the RX-stall watchdog now bumps `watchdog_resets` like the other reset paths. Detection logic is unchanged. Pushed.

PR #1462, opened 2026-07-12
fix(mcp2515): convert INT to level trigger, close RX-wedge root cause

slothish:
Root-cause fix for the RX wedge that the stall watchdog PR (#1461) detects and recovers from. Builds on that branch.

The MCP2515 INT pin is level-asserted — it stays low as long as any unmasked CANINTF bit is set, it doesn't pulse. The driver registers it as `GPIO_INTR_NEGEDGE`, so the ESP32 only reacts to the high→low transition. If a single service gets lost (concretely: `xQueueSendFromISR` onto a momentarily full rx queue, return value was never checked), RX0IF stays set, INT stays low forever, and no new edge can ever fire: RX is permanently dead with frozen counters. That's exactly the signature I hit twice on-car, and the watchdog confirmed it — CANINTF had RX pending while the ISR count stood still.

Fix, using the standard level-trigger handshake:

- INT GPIO switched to `GPIO_INTR_LOW_LEVEL`
- ISR masks its own pin before returning (direct `GPIO.pin[n].int_ena = 0` write — the `gpio_intr_disable()` call isn't IRAM-safe) and queues the rx task as before, now with the send result checked (failure counted in a new `isr_queue_overrun` status field)
- the rx task's drain loop re-arms the pin via `gpio_intr_enable()` on every drained exit — level semantics mean a frame arriving between the final CANINTF read and re-enable just re-fires immediately, no lost-event window
- `Start()` unconditionally re-arms as a safety net, so any Reset cycle restores a known-good state
- INT pins are per-instance on this board (checked ovms_peripherals.cpp), no shared-line complications; swcan inherits the same pattern

On-car validation: 7M+ frames on can3 at ~1.3k fps with `Wedge Rsts: 0` and `Isr Ovrflw: 0` — the same drive profile that wedged twice on edge-triggered firmware the day before.

Same disclosure as the watchdog PR: AI-assisted development, validated on the real car.

### DIFF
--- a/vehicle/OVMS.V3/components/can/src/can.cpp
+++ b/vehicle/OVMS.V3/components/can/src/can.cpp
@@ -364,6 +364,7 @@ void can_status(int verbosity, OvmsWriter* writer, OvmsCommand* cmd, int argc, c
   writer->printf("DBC:       %s\n",(sbus->GetDBC())?sbus->GetDBC()->GetName().c_str():"none");
 
   writer->printf("\nInterrupts:%20" PRId32 "\n",sbus->m_status.interrupts);
+  writer->printf("Isr Ovrflw:%20d\n",sbus->m_status.isr_queue_overrun);
   writer->printf("Rx pkt:    %20" PRId32 "\n",sbus->m_status.packets_rx);
   writer->printf("Rx ovrflw: %20d\n",sbus->m_status.rxbuf_overflow);
   writer->printf("Tx pkt:    %20" PRId32 "\n",sbus->m_status.packets_tx);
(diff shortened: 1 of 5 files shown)


dexterbg:
This PR includes all changes of PR #1461, may need a separate branch to avoid conflict… but maybe Github meanwhile can sort this out.

On the changes:
- `isr_queue_overrun` in this case is worth a separate counter, and this means it also should be included in the status logs (as explained in PR #1461). For this, `canbus::StatusChanged()` needs to take this into the checksum, which is missing here. Also, `can_status()` should show this right below the interrupt count, not at the end.
- This is probably wrong: "Direct register write, not gpio_intr_disable(): that call is not IRAM-safe and must not be invoked from an ISR (it may run with the flash cache disabled)". I don't think this applies to IDF 3.3, only to later IDF versions. A direct `GPIO` access is probably still OK, but the comment is at least misleading…
- …and I do find `gpio_intr_disable()` also clears the interrupt status for the pin, which is missing from your code without explanation. So please also check and explain why this has been / can be omitted.

slothish:
Pushed fixes for all three.

**isr_queue_overrun** — kept the counter, and wired it up properly: added to the `StatusChanged()` checksum, added to both status log sinks (`LogStatus` and `canformat_crtd::get`, as `isrovr`), and moved its `can_status` output up to sit right under the interrupt count. Also folded `rxstall_resets` into `watchdog_resets` here to match #1461 (this branch still stacks it for now).

**The IRAM comment** — I dug into this and I think it does hold for 3.3, it just wasn't explained. OVMS installs the GPIO ISR service with `ESP_INTR_FLAG_IRAM` (`main/ovms_peripherals.cpp`), so the per-pin handler runs in an IRAM context and can fire while the flash cache is disabled. `gpio_intr_disable()` in `components/driver/gpio.c` isn't `IRAM_ATTR` in v3.3.4 (only `gpio_intr_service` is), so calling it from the ISR isn't safe here. It's the install flag that decides it, not the IDF version — I rewrote the comment to say that explicitly and cite the flag.

**The status clear** — good catch, you're right that `gpio_intr_disable()` also does `gpio_intr_status_clr()`. I skip it on purpose: every re-arm path goes through `gpio_intr_enable()`, which clears the latch *before* it sets `int_ena` (also in gpio.c), so the latch is cleared exactly at re-enable, which is the only moment it matters. Clearing it at mask time would be redundant, and moot anyway while INT is still held low since the level condition just re-latches it immediately. Added a comment spelling that out.

dexterbg:
No, Github still cannot sort out stacked PRs, so please resolve. For the future: always create separate branches for each functional group of changes.

PR #1479, opened 2026-07-25
refactor(vwegolf): split KCAN TX/actuation path into vwegolf_tx.cpp

slothish:
Pure code move, no functional change. Discussed on ovmsdev — Michael's answer was that restructuring is fine as long as it lands as its own PR without functional changes, so that's what this is.

**Why:** two forks are actively working this component now — mine (climate/BAP, charging next) and @SCjona's (the ignition-status work that merged as #1453). We're both editing `vehicle_vwegolf.cpp`, and recent master syncs have cost me batches of mechanical moved-function-body conflicts. That only gets worse as the file grows, and charging control lands entirely on the TX side.

**What moves:** the trailing TX/actuation block — `CommandLock`, `CommandUnlock`, `CommandMirrorFoldIn`, `CommandHorn`, `CommandIndicators`, `CommandPanic`, `WakeKcanBus`, `CommandWakeup`, `SendOcuHeartbeat`, `SendNmAlive`, `SendClimaBapBurst`, `CommandClimateControl` → new `src/vwegolf_tx.cpp`. Decode/RX (`IncomingFrameCan2`, `IncomingFrameCan3`, `Ticker1`) stays in `vehicle_vwegolf.cpp`.

Function bodies are byte-identical to master. No reordering, no reformatting, no clang-format pass — the diff should read as a move and nothing else. Beyond the move itself: the new file carries the same license header, include, and `TAG` define as its parent; `CMakeLists.txt` and the native test Makefile get the new source; and one comment in `test_clima.cpp` that pointed at the old home of `SendClimaBapBurst` now points at the new one. `component.mk` needs nothing, it globs `src/*.cpp`.

Native tests pass. Concurrency fixes I have stacked on top of this in my branch will come as separate small PRs so this one stays trivial to review or revert.

@SCjona — you're the other one paying the conflict tax here, so I'd rather have your blessing than surprise you. Does the RX/TX seam work for where your changes land? Happy to hold this until you're at a quiet point so we both rebase once instead of twice.

Usual disclosure: like the rest of my e-Golf work this is heavily AI-assisted, validated against the native tests, compiler feedback and real CAN captures. The move itself is verified behavior-preserving by the component's test suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SCjona:
not sure that will help a whole lot. maybe splitting by feature is better. `vehicle_vwegolf_climate.cpp` / `vehicle_vwegolf_commands.cpp` (other simpler commands) / `vehicle_vwegolf_charge.cpp`. keep `vehicle_vwegolf.cpp` for RX (metrics & others with forwarding calls to other files if relevant) and tickers (also forward call these to other feature files if it makes sense) and whatever else doesn't fit into the other files

I've been working on climate control v2. full BAP library, error handling & feedback (notification on failure), completely getting rid of OCU wakeup/heartbeats for climate (as this sets DTCs in the original OCU) and instead using unassigned ECU ids to trigger the wakeup. haven't pushed it to my fork yet, but works fine so far. will file a PR for that soon

slothish:
Actually I like the sound of that, I'm gonna close this PR and take a look at your code

SCjona:
here's the climate v2 code, but i haven't reviewed/validated it in depth, this is mostly just Claude:
https://github.com/SCjona/Open-Vehicle-Monitoring-System-3/commit/aec4e59f625baa0c4939a1c0bbcc5b63e421e62e
https://github.com/SCjona/Open-Vehicle-Monitoring-System-3/commit/62669373606fd87c2bdf5baa921f8e6fa8689b8e
it's also not synced with OVMS upstream yet

slothish:
I don't have my module live right now, I had to gift my cable to a friend to get his module up and running, but I'll check it against the recordings I already have and later next week when i have the parts for a new cable Ill check it on the car as well.

slothish:
Went through both commits. The node 0x67 thing checks out — ran it against my own captures and you're right.

`0x1B000067`: 6369 frames across 20 of my 41 captures, sustained cadence. The payload my firmware sends (`67 10 41 84 14 00 00 00`) appears zero times in 73264 NM frames. And the payloads that *are* there vary within a session — `67 10 04 01 01`, `67 10 08 01 01`, some with byte1=0x00 which my code never emits. It's in captures where OVMS transmitted nothing at all. So yeah, live ECU, and I've been stepping on it.

Haven't confirmed the DTCs with a scan tool, that part's still yours.

I'm switching to 0x7D as well so we don't end up with two different ids upstream. On the payload though — `49 85` isn't in my corpus. Every byte2 I see sets exactly one low-nibble PNC bit, optionally |0x40: `01 02 04 08 41 42 44 48`. `0x49` sets two. I went with `7D 10 08 01 14 00 00 00`: byte2/byte3 lifted from node 0x46, byte1/byte4 carried over from my old frame. Being straight about it — that exact composite isn't observed either, node 0x46 doesn't use `10`/`14`. It's a shape-plausible guess, not a confirmed one. Wake test on a sleeping bus is the only thing that settles it.

Two things your BAP lib gave me for free, both closing open questions in my RE notes:

- `49 58 <flag>` is a real climate-state echo. Correlated it against `0x3B5` ClimaRunning bit7 — 9/9 transitions across 5 captures, stop tracks within 70ms, start lags ~3s. My docs had func 0x18 down as trigger-only. Wrong.
- Your ProfilesArray STATUS layout parses my reassembled `49 59` bodies cleanly — total/flags/startIndex/count then 20-byte records. Names come out exact, which is a good sign the record framing is right: profile 0 (the hidden global one, eManager's base clima settings) is "Optionen", second record "100% 10A w AC" at nameLen 0x0d. Byte 12 decodes as a plausible temperature under `raw = °C×10−100` but I'm not claiming it — I don't have a capture where the eManager setting was logged beforehand, so I can't check the decode against anything. Getting that on the next probe run.

One concern with dropping the ProfilesArray write. You describe it as only setting maxCurrent, but byte 4 of that record is `operation`, and 0x06 is climate + climateWithoutExternalSupply. On a car whose stored profile doesn't have bit 2 set, unplugged, `29 58 00 01` on its own might not start anything — and the failure surfaces as your no-response timeout. Works on our cars because the stored profile already has 0x06.

Easy to settle: set the profile to climate-plugged-only in the infotainment, unplug, send the OperationMode trigger with no ProfilesArray write. If it starts, I'm wrong.

One naming thing — you've got byte 0 as `arrayId` with `0x29` = full list / `0x2A` = changed. I think that's the ASG-ID/TID pair, not an enum. I see it walk `0x28..0x2f` and wrap, and it comes back `0x82/0x83` on write-ACKs, which is the `|0x80` echo (16/16 write bursts in my corpus). The profile selector is the index field, not byte 0 — and there profile 0 really is the base settings, matching what the eManager shows. Same trap bit me on `0x22/0x23`, which I had documented as a start/stop selector until a probe showed direction comes from frame 3 alone.

Last thing — your diff moves climate off the collision but `CommandWakeup()` and the 0x5A7 heartbeat still impersonate 0x67, so the wake button and lock/mirror/horn are still on the old path.

And agreed on splitting by feature. I'll re-derive along those lines.

SCjona:
> One concern with dropping the ProfilesArray write.

You're probably correct in your assumption that it won't start climate if unplugged if the feature is turned off in infotainment, but I decided to drop that anyways because I didn't want to write persistent data. It is technically now possible to write extended profile data (target temperature, max climate duration), although it's kinda lacking the user interface, so I did not implement it.

Worth a note in the docs though. Also you'd likely get error feedback if you leave it off (climate ECU refusing your request)

SCjona:
> Last thing — your diff moves climate off the collision but CommandWakeup() and the 0x5A7 heartbeat still impersonate 0x67, so the wake button and lock/mirror/horn are still on the old path.

out of scope for my climate work.

to be tested - try 0x5A7 heartbeat but with the custon NM wake node ID triggering the wakeup and see if that sets an DTC. might be possible to also drop the 0x5A7 heartbeat too (and just send NM wakeups instead) except when you need to issue commands (lock/fold mirror/etc).

PR #1486, opened 2026-08-08
Fix : MCP2515 initialisation now respects filters and rollover to RXB1

### LINKED ISSUE #1485: MCP2515 setaccfilter filters being ignored
Testing on ZE1 leaf, trying to restrict traffic on can2 and only receive 0x385 packets.  Using the command...

`can can2 setaccfilter FFE00000 70A00000 70A00000 FFE00000 70A00000 70A00000 70A00000 70A00000`

Followed by 

`can log start vfs crts /sd/cap.crtd`

then

`can log stop`

On investigation there are many packets in capture from can2 that have id's other than 0x385.  It seems that the can filters are being ignored.

wayne-love:
Fixes #1485.

MCP2515 was being initialised to accept all (ignore filters) and rollover was disabled (contrary to the code comments).  I'm not a MCP2515 expert but the changes seem sound and test ok on my Leaf.


dexterbg:
You're probably right regarding the acceptance filter mode needing activation in `REG_RXB0CTRL`. That seems to be missing.
But that must only be done when an acceptance filter actually is configured, not as a default.

Your rollover change doesn't make any sense. The rollover is already enabled (as commented) and you try to set a read-only copy of the control bit.

PR #1494, opened 2026-08-12
SmartEQ: Set CAN transceiver to listen-only mode and add option CAR-Off

zbchristian:
In case the CAN write acces is disabled by user selection, the CAN transceiver is set into listen-only mode via the corresponding GPIO. The listen-only mode ensures, that a mistake in the firmware or a crash of the module is not interfering with the car.
This had already been present for the SmartEQ, but somhow got lost again. 

In addition the different CAN write access selections (`SmartEQ->Features`) has been replaced by a radio button to have a mutual exclusive selection.

Added again the `CAR OFF` selection, which prohibits any interference of the module with the car, while driving. The SmartEQ is sending a lot of information over the bus anyway and this is available in listen-only mode as well. At least for my application, there is no need for polling while driving. The `CAR OFF` CAN write access allows polling, while the car is parked and/or charging.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
@@ -46,7 +46,7 @@ static const char *TAG = "v-smarteq";
 void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
   uint8_t *data = p_frame->data.u8;
   uint64_t c = swap_uint64(p_frame->data.u64);
-  
+  m_can_msg_received = true;
   switch (p_frame->MsgID) {
     case 0x17E: //gear shift
       {
@@ -223,6 +223,15 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
       can_charge_inprogress = (CAN_BYTE(5) & 0x20) != 0;
       break;
       }
+    case 0x668:
+      // Car on/off state used if CAN write disabled, otherwise utilize 0x350 polling
+      if (!IsCANwrite() )
+        {
+        REQ_DLC(1);
+        can_env_on = (CAN_BYTE(0) & 0x40) > 0;
+        }
+      break;
+
     case 0x673:
       {
       // TPMS pressure values only used, when CAN write is disabled, otherwise utilize PollReply_TPMS_InputCapt
@@ -249,6 +258,7 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
       break;
       }
     default:
+      m_can_msg_received = false;
       //ESP_LOGI(TAG, "PID:%x DATA: %02x %02x %02x %02x %02x %02x %02x %02x", p_frame->MsgID, data[0], data[1], data[2], data[3], data[4], data[5], data[6], data[7]);
       break;
   }
@@ -257,6 +267,7 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
 // Sync CAN datapoints to OVMS metrics, called by Ticker1, because CAN refresh rate is too high
 void OvmsVehicleSmartEQ::smartCAN2Metrics()
 {
+  m_can_msg_received = false; 
   if (m_cmd_locked && !can_locked)
     {
     // prevent desync of command lock and actual lock status    
@@ -267,6 +278,7 @@ void OvmsVehicleSmartEQ::smartCAN2Metrics()
     StdMetrics.ms_v_env_locked->SetValue(can_locked);
     }
   StdMetrics.ms_v_env_on->SetValue(can_env_on);
+  CheckCANAccess();  // car on/off state might have changed - check, if CAN write access has to be changed 
   StdMetrics.ms_v_env_awake->SetValue(IsAwakeEQ());
   StdMetrics.ms_v_env_hvac->SetValue(can_hvac);
   StdMetrics.ms_v_env_handbrake->SetValue(can_handbrake);
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_web.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_web.cpp
@@ -82,14 +82,15 @@ void OvmsVehicleSmartEQ::WebCfgFeatures(PageEntry_t& p, PageContext_t& c)
   auto lock = MyConfig.Lock();
 
   std::string error, full_km, rebootnw, contactor_1h_limit;
-  bool canwrite, canwrite_caron, canwrite_sleep, led, resettrip, resettotal, bcvalue;
+  bool canwrite, canwrite_caron, canwrite_caroff, canwrite_sleep, led, resettrip, resettotal, bcvalue;
   bool charge12v, extstats, unlocked, tripnotify, opendoors;
   bool obdii79b, obdii79b_cell, obdii743, obdii745, obdii745_tpms, obdii7e4, obdii7e4_dcdc;
 
   if (c.method == "POST") {
     // process form submission:
-    canwrite    = (c.getvar("canwrite") == "yes");
-    canwrite_caron = (c.getvar("canwrite_caron") == "yes");
+    canwrite    = (c.getvar("canwrite_mode") == "ON");
+    canwrite_caron = (c.getvar("canwrite_mode") == "CARON");
+    canwrite_caroff = (c.getvar("canwrite_mode") == "CAROFF");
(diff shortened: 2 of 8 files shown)


dexterbg:
I need to ask if this new config option is worth risking someone's Smart EQ to be bricked.

Keeping the transceiver stable in active state was a deliberate choice, as part of the measures against the BMS glitch issue #1405. It's still unclear if this is relevant, but the listen-only mode may cause issues on the bus, if the module hasn't been altered to enable the termination resistors. At least one case has been reported where the listen mode even causes issues with termination enabled (that was on the Nissan Leaf, see issue #980).

Did you discuss this change with @zorgms ?

In any way, any change to the basic CAN communication on the Smart EQ now needs to take testing for the glitch trigger into account, i.e. monitoring of server records for new cycle counter changes. So you need to ask for this support explicitly, and plan on checking raw data delivered e.g. by me to you for signs of the issue for some weeks, and be prepared to change the code as quick as possible if necessary.

zbchristian:
I did not really follow the discussion wrt the Smart EQ problem, but I seriously doubt, that the listen-only mode can cause any glitch on the CAN bus in this case. For the ESP32CAN the active state is actually the risky state!

This has nothing to do with the bus termination, which is done on the car side of the transceiver, so nothing changes when switching between active and listen-only mode wrt the termination. This is hardware-wise impossible!

As already discussed in #980, the above statement is at least true for CAN1 (ESP32CAN) and CAN2 (MCP2515), but maybe not for the 3rd CAN bus attached to the MAX7317. So I think, that the issue seen for the Nissan Leaf does not apply at all to the Smart EQ with a single CAN bus. 

With respect to the correct termination, the situation is much more difficult and whether the module should have the terminations enabled or not, depends on the car. One could imagine problems for misconfigured cases, where the communication is anyway on the edge:
- OBD connector w/o a termination resistor and the module has the termination disabled. This can lead to an unreliable communication on the CAN bus (depending on the cable length).
- OBD connector includes the termination resistor. In this case, the module has to have the termination disabled, otherwise again, the communication will be unreliable. With a long cable to the module, this is a difficult case, since reflections on the cable can cause problems. For an existing OBD side termination, only a short cable should be used.

To be clear about this, there is no standard for the termination of the OBD connector. Some cars have a termination built in and other do not. That is actually the reason, why some cheapy OBD adapters work on some cars, but not on others.  

There is one edge case, where a glitch could occur during the state transition of the transceiver. If the module is still active on the CAN bus, while the GPIO pin is changing its state, a glitch could occur. This would mean, that the modules CAN bus activity has not been suspended before asking for the state change. But that means, that there is a software problem.

Just to be on the safe side, I would propose to just use a (temporary) flag to build the firmware with or without the ability  to set the transceiver to the listen-only mode. 
@dexterbg and @zorgms, would that be an option for the time being?

Just to make the point again: 
After my experience with a completely unusable SmartEQ, even the power steering failed, after a firmware crash, I will no longer allow the module to be active on the CAN bus, while driving! The only means to ensure this, is the listen-only mode.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
```diff
@@ -223,6 +223,15 @@ void OvmsVehicleSmartEQ::IncomingFrameCan1(CAN_frame_t* p_frame) {
...
       }
+    case 0x668:
+      // Car on/off state used if CAN write disabled, otherwise utilize 0x350 polling
+      if (!IsCANwrite() )
+        {
+        REQ_DLC(1);
+        can_env_on = (CAN_BYTE(0) & 0x40) > 0;
+        }
+      break;
+
```
zorgms, line 234: Why populate a variable from two sources (0x668)? 0x350 clearly indicates the status.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_web.cpp
```diff
@@ -209,17 +212,23 @@ void OvmsVehicleSmartEQ::WebCfgFeatures(PageEntry_t& p, PageContext_t& c)
...
+  else if ( canwrite_caroff ) canwrite_mode = "CAROFF";
+  c.input_radiobtn_start("CAN Bus Write Access", "canwrite_mode");
+    c.input_radiobtn_option("canwrite_mode", "Off", "OFF", canwrite_mode == "OFF");
+    c.input_radiobtn_option("canwrite_mode", "On", "ON", canwrite_mode == "ON");
+    c.input_radiobtn_option("canwrite_mode", "Car Off", "CAROFF", canwrite_mode == "CAROFF");
+    c.input_radiobtn_option("canwrite_mode", "Car On", "CARON", canwrite_mode == "CARON");
+  c.input_radiobtn_end(
+    "<p>Enable write access to CAN bus.<br> Off - no access (Listen only)<br>On - allow full access<br> Car Off  - only when car is off (parked)<br> Car On - Only when car is on</p>");
+
+    c.input_checkbox("Disable CAN polling during sleep", "canwrite_sleep", canwrite_sleep,
```
zorgms, line 228: I find the description of the modes a bit confusing.

zorgms:
I understand your intention, but you can also manage your idea via the poll handler. It is true that the bus remains active the whole time, but the poll list is modified depending on the state. Could you consider simply clearing the poll list when the car is on?
E.g., in Ticker1:
```
if(m_can_active && IsOnEQ() && m_enable_write_caroff)
{
smartCoolDownPolling(); 
smartOBDpolling(false); 
}
```
That would be just a minimal change.

zorgms:
Switching between CAN LISTEN and ACTIVE modes could also be handled within the poll handler; this would certainly be preferable, as clearing the poll list—combined with the cooldown—might even be slightly safer.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
zbchristian: This depends, if the 0x350 is send by the Smart on its own, or is just available by polling. in the latter case, there would be no information about the cars state in listen-only mode.

You might know, if the car is actively sending a 0x350 message or not.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_web.cpp
zbchristian: Ok, this could be clarified to make more clear, what the different states mean. Will make a proposal.

zbchristian:
The vulnerability of the cars CAN bus originates from two sources:
- software problems, like very high loads on the CAN bus
- Module reboot/abort. If the transceiver is in active mode, this has a high chance to trigger an error on the CAN bus. This is especially true for CAN1 (ESP32CAN). In cases, where the module is rebooting frequently, this is a serious safety concern. For CAN1, one could even imagine a case, where the corresponding GPIO pin is for a longer time at a state, blocking the CAN bus completely. For the MC2515 based CAN buses, the risk is much smaller, because the ESP32 is not directly connected to the CAN bus.

Just to clear the polling list and rely on the firmware to be ok, is insufficient and has practically no effect on the safety of the whole system.

You probably should have a look at my study to measure the interference of the module with the CAN bus at https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/issues/980#issuecomment-2970786192
  
Since the study shows, that CAN1 is inherently sensitive to crashes and reboots, I modified my OBD cable to the module and use CAN2 instead. This already buys a lot of safety margin.

vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can.cpp
zorgms: 0x350 is actively transmitted, comparable to 0x668.

zorgms:
I have no concerns about switching between active and listening modes. I just wanted to point out that this can be integrated into the poll handler.

https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/51097a2fc6cfd8f5614b363352d86ace3aa3a9a1/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_handle.cpp#L101

```
if (!m_can_active)
    {
    ESP_LOGD(TAG, "HandleOBDpolling(): OBD polling disabled (CAN bus not active)");
    RegisterCanBus(1, CAN_MODE_LISTEN, CAN_SPEED_500KBPS);
    m_can_last_acc_state = false;
    return;
    }

if (!m_can_last_acc_state)
   {
   RegisterCanBus(1, CAN_MODE_ACTIVE, CAN_SPEED_500KBPS);
    m_can_last_acc_state = true;
   }
```

zorgms:
Active polling is controlled via "smartOBDpolling(true/false);", so I think it makes sense for the switching to be controlled that way as well.

zbchristian:
New PR #1509 superseeds this PR

PR #1496, opened 2026-08-12
vwegolf: climate control v2 + charge control

SCjona:
# VW e-Golf: remote climate + charge control over BAP (self-contained BatteryControl module)

## Summary

This supersedes the merged KCAN-BAP remote-climate implementation (#1430) with an
on-car-validated one, moves it out of `vehicle_vwegolf.cpp` into a dedicated
`vehicle_vwegolf_bat_ctrl.{h,cpp}` module plus a small reusable BAP library
(`src/bap/`, `src/egolf/`), and **adds remote charge control** (start/stop and
charge-current limit) over the same channel.

From the user's side, remote pre-heat/cool is still the app / standard
`climatecontrol` command; charge control adds the standard `charge start`,
`charge stop`, and `charge current <amps>` commands. What changed underneath is
how OVMS wakes the car, arms the command, and confirms it — which fixes several
problems in the #1430 version and makes charge control possible.

> **Note for maintainers:** #1430 is already on `master`. This replaces its
> `changes.txt` entry rather than appending a second one. If you'd prefer a
> different shape (e.g. revert-then-add), happy to restructure.

### Why climate and charge in one PR?

They aren't really separable — charge control is why the climate path was
reworked, not a bolt-on:

- **Shared code.** Both drive the *same* BatteryControl channel (LSG `0x25`)
  through one module and one command path: spare-node wake → handshake → read
  profile 0 → read-modify-write arm → OperationMode trigger → confirm. They differ
  only in the operation byte the arm writes. Splitting them would either duplicate
  that whole path or land it half-wired with dead code.
- **Interdependent behaviour.** The immediate trigger acts on a *single* shared
  "immediate" profile (profile 0), so the two operations have to agree on who owns
  which op bits, and several rules only exist because both are present:
  climate force-sets the on-battery flag *because* charge must clear it; a charge
  stop is refused while climate is running; stop is operation-specific for the same
  reason. None of those rules — or their tests — mean anything with only one half
  in the tree.

Landing them together keeps the module, its invariants, and its test suite
coherent in a single reviewable unit.

## Why change #1430?

The previous climate implementation had issues that showed up on the car:

1. **It impersonated the factory OCU** to wake/hold the comfort bus — sending both
   the OCU's node-`0x67` CAN-NM frame (`0x1B000067`) and its `0x5A7` heartbeat.
   This car has a live OCU transmitting both, so OVMS's copies collide (same id,
   different data) and left OCU faults (U001100 / U120100) in the OCU's memory.
2. **It fired the trigger once, optimistically**, setting `v.e.hvac` immediately
   without waiting for the car to accept — so a declined request (e.g. unplugged)
   or a bus hiccup reported a success that never happened.
3. **The immediate trigger acts on the car's global "immediate" profile**, so it
   is only deterministic if that profile is armed to the intended operation first.
   Getting that arm right (and not clobbering the owner's other settings) needs a
   read-modify-write with correct sequencing — see below — which also unlocks
   charge control on the same mechanism.

## What this PR does

- **Wakes the comfort/EV cluster with a CAN-NM message from a spare, unused node
  id** (`0x1B00007D`) instead of impersonating the OCU — no OCU DTCs. The OCU
  `0x5A7` heartbeat is never asserted for this path.
- **Runs an event-driven BAP state machine** (vendored `src/bap` transport +
  `src/egolf` BatteryControl codec on LSG `0x25`), advanced by the Battery Control
  unit's own replies, not by blind per-tick resending:
  1. channel-open handshake `GET 19 42` / `19 41`;
  2. **after** its ack (`49 41` / `49 42`) — required, the BCU silently drops a
     function GET sent before registration — a segmented get-all read of the
     global profile `80 04 19 59 <txn> 00 00 04`;
  3. a full read-modify-write "arm" of profile 0 (`29 59 <txn> 00 00 01 <record>`)
     that changes **only the operation byte** and preserves every other stored
     field (target temperature, min-charge, name, …);
  4. the immediate OperationMode trigger (`29 58 00 01` / `00 00`), **gated on the
     arm-write echo** so it never races the multi-frame write and runs a stale op;
  5. confirmation on the BCU's `49 58 <flag>` echo.
  If the wake window expires unconfirmed, OVMS raises a user notification naming
  the cause (CAN transmit error, BAP ERROR / rejected, profile read failure, or
  timeout) instead of assuming success.
- **Climate** arms operation `0x06` — climate + `climatiseWithoutExternalSupply`
  ("Climatise on battery power"). That flag is **force-enabled** so remote
  pre-conditioning works while parked and unplugged, and so it survives a charge
  (which must clear it, below). It is an *allow* flag, not a force: with the cable
  connected the car still uses wall power, as it does with its own default profile.
- **Charge** arms pure charge (`0x01`), clearing both climate bits — on this car
  the on-battery flag is itself a standalone climate trigger, so leaving it set
  made a charge start run climate-on-battery when unplugged. Start/stop is the same
  gated trigger. **Stop is operation-specific** (it only stops what profile 0 is
  armed to), so a charge stop is **refused while climate is running** (it would
  otherwise disrupt climate — stop climate first).
- **Set charge current** edits only `maxCurrent`, snapped to the car's allowed
  steps **{5, 10, 13, 32} A** (per phase; the car overrides out-of-set values),
  with no trigger when idle; while charging it re-applies to the running charge.
- **Metrics:** `v.e.hvac` (climate, from the `49 58` echo + the clima-ECU `0x5EA`
  conditioning bit, cleared on bus sleep), `v.c.climit` (charge-current limit), and
  `v.e.cabinsetpoint` (stored target temperature).

## Safety

`maxCurrent` is **hard-capped at `0x20` (32 A)** on every profile-write path
(`clampMaxCurrent` and again in the arm encoder). Writing a higher value breaks the
car's charging until the settings are reset — the cap is a single chokepoint that
even a preserved/garbled read-back cannot bypass.

## Behaviour notes / caveats (documented in the user guide)

- All charge control acts on the car's **default/global profile only**. A departure
  timer uses its own profile, so an OVMS-set charge-current limit does not apply to
  a timer-scheduled charge, and starting charge via OVMS switches the car to the
  default profile (which charges to 100% SoC). You re-select a timer from the car.
- A charge command clears "Climatise on battery power"; the next climate command
  re-enables it automatically (no action needed in the car).
- Climate and charge can run at the same time on the car; OVMS serialises the
  command *sequences* (one arm+trigger in flight at a time).

## Architecture

The logic lives entirely in `vehicle_vwegolf_bat_ctrl.{h,cpp}` as a self-contained
`VWeGolfBatteryControl` helper. `vehicle_vwegolf.cpp` keeps RX/metric decoding and
the tickers and only forwards to it:

- `CommandClimateControl()` → `m_batctrl.Climate()`
- `CommandStartCharge()` / `CommandStopCharge()` → `m_batctrl.Charge()`
- `CommandSetChargeCurrent()` → `m_batctrl.SetChargeCurrent()`
- `Ticker1()` → `m_batctrl.Ticker1(bus_alive)`
- the `0x17332510` and `0x5EA` frames in `IncomingFrameCan3()`

The **OCU `0x5A7` heartbeat path** used for the body commands (mirror fold / lock /
horn) is unchanged and untouched — BatteryControl rides its own spare-node wake and
never asserts `m_ocu_active`. This is also out of scope for this PR.

`bap-lib` (`src/bap`, `src/egolf`) is a dependency-free C++ library (no
ESP-IDF/Arduino includes) implementing the VW BAP application-layer transport and
the BatteryControl (LSG `0x25`) command/status codec. It compiles unchanged in both
the firmware and the host test build, and is reusable for future BAP features on
this and other MQB vehicles.

## Testing

- **Native unit tests** (`tests/`, host `g++`, no hardware): cover the spare-node
  wake, the handshake→GET gating, the read-modify-write arm and its echo-gated
  trigger, climate vs charge operation bytes, op-specific stop, the
  charge-stop-while-climate refusal, the `maxCurrent` hard cap and step snapping,
  BAP-ERROR/timeout notifications, an empty/short `49 59` frame (no crash), and
  HVAC state from both sources. **136/136 assertions pass.**
- **Valgrind** (memcheck, `--leak-check=full --show-leak-kinds=all`): the native
  suite runs with **0 errors and 0 bytes definitely / indirectly / possibly lost**.
  The only remaining blocks are "still reachable" — the mock's global metric
  singletons that live for the process lifetime, as the framework's metrics do on
  the real firmware.
- **Firmware build** against ESP-IDF (VWEGOLF enabled) links cleanly with the new
  module and bap-lib.
- **On-car:** validated on a 2020 e-Golf — climate start/stop and charge
  start/stop/current apply and are confirmed by the BCU (`49 58 01` / `49 58 00`),
  with no OCU DTCs raised.

## Files

New:
- `src/vehicle_vwegolf_bat_ctrl.{h,cpp}` — the BatteryControl controller module
- `src/bap/{bap.h,bap.cpp}` — BAP application-layer transport
- `src/egolf/{battery_control.h,battery_control.cpp}` — BatteryControl (LSG 0x25) codec
- `tests/test_bat_ctrl.cpp`, `tests/mock/ovms_notify.h` — test suite + mock stub
- `docs/bat-ctrl-bap.md` — protocol notes

Changed:
- `src/vehicle_vwegolf.{cpp,h}` — remove the #1430 climate code, forward to the module
- `CMakeLists.txt`, `component.mk` — build the new sources
- `tests/` — mock additions (`ESP_QUEUED`, notify capture, charge command stubs)
- `docs/index.rst`, `docs/vw-bap-protocol.md`, `docs/vwegolf.dbc`, `changes.txt` — docs

Removed:
- `docs/clima-control-bap.md`, `tests/test_clima.cpp` — renamed to the `bat-ctrl` equivalents


Nueueuet:
@SCjona Profile read/modify/write works on the car now — this is the piece your draft is waiting on, so here's what I found before you spend the days I did.

**The BCU does not serve a GET on ProfilesArray.** Bare `19 59`, and array GETs carrying the range header `[asgTxn][flags|RecordAddr][startIndex][count]` in four variants (with and without PosTransmit, 1 record and 4) — silence every time, not even a BAP ERROR. It only publishes the array as STATUS: unsolicited when something changes in the car, and as the reply to a write.

So a template has to be captured before anything can be written, or you put zeroes into temperature, maxCurrent and the name. I persist profile 0 to config as hex, otherwise the template is gone on every reboot and the feature is dead until the owner next changes something in the car.

**Writes verified on a 2016 e-Golf** via `sendProfileWrite` (RecordAddr 0, position 0), the car echoing the updated record within a second each time:

```
operation   06 -> 02 -> 06     PO_ALLOW_BATTERY off/on
temperature 78 -> 55 -> 78     22.0 -> 18.5 -> 22.0 C
```

maxCurrent, the name and the [RE] middle fields came back unchanged both times. Your `battery_control.h` layout is correct as written — every field I could move from the car's UI landed where your comments say.

One naming thing: what the e-Manager calls the charge target follows byte **[3]** (`minChargeLevel` in your header), not `[6]`. Stepped it through 100/50/10/100 on the car back in August, offset moved with it every time; `[6]` sat at 0 throughout.

Branch: https://github.com/Nueueuet/Open-Vehicle-Monitoring-System-3/tree/climate-v2-with-app-capability (on top of your `2026-08-11_1`, plus the `MP-0 VC26` capability advertisement so the official app shows the climate button — `TransmitMsgCapabilities()` is still an empty stub on master).

Happy to test whatever you want on the car, including cold-bus cases.

SCjona:
thanks, but i already have figured profile read/modify/write out. i just have to test charge control (start/stop, set current) now, gonna add that to the PR, it shares like 90% of the code, so no point in it being an extra PR

Nueueuet:
Good — yours is the one that should land, mine was only ever a feasibility check. I'll drop it.

Two things before you settle the shape of it, though.

**There is a consumer on the app side.** I've added e-Golf controls to the Android climate tab: a target temperature slider and a toggle for pre-conditioning without the charging cable, both driving the stored profile through the module over command 7. It talks to my own commands today, so it will break the moment your version replaces them.

All it needs from the module is something machine-readable to read the current state back — the app can't learn these from the v2 protocol, there are no fields for them. Mine is one line:

```
xvg ccstatus  ->  cctemp=22.0 onbat=1 current=32 valid=1
```

If you expose something equivalent, whatever you call it, I'll point the app at yours. `valid=` matters: the app must be able to tell "not read yet" from a real zero, otherwise it offers a slider for a value nobody knows.

**Offering a second model year for the charge control testing.** Mine is a 2016, yours a 2020, and on this car the differences have not all been where I expected. Happy to run whatever you want on it, including from a cold bus — that case seems to be the awkward one to come by.

One finding from the write path that will hit charge control too, since it is the same mechanism: a write to a sleeping comfort bus fails silently. The driver accepts the frame, nobody acknowledges it, and the caller sees success — so the app cheerfully showed a value the car never received. Worse, those frames sit in the TX queue and after ~30 of them it overflows, which kills receive as well until the module reboots. I ended up gating writes on how long ago the BCU last spoke (refuse past 5 s) and waiting for the car to echo the record back before reporting success; the echo comes in well under a second when it does come.

SCjona:
well regarding your values, there are standard metrics for cctemp (ms_v_env_cabinsetpoint) and current (ms_v_charge_climit). valid can be as simple as checking cctemp or current values against 0. cctemp min value (as per encoding) is 10°C so a value of 0 is literally not possible.

the climate on battery toggle is a bit tricky. i need to unset it for start charging (otherwise it will start climate instead if not plugged in), likewise re-set it on climate to allow climate to work on battery. i still need to figure out how to get climate from wall power working though, in my test it seems like it just started to do it from battery power. need to do more testing

SCjona:
charging control added, ready for testing/review. tested and confirmed working on my car.

Nueueuet:
First, a correction to what I posted earlier. I said the BCU does not serve a GET on the ProfilesArray. That was wrong, and stated far too confidently — I had only ever sent short single frames (bare `19 59`, and four variants with a range header). Your segmented `80 04 19 59 <asgTxn> 00 00 04` is answered, and answered on a 2016 as well. My "does not serve" was really "does not answer the framing I happened to try".

**Tested `dc86d82` on a 2016 e-Golf** (plus my `MP-0 VC26` patch so the app button shows). Everything worked:

```
Climate ON
  Climate arm: profile 0 already op 0x06 — no write needed     <- profile GET answered
  Climate START BAP trigger sent
  BatteryControl OperationMode 49 58 01 -> HVAC ON
  Climate ON confirmed — releasing NM, cluster self-sustains

Climate OFF
  49 58 00 -> HVAC OFF, confirmed

charge current 16   (car plugged in and charging)
  SetChargeCurrent -> 13 A                                      <- snapped, as intended
  arm: profile 0 maxCurrent 0x20 -> 0x0d
  written — re-applying to running charge
  v.c.climit = 13A

charge current 32   -> 0x0d -> 0x20, v.c.climit = 32A
```

`v.e.cabinsetpoint` (22.0) and `v.c.climit` both populate. The arm correctly skipped the write when the op byte already matched.

One small thing: during the `SetChargeCurrent` re-apply the log reads "Climate START BAP trigger sent". It is the shared trigger, and `m_apply` is properly gated on a charge being in progress — but reading the log live, it took me a minute to convince myself the car had not started conditioning while plugged in. A different wording for the re-apply would save the next person that moment.

**On the metrics for an app-side control:** `ms_v_charge_climit` does reach the app — it is in the server-v2 status record. `ms_v_env_cabinsetpoint` does not; grepping `ovms_server_v2.cpp` it appears nowhere in the protocol. So the temperature cannot be read app-side through metrics, only through a command.

Which leads to the one thing your module does not have: setting the target temperature. `SendArm` already does the RA0 read-modify-write and explicitly preserves the temperature field, so a `SetClimateTemp` would be a small addition to what is already there — byte 12, `raw = degC * 10 - 100`, half-degree steps between 15.5 and 30.0. Is that something you want in this PR? Happy to send it as a patch against your structure rather than have it live off to the side in mine, and I can test it here.

Nueueuet:
Built the temperature setter against your structure and tested it on the 2016. One commit on top of `dc86d82`, yours to take or leave:

https://github.com/Nueueuet/Open-Vehicle-Monitoring-System-3/commit/390c3b4a

It is `BC_SET_TEMP`, shaped like `BC_SET_CURRENT`: read profile 0, change the temperature byte, write it back, confirm on the write echo, no trigger. The operation byte is left exactly as read, so it can neither start nor stop anything. The raw encoding (`degC * 10 - 100`) goes in as the command param so it fits the existing `uint8_t` — 15.5 C is 55, 30.0 C is 200.

Three places needed the new op next to `BC_SET_CURRENT`, so they share an `IsSettingsEdit()` predicate now: the post-arm phase choice, the lost-echo retry, and the skip-write check. That last one is the one that bites — it compares only the operation byte, so my first attempt looked like a no-op, wrote nothing at all, and then fired the trigger. Confirmation lives inside the `BC_SET_CURRENT` branch of the write-echo handler too; without its own the command never completed and re-armed every two seconds until the window closed. Both are the kind of thing only the state machine's author would spot at a glance, so do look at those hunks with suspicion.

Also in the commit, and easy to drop if you would rather not carry them: `xvg cctemp <15.5..30.0>` and `xvg ccstatus` (`cctemp=22.0 current=32 valid=1`). The Android climate tab I mentioned uses them — `valid` follows your suggestion of testing the temperature against 0.

On the car:

```
SetClimateTemp arm: profile 0 temperature 0x78 -> 0x5f (5 frames)
SetClimateTemp 19.5 C written -- BCU echoed 49 59 b3
SetClimateTemp confirmed — releasing NM, cluster self-sustains
```

22.0 → 19.5 → 22.0, one write each, echo inside a second, and `ms_v_env_cabinsetpoint` follows. Charge current and climate still behave as before alongside it.

The on-battery toggle is gone from my app work, by the way — you were right that it cannot be a user setting while the module owns that bit. Climate always arms it now, which is what my car needs anyway since it is usually parked unplugged.

SCjona:
@Nueueuet you can file that as follow up PR if you want. I'm not gonna include it here. Also I think you still would have to restart (just start, no stop needed i think) climate to apply the changes (and write the temp before starting climate). The BCU did not apply the charge current when editing the profile, so I think the same goes for the temperature setpoint. But I'm sure you'll test that before submitting a PR by switching between cooling and heating while running climate pre-conditioning.

Nueueuet:
You were right to make me check, and the answer is the opposite of the charge current: the BCU **does** pick up a new setpoint while conditioning is already running. No restart, no second trigger.

2016 e-Golf, plugged in but not charging, SOC 83.5 %, ambient 28 °C, on `dc86d82` + `BC_SET_TEMP`. Numbers are `v.e.cabintemp` (0x5EA), polled every 20 s from the module console.

**Run 1 — target 30.0, start, then drop to 16.0 mid-run**

```
[  75.4] >>> climatecontrol on
[  78.8] Climate arm: profile 0 already op 0x06 — no write needed
[  82.7] BatteryControl OperationMode 49 58 01 -> HVAC ON
[  83.5] Climate ON confirmed — releasing NM, cluster self-sustains
[ 300.1] >>> xvg cctemp 16.0
[ 303.3] SetClimateTemp arm: profile 0 temperature 0xc8 -> 0x3c (5 frames)
[ 303.3] SetClimateTemp 16.0 C written -- BCU echoed 49 59 b6
[ 660.3] >>> climatecontrol off
```

| window | target | cabin | rate |
|---|---|---|---|
| 1:23 – 5:00 | 30.0 °C | 35.2 → 34.5 °C | −0.19 °C/min |
| 5:43 – 7:23 | 16.0 °C | 34.0 → 30.3 °C | **−2.2 °C/min** |
| 7:23 – 11:00 | 16.0 °C | 30.3 → 27.9 °C | −0.66 °C/min |

Three and a half minutes at a 30 °C target with a 35 °C cabin and it barely moved — it was honouring the setpoint. Roughly 40 s after the write the curve bends and the cooling rate goes up tenfold. It levels off just under 28 °C, which is ambient.

**Run 2 — the other way round, cooling to heating**

```
[   5.1] >>> climatecontrol on
[  10.9] BatteryControl OperationMode 49 58 01 -> HVAC ON
[ 240.4] >>> xvg cctemp 30.0
[ 243.2] SetClimateTemp arm: profile 0 temperature 0x3c -> 0xc8 (5 frames)
[ 243.2] SetClimateTemp 30.0 C written -- BCU echoed 49 59 bb
[ 600.4] >>> climatecontrol off
```

| window | target | cabin | rate |
|---|---|---|---|
| 0:22 – 3:43 | 16.0 °C | 31.0 → 26.7 °C | −1.3 °C/min |
| 4:03 – 10:00 | 30.0 °C | 26.7 → **28.4 °C** | +0.29 °C/min |

Still falling when the write went in, flat for one poll, climbing after that. Same latency both times, about forty seconds from the write to the reversal.

`v.e.hvac` stayed `yes` across both ten-minute runs and the only `49 58 00 -> HVAC OFF` in either log came from the stop command — the cluster sustained the session on its own after your NM bridge let go, eight and thirteen seconds in.

One thing that cost me a moment and is worth knowing for anyone reading `v.e.cabintemp`: it read 27.9 °C with the car asleep and jumped to 31.0 °C within twenty seconds of the blower starting. The sensor only sees real cabin air once air moves past it, so the first reading after a start measures nothing.

So `xvg cctemp` needs no re-trigger and a slider in the app does what it says at the moment it is moved. I will put the setter up as a follow-up PR once this one lands, without the app-capability commit that was bundled in the branch I linked earlier — that one is a separate concern and belongs in its own PR.

Issue #196, opened 2019-01-19
VOLT / AMPERA support


liebefuroscar:
Opel Ampera / Chevrolet Volt users need to be included with priority:
- Preclimatization (OBD MX already has it)
- Load amperage selection

In a less priority way:
- The estimated autonomy is displayed (show 0)
- Tire pressure shown (show 0 psi)
- The motor temperature is displayed (show 0)
- Odometer Km shown in the app (show 0)

We are many users of Opel Ampera / Chevrolet Volt with OVMS that we miss these functions, which already has, for example, Tesla Roadster or Nissan Leaf.

I am also sure that if these features are added, many more users of these VE will buy an OVMS.
Thank you

![photo5917743255284461601](https://user-images.githubusercontent.com/25435711/51430147-1265d580-1c17-11e9-8de0-fe31300d3e2e.jpg)
![photo5917743255284461602](https://user-images.githubusercontent.com/25435711/51430148-1265d580-1c17-11e9-8c60-239241ede49f.jpg)

raulcorreo:
Por favor, necesitamos estás funciones.
Hemos confiado comprando el dispositivo... Creo que lo merecemos....

markwj:
We are limited by what the car can show on the CAN BUS, and what we can decode (or discover) as available. Very willing to implement these features, but we need the codes (which are not publicly available).

lolvince:
Did you already see that ?
https://m.youtube.com/watch?v=xcfdSGDaqfs

mjuhanne:
I'm in the middle of implementing extended Volt/Ampera support for OVMS (see my fork for preliminary work) and I'll make a pull request when it gets more stable and is usable on the mobile app.
However most of the interesting stuff and functionality happens on Single-Wire CAN (SWCAN, Low speed GMLAN) bus and unfortunately the CAN transceivers OVMS uses (MCP2515) won't be able to drive SWCAN with full functionality. It should be able to read and write CAN messages without bus errors, but waking the car (for unlocking/locking doors, remote heating, getting metrics updates etc) requires driving the SWCAN bus to 12 volts (High Voltage Wake Up) and this MCP2515 is unable to do. 
[...]

PR #1502, opened 2026-08-21
Volt/Ampera: Gen 2 SWCAN telemetry, charge limit and vehicle controls

lululombard:
Adds GMLAN low speed (SWCAN, can4) decoding and a set of controls for the Gen 2 Volt alongside the existing HS-GMLAN polling, plus a documentation page.

Addresses some of the concerns of #196

Broadcasts are matched on the 13 bit PID rather than the full arbitration ID, since GMLAN packs 29 bit IDs as priority | PID | source and the source byte differs between cars. Motorola sawtooth signals are extracted with a shared helper pinned against known DBC layouts.

Telemetry:
  - drive cycle energy, distance and efficiency, plus gasoline economy
  - both state of charge scales, the 8 bit dashboard figure and the 16 bit pack figure, which are separate reads and not derived from each other
  - EV range, gasoline range and their total, in whole km as the car reports them. Gasoline range stays at 0 until the car is switched on: the DBC validity bit next to it never asserts, so the field itself is the signal
  - oil life, read only, with no reset path anywhere
  - charge energy for the session and lifetime, window positions, climate and battery thermal values

Window position is only readable while the glass moves. At rest the BCM sends 5 in a window's field to mean "no reading", and it does so per window rather than only for the whole frame, so a climate start would otherwise decode as both rear windows sliding 83% open with nothing having moved. Measured over 476 readings: the driver window reports 0, 1, 2, 3 and 6 and never 5, while the other three report only 0, 5 and 6.

Controls, all behind xva/control.enabled:
  - charge limit by deferring and resuming the charge, with a one off override to reach 100% and a geofence so it never caps at a public charger
  - heated seats, by transmitting a button press and letting the seat module respond
  - horn, exterior light flash and both together, via the OnStar telematics frame
  - windows, trunk and an engine override

Departure mode is persistent in the HPCM, so a charge deferred at the limit outlives the unplug and the car refuses to charge next time. It is undone on the way out, when the car is unplugged and awake, rather than detected on the way back in: at a public charger the cable goes in seconds after the driver gets out and the car is still awake, but at home the car is unloaded first and is asleep before the cable goes in, which is exactly when nothing is polling and a stale defer cannot be noticed. The flag recording the pending write is in config rather than RAM, since losing it to a reboot is what strands a car in departure mode with nothing left that knows why.

Every control path is non-blocking. Commands arrive on the events task and tickers run on the vehicle task, so long sequences are driven from ticker state machines rather than loops.

Window commands always run a full wakeup first. A bare high voltage frame wakes the bus electrically but leaves the BCM unable to act on the command, and gating the wakeup on recent bus traffic makes the second of two commands fail silently.

Charge state notifications are left to the framework. Writing v.c.state arms its own notification, so notifying alongside it sends every charge message twice.

Gen 1 cars keep what they had. The transmission temperature poll the drive unit pair replaced is restored behind it, v.c.climit falls back to 16 A only while nothing better has been read, and an unrecognised charger supply level is ignored rather than acted on.

The web page presents status and controls as rows carrying their own state, with both directions of every control available at all times. State decides which button is highlighted, never whether a button exists, because the car stops reporting the moment it sleeps and a control that points the wrong way on a stale reading is worse than no control at all.


dexterbg:
Lulu, welcome :-)

Lots of added telemetry and control features in your PR, very nice!

I have no means to test vehicle specific functions, but generally any control function that can be misused to get access to a vehicle or control the vehicle behaviour in a potentially dangerous way (like turning off the engine remotely while someone is driving the vehicle at high speed…) must be protected by the additional PIN check.

This obviously also applies to the lock/unlock commands and your special window control commands. Obviously the current/previous version of the adapter also lacks PIN checks on the lock/unlock commands, but that should not be taken as a good example, rather the opposite ;-)

The framework includes the `OvmsVehicle::PinCheck()` method, so adding the check is simple for commands already expecting the PIN. For new and custom commands falling into the security relevant category, you need to add a mandatory PIN argument.

Regards,
Michael

lululombard:
Thanks for the welcome @dexterbg!

I have to admit I took the Bolt EV as my main template, and it has no PIN checks on lock/unlock or anything else, so I inherited that. Point taken.

I've gated lock, unlock, both window directions, the trunk release and forcing the engine on or off behind `PinCheck()`. `xva engine auto` is deliberately not gated, since it only releases an override, and a forced state that can't be released seemed worse than one that could be set.

One thing I left ungated and would like your read on: horn, lights and the OnStar "vehicle locate" that fires both. They're loud, but they don't open the car or change how it drives, so they didn't seem to meet the bar. Happy to gate them if you disagree.

For context on the engine, it's mostly a generator and switching it off just drops the car back to EV mode (which still delivers full power and same top speed) rather than cutting drive. It also can't be forced on when the car is off: `0x7E1` only answers with the car powered, so it needs the car on or remote climate already running. I gated it anyway, because `climatecontrol` isn't gated on any car and it supplies exactly that precondition, so climate on followed by engine on would otherwise be an unauthenticated path to a running engine, which could be bad, for instance in a closed garage.

dexterbg:
That should be OK, thanks for the quick rework.

PR #1504, opened 2026-08-22
webserver: stop the file API holding three copies of the body in RAM

kezarjg:
`HandleFile` serves any readable path under `/store` or `/sd`, so the body it returns on a GET is bounded only by the file it is pointed at. That body was passed to `PageContext::print()`, and the peak cost of doing so is three full-size copies of the file:

| | source | size |
|---|---|---|
| the loaded body | `load_file()` into an `extram::string` | N |
| the print argument | `PageContext::print(const extram::string text)` takes it **by value** | N |
| the connection buffer | `mg_send_http_chunk()` appends it; the mbuf grows by `MBUF_SIZE_MULTIPLIER` 1.5 | 1.5N |

Nothing frees the earlier copies before the later ones are built, because the send buffer cannot drain while the handler is still running. So the peak is roughly 3.5x the file, and the mbuf leg needs its 1.5N contiguous. All three live in SPIRAM, since `mg_locals.h` maps the mongoose allocators onto `ExternalRam*`; they fall back to internal RAM only when SPIRAM cannot supply a contiguous block.

This streams the body straight out of the buffer it was loaded into, so neither of the other two copies happens. Peak drops from roughly 3.5x the file to roughly 1x.

## Evidence

The measured evidence comes from the same code shape rather than from this handler. My e-TNGA vehicle module writes a report at the end of each charge session and offers a page to browse and download them; a long DC session CSV runs to a few megabytes. That page printed whole bodies and died reproducibly after about two dozen files and roughly 660 KB in one session, with the failing allocation usually somewhere unrelated. After handing the body to the chunked sender it completed 72 files and about 20 MB cleanly.

Fixing that is what produced `HttpExtRamStringSender`, and `HandleFile` is the upstream code with the same shape, which is why this PR exists as its own change rather than buried in a vehicle PR.

`/api/file` itself has been exercised at 2.7 MB after the change, with byte-identical output. I do not have a pre-change failure on this specific handler.

Measured on an OVMS v3 module reading SD-card files over the local network. Those runs were on my own tree, which carries this change plus unrelated work: the code path is identical, but the module hours are not attributable to this branch alone.

## The change

`HandleFile`'s GET path hands ownership of the loaded body to a chunked sender and returns. The sender frees the body and itself after the last chunk and emits the terminating zero-length chunk, which is why this path deliberately does not fall through to `c.done()`. An empty body, meaning a directory path, is handled: the sender terminates on its first event.

`HttpStringSender` is generalised to `HttpStringSenderT<StringType>`:

```
typedef HttpStringSenderT<std::string>      HttpStringSender;
typedef HttpStringSenderT<extram::string>   HttpExtRamStringSender;
```

The old name survives as a typedef, so this is source compatible, and nothing else in the tree referenced the class, so `HandleFile` is the only behavioural change. The implementation moves into the header because it is now a template, which is a mechanical move with the `ESP_EARLY_LOGV` traces preserved verbatim (the log tag is the literal `"webserver"`). The one functional edit inside the sender is computing the chunk length with a comparison rather than the `MIN` macro, which is not visible in the header.

## Testing

Builds clean in CI with no vehicle modules enabled, so the result is attributable to this change alone. That covers compilation; the figures above are the behavioural evidence.

## Worth knowing

The streaming path is roughly 1.6x slower than serving the same file statically. Irrelevant when editing a script, but a client pulling a multi-megabyte file through `/api/file` needs a timeout above about 150 seconds, and a default-timeout client may now give up where it previously succeeded.

I could not find an existing issue for this. It surfaced while chasing an unrelated memory problem rather than from a user report.

## Scope: GET only

The save path is untouched and cannot be fixed the same way. `PageContext::getvar(name, extram::string&)` allocates a buffer the size of the entire url-encoded body while `hm->body` is still in `recv_mbuf`, and mongoose has already buffered the complete request before `MG_EV_HTTP_REQUEST` fires. Streaming the save side means the multipart or streaming request path, which is a different API shape for `/api/file` rather than an extension of this change. Happy to write it as its own PR if that is wanted.

## Three decisions I would happily revisit

**Only HandleFile is converted.** The other `c.print()` sites on an `extram::string` are `PluginHandler`, which holds a reference to registry-owned content and so cannot pass ownership without a copy, and `PluginCallback`, which emits a fragment into a composed page where a stream-and-return sender would truncate everything after it. Plugin content is also bounded in practice, unlike an arbitrary path under `/sd`. Happy to add a non-owning variant if you want those covered.

**Template versus a second class.** Duplicating forty lines to change one type seemed the worse trade, but this does touch code that already works. Easy to switch to a standalone extram sender alongside the concrete class if you prefer.

**Streaming is unconditional.** Small files pay the slowdown for a problem they do not have. A size threshold avoids that at the cost of two paths to keep correct, and I erred toward one.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.cpp
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.cpp
@@ -871,58 +871,6 @@ int HttpDataSender::HandleEvent(int ev, void* p)
 
 
 /**
- * HttpStringSender: chunked transfer of a memory region (needs to be const during xfer)
- */
-HttpStringSender::HttpStringSender(mg_connection* nc, std::string* msg, bool keepalive /*=true*/)
-  : MgHandler(nc)
-{
-  m_msg = msg;
-  m_sent = 0;
-  m_keepalive = keepalive;
-  ESP_EARLY_LOGV(TAG, "HttpStringSender[%p]: init msg=%p, %d bytes", nc, m_msg, m_msg->size());
-}
-
-HttpStringSender::~HttpStringSender()
-{
-  if (m_sent < m_msg->size()) {
-    ESP_EARLY_LOGV(TAG, "HttpStringSender[%p]: abort msg=%p, %d bytes sent", m_nc, m_msg, m_sent);
-  }
-  delete m_msg;
-}
-
-int HttpStringSender::HandleEvent(int ev, void* p)
-{
-  switch (ev)
-  {
-    case MG_EV_SEND:          // last transmission has finished
-    {
-      if (m_sent < m_msg->size()) {
-        // send next chunk:
-        size_t len = MIN(m_msg->size() - m_sent, XFER_CHUNK_SIZE);
-        mg_send_http_chunk(m_nc, (const char*) m_msg->data() + m_sent, len);
-        m_sent += len;
-        ESP_EARLY_LOGV(TAG, "HttpStringSender[%p] msg=%p sent %d/%d", m_nc, m_msg, m_sent, m_msg->size());
-      }
-      else {
-        // done:
-        if (!m_keepalive)
-          m_nc->flags |= MG_F_SEND_AND_CLOSE;
-        mg_send_http_chunk(m_nc, "", 0);
-        ESP_EARLY_LOGV(TAG, "HttpStringSender[%p]: done msg=%p, %d bytes sent", m_nc, m_msg, m_sent);
-        delete this;
-      }
-    }
-    break;
-
-    default:
-      break;
-  }
-
-  return ev;
-}
-
-
-/**
  * CheckLogin: check username & password
  *
  * We use "admin" as a fixed username for now to be able to extend users later on
--- a/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.h
+++ b/vehicle/OVMS.V3/components/ovms_webserver/src/ovms_webserver.h
@@ -334,23 +334,75 @@ class HttpDataSender : public MgHandler
 
 
 /**
- * HttpStringSender transmits a std::string in HTTP chunks of XFER_CHUNK_SIZE size.
+ * HttpStringSenderT transmits a string in HTTP chunks of XFER_CHUNK_SIZE size.
  * Note: the string is deleted after transmission.
+ *
+ * Two instantiations are provided: HttpStringSender for a std::string, and
+ * HttpExtRamStringSender for an extram::string. Prefer the latter for file-sized
+ * bodies: the sender takes ownership of the buffer the body was loaded into and
+ * emits it in place, so serving a large body needs no second full-size allocation.
  */
-class HttpStringSender : public MgHandler
+template <class StringType>
+class HttpStringSenderT : public MgHandler
 {
   public:
-    HttpStringSender(mg_connection* nc, std::string* msg, bool keepalive=true);
-    ~HttpStringSender();
+    HttpStringSenderT(mg_connection* nc, StringType* msg, bool keepalive=true)
+      : MgHandler(nc)
+    {
+      m_msg = msg;
+      m_sent = 0;
+      m_keepalive = keepalive;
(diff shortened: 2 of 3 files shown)


dexterbg:
Some notes on this for discussion -- not intending to dismiss this, replacing potentially large RAM buffers by streaming is normally a good idea.

First of all, the assumption of Mongoose buffers lying in internal 8-bit RAM is wrong. We generally override the Mongoose memory functions with the SPIRAM versions -- see: https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/blob/master/vehicle/OVMS.V3/components/mongoose/include/mg_locals.h#L87

Still the double buffering may be the source of your observed issue, because SPIRAM also is limited, and both buffers need to be contiguous, so fragmentation may become an issue.

Second: I probably missed stating this somewhere explicitly: using the file API to retreive huge files is far outside the intended use. The API was/is meant to provide an additional data storage option aside the config store to web plugins, i.e. comparably small files that normally reside on `/store`.

Accessing log files et al is meant to still be done via direct file download (paths mapped into web root, streamed by Mongoose's `mg_serve_http()` directly from disk = with small & fixed RAM overhead).

That doesn't necessarily mean we should not leverage the file API for this use case, but can you please elaborate on why you want or need to use the API for log files?

Third: if (?) handling large files shall now be defined to be a valid API use case, then *any* buffering of the whole file in RAM would need to be avoided -- on saving just as well as on loading. In that case, the proposed change is less than half the solution, the full solution would need to stream both GET & POST directly from/to disk.

Regards,
Michael

kezarjg:
Michael, thanks for reviewing this. Taking your three points in order, and the first one is a straight correction of mine.

## 1. The mongoose allocators: you are right, and the PR is wrong

The send buffer is SPIRAM. My "internal 8-bit RAM" framing is wrong, and it is in the PR title, the body, the commit message and a code comment. I have rewritten them.

## 2. Why I was in the file API at all: fair question, and my evidence paragraph was misattributed

Straight answer: I do not need the API for regular log files My e-TNGA vehicle module writes a report at the end of each charge session and offers a page to browse and download them, HTML to read in the browser and CSV to keep. A long DC session CSV runs to a few megabytes. That page originally printed the whole body, and it is where the failure I quoted was measured: it died reproducibly after about two dozen files and roughly 660 KB in a session, and after handing the body to a chunked sender it completed 72 files and about 20 MB clean. Fixing it is what produced `HttpExtRamStringSender`. My apologies for the confusion.

## 3. GET only: agreed, with one structural caveat

You are right that this is not the whole thing. It is worth saying that the two halves are not symmetric, though.

POST cannot be fixed at handler level at all. `PageContext::getvar(name, extram::string&)` at `web_framework.cpp:151` does `dst.resize(hm->body.len)`, allocating a buffer the size of the entire url-encoded body while `hm->body` still sits in `recv_mbuf`; for binary content the encoded form is up to 3x the file. And mongoose has already buffered the complete request before `MG_EV_HTTP_REQUEST` fires, so no change inside `HandleFile` can avoid that. Streaming the save side means the multipart or streaming request path, which is a different API shape for `/api/file`, not an extension of this patch. Happy to do it, but it is its own PR and its own compatibility discussion.

So I think the decision in front of you is not "half a fix versus a whole fix", it is:

- **Take this as a strict improvement.** It does not declare large files supported. It removes about 2.5 of 3.5 whole-file SPIRAM allocations from a path any plugin can reach, is source compatible, and leaves the POST question exactly where it is today.
- **Enforce the intended use instead.** If the API is defined as small-files-only, the consistent fix is a documented size limit on both directions returning 400. I run that downstream at 512 KB and it works fine. It is a behaviour break for any plugin currently reading something larger, and at 3.5x a 512 KB file still costs about 1.8 MB transient, which is why I would want this change underneath a cap rather than instead of it.
- **Redesign both directions to stream from and to disk.** The real answer if large files become supported, and I am willing to write it, but I would want the API contract settled first.

dexterbg:
Still thinking about limiting both directions, but your point two is valid, and the limit should be a separate commit anyway, so merging this as is -- after all it *is* a major improvement on memory usage, regardless of the file size.

kezarjg:
Thanks. Maybe now I can finally merge my vehicle module! 

Let me know if you want me to take this further.

PR #1505, opened 2026-08-23
smartEQ: CAN transceiver switch (active/listen mode)

zorgms:
PR is based on #1494 and is an alternative proposal @zbchristian can switch the website to radio buttons and make further adjustments if necessary; I'll be AFK for a while.

### DIFF
--- a/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can_poll.cpp
+++ b/vehicle/OVMS.V3/components/vehicle_smarteq/src/eq_can_poll.cpp
@@ -391,6 +391,7 @@ void OvmsVehicleSmartEQ::PollReply_BMS_HVContactorCycles(const char* data, uint1
         smartOBDpolling(false);
         MyConfig.SetParamValueBool("xsq", "canwrite", false);
         MyConfig.SetParamValueBool("xsq", "canwrite.caron", false);
+        MyConfig.SetParamValueBool("xsq", "canwrite.caroff", false);
         }
       }
     // Alert if consumed cycles are above expected for a healthy contactor (e.g. above 50000 cycles consumed)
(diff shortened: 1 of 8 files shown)


zbchristian:
@zorgms , your code does not work. Transitions are not correctly triggered. 

Starting from your code, I streamlined the logic to avoid, that this is scattered over multiple functions. The code relies now mainly on the state of `IsCANWrite()`. 

I tested this version now with all possible combinations with my Arduino emulated SmartEQ and it does its job correctly now.

Probably you should have a look at the code, before I create a PR, or you add the changes to your PR.
The code is [here](https://github.com/zbchristian/Open-Vehicle-Monitoring-System-3-Lilygo-Support/tree/smarteq_add_listenonly_v2) in branch `smarteq_add_listenonly_v2`.

zorgms:
@zbchristian 
I’ve given it a quick once-over on my phone; it looks good. I think we can go ahead with it, provided there’s no reason not to switch over from the bus.

dexterbg:
Codewise looks good to me as well.

So if this is the way agreed upon, please create a PR for your modification, Christian.

zbchristian:
@zorgms, I went through the code once more and most calls of `smartOBDpolling()` are now just using `IsCANWrite()` as the parameter. This is redundant, since `IsCANWrite()` is already evaluated inside of the function. So, I set the default of the function to true, to avoid multiple evaluations of `IsCANWrite()` for a single call. Only setting `activate=false` will have a direct effect by  switching off the CAN access. In all other cases the state depends on `IsCANWrite()`.

You might want to have another look, before I submit the PR.

One more remark, the pull request would contain your changes to the 12V alert threshold as well. 
Is that ok with you?

zorgms:
I can currently only give the code a quick glance on my phone; if something isn't quite right, I'm sure the community will provide feedback. Please include the adjustments to the default settings; there was a case where trickle charging was supposed to start (at 11.6V) but didn't, apparently because the battery voltage was below 11.5V. I therefore adjusted the alarm threshold from 0.9V to 0.75V. At 11.75V, the same EQ was able to activate trickle charging.

dexterbg:
Closing in favor of PR #1509

PR #1506, opened 2026-08-23
Server V2: advertise capability C26 so the app can offer climate control

Nueueuet:
# Server V2: advertise capability C26 so the app can offer climate control

## The problem

`OvmsServerV2::TransmitMsgCapabilities()` is an empty stub, so no V3 module ever
sends the `V` message. The official Android app calls `hasCommand(26)` before it
enables the climate control button, and since that is never true it falls back to
a hard-coded list of car types:

```java
// The V3 framework does not support capabilities yet, but
// the Leaf, Smart, VW e-Up and Zoe Ph2 are the only cars providing command 26 up to now, so:
if (carData.hasCommand(26)
    || carData.car_type == "NL"
    || carData.car_type == "SE"
    ...
```

Anything outside that list gets a greyed-out button no matter what the firmware
can do. That is a real, reported problem: see
https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/issues/690#issuecomment-5337489845
— a 2020 e-Golf on current edge, where the firmware has climate control and the
app will not ask for it.

## What this does

* `OvmsVehicle::SupportsClimateControl()`, virtual, defaults to `false`, so no
  existing vehicle changes behaviour.
* `TransmitMsgCapabilities()` sends `MP-0 VC26` when the active vehicle
  overrides it to `true`, following the same requested/always pattern as the
  other `TransmitMsg*` methods.
* `vehicle_vwegolf` opts in.

Any other vehicle that implements `CommandClimateControl()` opts in with one
line and its users get the button without waiting for an app release. I have
deliberately not flipped the flag on vehicles I cannot test — the ones the app
hard-codes today (Leaf, Smart, e-Up, ZOE Ph2, Volt/Ampera) are the obvious
candidates, and their maintainers know better than I do whether the command is
complete on their car.

## Testing

* 2016 VW e-Golf, module on cellular: the climate button appears in the official
  app, and both start and stop reach the car.
* The e-Golf component's native tests still build and pass (88/88) — the mock
  mirrors the new virtual.

### DIFF
--- a/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
+++ b/vehicle/OVMS.V3/components/ovms_server_v2/src/ovms_server_v2.cpp
@@ -1735,7 +1735,20 @@ void OvmsServerV2::TransmitMsgEnvironment(bool always)
 
 void OvmsServerV2::TransmitMsgCapabilities(bool always)
   {
+  bool requested = m_now_capabilities;
   m_now_capabilities = false;
+
+  // Quick exit if not explicitly requested (mirrors the other TransmitMsg*)
+  if ((!always) && (!requested)) return;
+
+  OvmsVehicle* vehicle = MyVehicleFactory.ActiveVehicle();
+  if (vehicle == NULL) return;
+  if (!vehicle->SupportsClimateControl()) return;
+
+  // Protocol v2 message 0x56 "V": comma separated capability list,
+  // C<cmd> = vehicle supports command <cmd>.
+  // 26 = Remote Climate Control -> app shows its climate button.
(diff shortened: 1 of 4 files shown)


markwj:
This seems kludgy. Addressing that one capability, but ignoring all the others.

Wouldn't it be better to properly address the capabilities system and publish a valid capability string outlining all the capabilities?

SCjona:
@Nueueuet I created a PR that just adds VWEG to the hardcoded whitelist of the app https://github.com/openvehicles/Open-Vehicle-Android/pull/256

It's a minor change, it should get merged a lot faster then any capability based stuff.

PR #1521, opened 2026-09-12
vwegolf: drop spurious "Not charging" alert on power-off

SCjona:
## What
Stop the VW e-Golf emitting a "Not charging" notification every time the car ignition is switched off.

## Why
The 0x594 handler wrote `v.c.state` = "stopped" unconditionally on every non-charging frame. The first idle 0x594 after any OVMS (re)start therefore changed `v.c.state` from "" to "stopped", which the base vehicle framework reports as a charge-stop → a "Not charging" push notification on every power cycle. The handler also called `NotifyChargeStart/Stopped()` directly, which double-notified on top of the framework's own charge-state notifications.

## How
- Set `v.c.state` = "stopped" only on a genuine charging→not-charging transition (guarded by `was_charging`), not on every idle frame.
- Drop the redundant direct `NotifyChargeStart/Stopped()` calls — `OvmsVehicle` already notifies on the `v.c.state` transition.

`CommandStat` still renders "Not charging" for an empty/plug-less state, so the `stat` command output is unchanged.


dexterbg:
There's a non-trivial merge conflict from your previous PR, please resolve. Also please remove the `changes.txt` addition, as this is just a bug fix without functional change for the user.

SCjona:
conflict resolved, `changes.txt` addition removed

PR #1526, opened 2026-09-18
smartEQ: Require IsOnEQ for TPMS; fix polling; GPS logging

zorgms:
Only handle TPMS CAN frames when the vehicle is on by adding an IsOnEQ check in eq_can.cpp. In eq_features.cpp make the smartSleep() OBD-polling call explicit (smartOBDpolling(false)) to avoid accidental enabling via a negated flag.


dexterbg:
Introducing a new GPS history log is a user level change that needs to be included at least in the `changes.txt`, especially since it is mandatory. Users should know this is intentional. Speaking of mandatory: this should be optional, and users should need to enable this explicitly.

I know it's mostly the V2 `L` message, which also is kept for a day on the V2 servers already, but when moving to V3, users should get better control over data logged remotely carrying potentially private information.

I further recommend documenting the record structure in the Smart user manual section. Consider using this as a template: https://docs.openvehicles.com/en/latest/userguide/notifications.html#grid-history-log

I'll add a respective CSV header for the record type to my dowload tool.

dexterbg:
Sorry, was too quick, just saw an issue with your record type. I'll fix that for you.

PR #1528, opened 2026-09-19
Leaf: Add 12v Battery Current, Fix Battery Type detection, Update WebUI, Battery capacity calculation fix, Code tidy up

wayne-love:
- Add 12v Battery Current polling
- Fix battery type detection so that battery type logic works properly on ZE1
- Update WebUI with more descriptive comments on
- Remove code for packets that are now filtered out
- Remove the capacity calculation from 59e packets as 59e is for QC capacity not total capacity
- Explicitly constrain 5b3 SOH code to non ZE1 variants


wayne-love:
Some of these patches commits appear to be duplicated.  This is the result of a rebase that did during the code dev cycle.  Whist they add noise they do not appear to cause any issues with the code.

PR #1532, opened 2026-09-20
gloal: auto check climate control available

zorgms:
This routine checks whether the climate command for the active vehicle has been overridden, sets the `v.e.climate.ctrl` metric accordingly, and transmits the value via a V2D message.

Co-author Gemini 3.8 Flash

This is an alternative to #1506


dexterbg:
Marks comment on #1506 (https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3/pull/1506#issuecomment-5389367540) applies here just the same.

Please join the discussion on the developer list.

